PR #7, opened 2018-09-13
Add parity/parity image

LLFourn:
This PR: 
- adds the parity/parity ethereum node container image
- Changes the `.logs` method to return a struct containing both stderr and stdout so the caller can wait on a message in stderr
- Changes Ganache to depend on the rust-web3's github repo directly because of a dependency issue (https://github.com/tomusdrw/rust-web3/pull/154)


testcontainers/src/api.rs
```diff
@@ -175,6 +175,11 @@ impl Ports {
     }
 }
 
+pub struct DockerLogs {
```
thomaseizinger: I think the `Logs` is sufficient.

images/parity_parity/src/image.rs
```diff
@@ -0,0 +1,65 @@
+pub struct ParityEthereum {
+    arguments: ParityArgs,
+}
+
+#[derive(Clone)]
+pub struct ParityArgs {
+    pub unsafe_expose: bool,
+    pub jsonrpc_apis: String,
+    pub config: String,
```
thomaseizinger: Could we make this an enum instead?

images/parity_parity/src/image.rs
```diff
@@ -0,0 +1,65 @@
...
+            },
+        }
+    }
+}
+
+impl Image for ParityEthereum {
+    type Args = ParityArgs;
+
+    fn descriptor(&self) -> String {
+        "parity/parity".to_string()
```
thomaseizinger: This should have a tag associated to ensure stable behavior.

images/parity_parity/src/image.rs
```diff
@@ -0,0 +1,65 @@
+pub struct ParityEthereum {
+    arguments: ParityArgs,
+}
+
+#[derive(Clone)]
+pub struct ParityArgs {
```
thomaseizinger: The convention that we followed so far was `<ImageName>ImageArgs`, so this would be `ParityEthereumImageArgs`.

Although the docker image is named `parity/parity`, I think it would be better to name the whole crate `tc_parity_parityethereum` to make it more clear. They also renamed the github repository recently, so it is probably just a matter of time until the rename the docker image aswell.

images/parity_parity/src/image.rs
D4nte: We don't want to make it an enum because it's too brittle. If they add a new config type a user of this will be able to use that new config value. It's better to just leave it as a stirng. 

On the other hand we made jsonrpc_apis a Vec<String> since it is a list of strings separated by `,`.

thomaseizinger: Well, the purpose of the testcontainers library is to facilitate testing. It is not intended to be used to run a node against mainnet. Thus I would make an enum here with all the Config values that currently make sense. If that is only `dev` for now, it should only be possible to put that value there.

PR #24, opened 2018-10-09
Add documentation for most public items

### LINKED ISSUE #6: Improve documentation
The documentation of the crate is currently in a very early stage. The following things are missing:

- Proper Rust docs
- Link to the README on crates.io (`readme` key in Cargo.toml)
- Better examples

thomaseizinger:
This commit also directly increments all the necessary versions for
the next release.

As I was writing the documentation for the main traits + structs, I
realized that they actually expose more things publicly than needed.

Therefore, this commit also includes a breaking change for the tc_core
crate because some of the public methods are removed. For users of the
testcontainers crate, this shouldn't be noticeable though since from the
PoV of testcontainers, it is a non-breaking change and thus a
patch-release.

Fixes #6.

### DIFF
--- a/core/src/container.rs
+++ b/core/src/container.rs
@@ -1,7 +1,27 @@
-use std::{collections::HashMap, env::var, io::Read};
+use std::env::var;
 use Docker;
 use Image;
+use Logs;
 
+/// Represents a running docker container.
+///
+/// Containers have a [`custom destructor`][drop_impl] that removes them as soon as they go out of scope:
+///
+/// ```rust
+/// #[test]
+/// fn a_test() {
+///     let docker = Cli::default();
+///
+///     {
+///         let container = docker.run(MyImage::default());
+///
+///         // Docker container is stopped/removed at the end of this scope.
+///     }
+/// }
+///
+/// ```
+///
+/// [drop_impl]: struct.Container.html#impl-Drop
 #[derive(Debug)]
 pub struct Container<'d, D, I>
 where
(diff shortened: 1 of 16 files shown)


core/src/container.rs
```diff
@@ -59,31 +95,42 @@ where
...
+    fn rm(&self) {
         debug!("Deleting docker container {}", self.id);
 
         self.docker_client.rm(&self.id)
     }
 }
 
+/// The destructor implementation for a Container.
+///
+/// As soon as the container goes out of scope, the destructor will either only stop or completely the docker container.
```
luckysori: Missing word: "the destructor will only stop or completely *remove the docker container"

core/src/image.rs
```diff
@@ -1,16 +1,55 @@
...
+    ///
+    /// This should return a full-qualified descriptor.
+    /// Implementations are encouraged to include a tag that will not change (i.e. NOT latest)
+    /// in order to prevent test code from randomly breaking because the underlying docker
+    /// suddenly changed.
     fn descriptor(&self) -> String;
+
+    /// Blocks the current thread until the started container is ready.
+    ///
+    /// This method is the **bread and butter** of the whole testcontainers library. Containers are
```
bonomat:
```suggestion
    /// This method is the **🍞 and butter** of the whole testcontainers library. Containers are
```
🤣

core/src/container.rs
```diff
@@ -59,31 +95,42 @@ where
         resolved_port
     }
 
-    pub fn block_until_ready(&self) {
+    /// Returns a reference to the [`Image`] of this container.
+    ///
+    /// Access to this is useful if the [`arguments`] of the [`Image`] change how to connect to the
+    /// container (for example, because of authentication information).
```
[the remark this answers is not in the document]
bonomat: Suggestion in this case seems not to be working due to multiline changes... but I'll give it a try anyway :) ...

How about:
```suggestion
    /// Access to this is useful to retrieve [`Image`] specific information such as authentication details or other relevant information which have been passed as [`arguments`]
```

PR #26, opened 2018-10-19
added zmq support

evgeniy-scherbina:
I think it will be convenient if bitcoind's image will have zmq support. Of course you can select another docker image or copy my in yours docker organization.


thomaseizinger:
I am gonna close this due to inactivity.
Btw: our docker image is now also compiled with `zmq`. If you need any of the proposed configuration options, open an issue and we can discuss how to implement them!

Issue #17, opened 2018-09-20
Increase sleep period when waiting for logs to be ready


thomaseizinger:
At the moment we wait for 100ms for the logs to be ready. Apparently, this is not enough as the tests on || keep failing randomly.

https://github.com/coblox/testcontainers-rs/blob/814ee6da67c6231990aa09070b6790100f848378/testcontainers/src/docker_cli.rs#L50-L68

Increasing the timeout to something like 500ms should not hurt that much because the container starts up in the meantime anyway. It is basically just a matter of where we spend the time waiting: In the loop that polls the logs or in the sleep before that.

D4nte:
@thomaseizinger to check if done

thomaseizinger:
Done in https://github.com/testcontainers/testcontainers-rs/commit/b6eed5d920f10d4637186aaebcc7384f68f5685e.

PR #27, opened 2018-10-23
Conditionally wait longer before fetching logs

thomaseizinger:
We wait for at least one second before fetching the logs from a
container in order to avoid the problem of not seeing logs when we try
to get them immediately after starting the container.

Fixex #17.

### DIFF
--- a/cli_client/src/cli.rs
+++ b/cli_client/src/cli.rs
@@ -1,5 +1,7 @@
 use serde_json;
 use std::collections::HashMap;
+use std::sync::RwLock;
+use std::time::Instant;
 use std::{
     io::{BufRead, BufReader},
     process::{Command, Stdio},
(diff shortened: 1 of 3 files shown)


cli_client/src/cli.rs
```diff
@@ -8,11 +10,68 @@ use std::{
...
+            Ok(lock_guard) => lock_guard,
+
+            // We only need the mutex to not require a &mut self in this function.
+            // Data cannot be in-consistent even if a thread panics while holding the lock
+            Err(e) => e.into_inner(),
+        };
+        let start_timestamp = Instant::now();
+
+        trace!(
+            "Registring starting of container {} at {:?}",
```
bonomat:
```suggestion
            "Registering starting of container {} at {:?}",
```
small typo

cli_client/src/cli.rs
```diff
@@ -8,11 +10,68 @@ use std::{
...
+
+        result
+    }
+
+    fn wait_at_least_one_second_after_container_was_started(&self, id: &str) {
+        if let Some(duration) = self.time_since_container_was_started(id) {
+            if duration < ONE_SECOND {
+                sleep(ONE_SECOND.checked_sub(duration).unwrap_or_else(|| ZERO))
+            }
+        }
```
bonomat: can `time_since_container_was_started` return `None`?  If so, what went wrong and do you still want to sleep here?

thomaseizinger: It returns None if the `id` is not in the hashmap, which basically means we did not start this container.

If that is the case, we probably don't want to sleep.

PR #37, opened 2018-12-13
Adding support for the Amazon DynamoDB local container

ayax79:


ayax79:
It appears all tests passed with the travis-ci build, but the build failed. The only error I see in the logs is:
`error: binary `cargo-tarpaulin` already exists in destination as part of `cargo-tarpaulin v0.6.10``

thomaseizinger:
The problem is the formatting! :)
We check that the code is formatted with `cargo fmt -- --check` in CI.

`tarpaulin` is for measuring test coverage and is run after the build. Even if it would fail, that would not impact the build status.

images/dynamodb_local/src/image.rs
```diff
@@ -0,0 +1,85 @@
+use std::{env::var, thread::sleep, time::Duration};
+use tc_core::{Image, Container, Docker, WaitForMessage};
+
+const ADDITIONAL_SLEEP_PERIOD: &'static str = "DYNAMODB_ADDITIONAL_SLEEP_PERIOD";
+const DEFAULT_WAIT: u64 = 2000;
+const CONTAINER_IDENTIFIER: &'static str = "amazon/dynamodb-local";
+const DEFAULT_TAG: &'static str = "latest";
```
thomaseizinger: Using latest can lead to problems with builds suddenly breaking. Bear in mind that the library is used for testing. You don't want your tests to break because of the test-setup :)

images/dynamodb_local/src/image.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    fn wait_until_ready<D: Docker>(&self, container: &Container<D, Self>) {
+        container
+            .logs()
+            .stdout
+            .wait_for_message("Initializing DynamoDB Local with the following configuration")
+            .unwrap();
+
+        let additional_sleep_period = var(ADDITIONAL_SLEEP_PERIOD)
```
thomaseizinger: I see you got inspired by the `bitcoind` image :)
Is the additional sleep period actually necessary for this image or is the container ready after the above log message?
If yes, please remove this part of the code, we try to have as minimal configuration as possible.

images/dynamodb_local/src/image.rs
```diff
@@ -0,0 +1,85 @@
...
+    fn with_args(self, arguments: <Self as Image>::Args) -> Self {
+        DynamoDb {
+            arguments, ..self
+        }
+    }
+
+}
+
+impl DynamoDb {
+    pub fn with_tag_args(self, tag_str: &str, arguments: <Self as Image>::Args) -> Self {
```
thomaseizinger: A method `with_tag` would be sufficient here since all the methods return `Self` and can thus be chained easily:

```rust
DynamoDb::default().with_tag("latest").with_args(...)
```

Note that `with_args` already exists.

testcontainers/tests/images.rs
```diff
@@ -74,3 +88,53 @@ fn trufflesuite_ganachecli_listaccounts() {
 
     assert_that(&accounts).is_ok();
 }
+
+#[test]
+fn dynamodb_local_create_table() {
```
thomaseizinger: Is creating a table the most simple command to check that the database is reachable? I am currently adding support for the `postgres` image and I simply execute a query to add two numbers:

https://github.com/testcontainers/testcontainers-rs/blob/a47584d8e75bb7d7fd568e689f2af9f420ccc9dd/testcontainers/tests/images.rs#L80-L98

images/dynamodb_local/src/image.rs
ayax79: My thoughts were that latest was a reasonable default. Callers could use the DynamoDB::with_tag_args function to pin an exact build, though probably not necessary. The use case with DynamoDB is interesting as DynamoDB local's latest will match what is running in AWS. DynamoDB Local provided by Amazon purely for testing.

images/dynamodb_local/src/image.rs
ayax79: I would like to move away from the sleep period completely at some point. I was too lazy to add something like check for the port to be listening. When Dynamo starts up it will output the following:
`Initializing DynamoDB Local with the following configuration:
Port:	8000
InMemory:	true
DbPath:	null
SharedDb:	false
shouldDelayTransientStatuses:	false
CorsParams:	*`

The only static line is the first -- which isn't really that ideal. 

Originally I set the sleep time to 1 sec. Which seemed adequate most the time, but would occasionally fail. I increased it to 2 seconds, and it doesn't seem to have an issue on my brand new MBP. My fear is that I if I remove any ability override, someone else's hardware isn't going perform the same and it will always fear. So long term, I should probably just come  up with a better solution.

images/dynamodb_local/src/image.rs
ayax79: Good call!

testcontainers/tests/images.rs
ayax79: DynamoDB being a fairly simple "NoSQL" Db doesn't provide a lot of the niceties that you find in a full featured RDBMS like PostgreSQL. Originally, my plan was to do a ListTables command and return on successful completion. Rusoto didn't seem to like getting an empty result back as there aren't any tables yet (though maybe I should dig in a little more at some point). With that failing, the only other option was to just create a table. Creating a table is a good test anyways, as the first thing anybody is going to want to do in their test is to create a table, then perhaps populate some test data.

The reason I added a the testcontainer to begin with was to play around with AWS's new lambda support and see how easy it was to integrate rusoto. See: https://github.com/ayax79/hello_rust_lambda

images/dynamodb_local/src/image.rs
thomaseizinger: Interesting, I guess in this case `latest` is fine.

images/dynamodb_local/src/image.rs
thomaseizinger: Nah, things should be simple. No need for port checking :)
One interesting thing to mention is that the code behind `wait_for_message` uses `contains` so you can match on partial lines if the whole line is not static.

I see your point with the sleep period. I had to add it to the bitcoind image for the exact same shitty reason. It's a mess :D

Maybe we can build an abstraction at some point to unify that.

testcontainers/tests/images.rs
thomaseizinger: Fair enough :)

It does test that it works so I guess it's ok.

thomaseizinger:
Thanks for the contribution!

I'll release it in the upcoming days :)

D4nte:
@ayax79 please note your crate is now published https://crates.io/crates/tc_dynamodb_local

PR #38, opened 2018-12-14
Add a postgres image

thomaseizinger:
This PR adds support for a postgres image.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
     "images/elasticmq",
     "images/generic",
     "images/parity_parity",
+    "images/postgres",
     "images/redis",
     "images/trufflesuite_ganachecli",
     "testcontainers",
--- a/images/postgres/Cargo.toml
+++ b/images/postgres/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "tc_postgres"
 version = "0.1.0"
-authors = [ "CoBloX developers <team@coblox.tech>" ]
+authors = [ "Thomas Eizinger <thomas@eizinger.io>" ]
 description = "Testcontainers image for the postgres docker image."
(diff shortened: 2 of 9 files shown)


images/postgres/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "tc_postgres"
+version = "0.1.0"
+authors = [ "CoBloX developers <team@coblox.tech>" ]
+description = "Testcontainers image for the parity docker image."
```
thomaseizinger:
```suggestion
description = "Testcontainers image for the postgres docker image."
```

images/postgres/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "tc_postgres"
+version = "0.1.0"
+authors = [ "CoBloX developers <team@coblox.tech>" ]
```
D4nte: Not sure this is valid as we don't use postgres (do we ?)

thomaseizinger: Well the field doesn't say `users` but `authors` :)

Unless what you want to say is: I don't want to maintain this!

D4nte: I am saying I don't think CoBloX should maintain this.

thomaseizinger: Fair enough. I'll add my name then :)

thomaseizinger: Done.

Cargo.toml
```diff
@@ -6,6 +6,8 @@ members = [
     "images/dynamodb_local",
     "images/elasticmq",
     "images/parity_parity",
+    "images/postgres",
+    "images/trufflesuite_ganachecli",
```
D4nte:
```suggestion

```

Issue #56, opened 2019-01-30
Provide a generic image


ufoscout:
It would be good to have an equivalent of the [GenericContainer](https://www.testcontainers.org/features/creating_container/) Java class to easily use custom images.

In fact, at the moment, for using an image not already provided by testcontainers-rs, it is required to create a new implementation of the [Image](https://github.com/testcontainers/testcontainers-rs/blob/master/core/src/image.rs) trait which involves lots of boilerplate.

Is there any plan to add it?

thomaseizinger:
> Is there any plan to add it?

At the moment no because we don't have a need for it right now.

> It would be good to have an equivalent of the GenericContainer Java class to easily use custom images.

Would containers created in this way have any "wait strategy" attached to them? How do you know when they are ready to be interacted with? From the docs of testcontainers-java, I assume this feature is then not available?

I guess we could start out with a simple generic container and add parametrized wait strategies later on.

ufoscout:
> Would containers created in this way have any "wait strategy" attached to them? How do you know when they are ready to be interacted with? From the docs of testcontainers-java, I assume this feature is then not available?

Yes, they have a DSL that permits to define everything, including a wait strategy. For example, this is the way I create a custom Postgres image:
```java
public GenericContainer containerWithWaitStrategy = new GenericContainer("postgres:9.6-alpine")
                    .withEnv("POSTGRES_PASSWORD", "postgres")
                    .withEnv("POSTGRES_USER", "postgres")
                    .withExposedPorts(5432)
                    .waitingFor(
                            Wait.forListeningPort()
                           // OR: Wait.forLogMessage(".*ready to accept connections*\\s", 1)
                    );
```

D4nte:
Closed in #58

PR #58, opened 2019-01-31
Add generic image

ufoscout:
This is an attempt to solve  #56

It introduces a GenericImage implementation and a generic way of handling WaitFor strategies.

### DIFF
--- a/images/generic/README.md
+++ b/images/generic/README.md
@@ -5,17 +5,4 @@ This crate provides a generic `Image` for `testcontainers`.
 The `GenericImage` allows to use of whatever Docker Image for which a specific implementation
 is not provided by `testcontainers`. 
 
-E.g.:
-```rust
-// Creates a generic image with descriptor "postgres:9.6-alpine"
-let generic_postgres = images::generic::GenericImage::new("postgres:9.6-alpine")
-    // Instructs testcontainers to wait until the LogMessage
-    // "database system is ready to accept connections" is printed in the
-    // container system output
-    .with_wait_for(images::generic::WaitFor::LogMessage(
-        "database system is ready to accept connections".to_owned(),
-    ));
-
-// Runs the it
-let node = docker.run(generic_postgres);
-```
+Check [the integration tests](./testcontainers/tests) on how to use this image.
--- a/images/generic/README.md
+++ b/images/generic/README.md
@@ -1,4 +1,4 @@
-# Postgres
+# Generic Image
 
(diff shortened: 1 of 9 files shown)


images/generic/src/image.rs
```diff
@@ -0,0 +1,68 @@
...
+pub enum WaitFor {
+    Nothing,
+    LogMessage(String),
+}
+
+impl WaitFor {
+    fn wait<D: Docker, I: Image>(&self, container: &Container<D, I>) -> Result<(), WaitError> {
+        match self {
+            WaitFor::Nothing => Ok(()),
+            WaitFor::LogMessage(message) => container.logs().stderr.wait_for_message(message),
```
thomaseizinger: Not sure if hardcoding `stderr` here is a good idea.

The other thing is, a user would currently have to call `to_owned` on a static string but most times, what they will pass is a `&'static str`.

Maybe we can kill two birds with one stone here and create factory functions on `WaitFor` to initialize it from a `&str` and at the same time configure a flag, which stream the message is expected on.

```rust
impl WaitFor {

    fn message_on_stdout(message: &str) -> Self {
        Self::LogMessage {
            message: message.to_owned(),
            stream: Stream::StdOut
        }
    }


    fn message_on_stderr(message: &str) -> Self {
        Self::LogMessage {
            message: message.to_owned(),
            stream: Stream::StdErr
        }
    }
}
```

ufoscout: @thomaseizinger  
Fixed

ufoscout:
@thomaseizinger 
Comments fixed. Let me know if it's fine.

images/generic/README.md
```diff
@@ -0,0 +1,21 @@
...
+is not provided by `testcontainers`. 
+
+E.g.:
+```rust
+// Creates a generic image with descriptor "postgres:9.6-alpine"
+let generic_postgres = images::generic::GenericImage::new("postgres:9.6-alpine")
+    // Instructs testcontainers to wait until the LogMessage
+    // "database system is ready to accept connections" is printed in the
+    // container system output
+    .with_wait_for(images::generic::WaitFor::LogMessage(
```
thomaseizinger: The documentation is now outdated :sweat_smile: 

I'd say remove the whole block in favor of a link to the integration tests or rust-doc.

ufoscout:
@thomaseizinger 
Outdated code replaced with a link to the integration tests.

PR #60, opened 2019-02-05
Update parity image to 2.2.9

luckysori:


luckysori:
Should not modify the default image version. Just pass in a new version using `with_tag`.

PR #66, opened 2019-02-10
Add environment variables

### LINKED ISSUE #57: Define environment variables
How can I define environment variables when starting an image?

Many images can be customized through env vars, for example, Postgres permits to define user, password db name, db init arguments and so on:
```bash
> docker run -e POSTGRES_PASSWORD=mysecretpassword -d postgres
```

For what I see in the Cli ([here](https://github.com/testcontainers/testcontainers-rs/blob/cc8fdbe1645c2d13c6eb74be600e4198316f1b63/cli_client/src/cli.rs#L80)) this is not possible with the current implementation.

ufoscout:
Attempt to fix #57

This PR introduces:
- a new method to return env vars added to the Image trait
- environment variables handling added in the Cli
- a new method to declare custom env vars added to the GenericImage

It does not provide any solution for the non-generic images; a per-image specific solution is outside the scope of this PR.

### DIFF
--- a/images/generic/src/image.rs
+++ b/images/generic/src/image.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use tc_core::{Container, Docker, Image, WaitError, WaitForMessage};
 
 #[derive(Debug, PartialEq, Clone)]
@@ -42,6 +43,7 @@ impl WaitFor {
 pub struct GenericImage {
     descriptor: String,
     arguments: Vec<String>,
+    env_vars: HashMap<String, String>,
     wait_for: WaitFor,
 }
 
@@ -50,6 +52,7 @@ impl Default for GenericImage {
         Self {
             descriptor: "".to_owned(),
             arguments: vec![],
+            env_vars: HashMap::new(),
             wait_for: WaitFor::Nothing,
         }
     }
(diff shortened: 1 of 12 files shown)


images/generic/src/image.rs
```diff
@@ -60,10 +63,16 @@ impl GenericImage {
         Self {
             descriptor: descriptor.into(),
             arguments: vec![],
+            env_vars: HashMap::new(),
             wait_for: WaitFor::Nothing,
         }
     }
 
+    pub fn with_env_var<S: Into<String>>(mut self, key: S, value: S) -> Self {
```
thomaseizinger: I think we should have two parameters here otherwise, you cannot do something like:

`with_env_var("my_var", a_variable)`

if `a_variable` is of type `String` because `S` will already resolve to `&'static str` from the `key` parameter.

thomaseizinger:
Also: To avoid `Fix format` commits in the future, you can activate the `pre-commit` hook that comes with this repository. Have a look here:

https://github.com/testcontainers/testcontainers-rs#developing

images/generic/src/image.rs
ufoscout: good catch! Fixed!

ufoscout:
@thomaseizinger 
I implemented your comments and rebased to avoid useless commits.

thomaseizinger:
Awesome! Thanks a lot :)

PR #87, opened 2019-05-05
Upgrade parity container

bonomat:


.travis.yml
```diff
@@ -14,7 +14,7 @@ services:
 before_script:
```
thomaseizinger: I think we might be able to just remove all these at some point. The duplication is probably not worth it.

D4nte: Yes this is the same discussion than we had for comit-rs.

bonomat: Sorry, did not see this in time but created an issue for it.

PR #120, opened 2019-10-03
Fix Rust version to 1.32.0 for MSRV and 1.34.0 for the tests

### LINKED ISSUE #119: Pin rust version in `rust-toolchain` file to MSRV
## Problem

Currently, our `rust-toolchain` file says `stable`.
This can cause your CI to fail because `stable` is a moving target and components like rustfmt might have different opinions about how code should be formatted for different versions.

## Goal

1. Have a stable CI
2. State a MSRV (minimum supported rust version) for other users

## Recommendation

Figure out what the MSRV for our project is an pin it to that.

thomaseizinger:
Fixes #119.


thomaseizinger:
cc @D4nte @luckysori had to make some more changes for this to actually work.

PR #132, opened 2019-11-06
Replace web3 with reqwest

thomaseizinger:
Web3 broke our build, let's replace it with something that doesn't break.


Cargo.toml
```diff
@@ -29,4 +29,5 @@ rusoto_credential = "0.40"
 rusoto_dynamodb = "0.40"
 rusoto_sqs = "0.40"
 spectral = "0.6"
-web3 = "0.8"
+web30 = "0.4"
+futures = "0.1"
```
D4nte: `cargo tomlfmt`?

Issue #129, opened 2019-10-23
Ability to run tests inside a docker container themselves


wngr:
Hey,

the current approach of this library assumes that `cargo test` is executed on the host system running docker. For having a reproducible environment, it makes sense to run the tests themselves inside a docker container. To be able to have those tests spin up and access other containers, we'd need to:
1. Be able to access the docker socket from within the container (and have `docker-cli` installed)
1. Make sure the networking is properly set up, e.g. same docker network / bridging
1. Ability to get the IP of the created container --> `docker inspect` has it

My proposed approach would be to create another `Client` and implement _2_ and _3_ as enumerated above. Obviously, to be able to use it, the library users needs to make sure _1_ is properly set up.

What do you think about this approach? Would you be willing to accept a PR implementing the above?

D4nte:
Maybe look into the shiplift crate so you don't create a docker api client from scratch.

thomaseizinger:
> What do you think about this approach? Would you be willing to accept a PR implementing the above?

Yes definitely!
That is the direction we would want to take this library in, we just didn't have any needs for it so far :)
We would definitely accept a PR for this!

wngr:
> Maybe look into the shiplift crate so you don't create a docker api client from scratch.

Thanks for the hint, will give it a spin then!

DDtKey:
I think this should be possible with current state of `testcontainers`. We don't use `cli` client anymore and there is a [detection of host inside of container](https://github.com/testcontainers/testcontainers-rs/blob/8f4279fcef8279ca1ac2bf9fdb8793c28973347d/testcontainers/src/core/client.rs#L368-L387)

So I'm going to close the issue because of outdated state, but if you feel it still doesn't work, feel free to re-open or create a new one.

Thank you!

PR #139, opened 2019-12-05
Add methods to get and check host address to Container

4meta5:
I received direction from @bsideup but this PR is still a work-in-progress. I've bumped the versions and added the `url` crate as a dependency to parse the key for the `DOCKER_HOST` process (see `Container::set_host` ). 

## why

The point is to add `set_host` such that the `Container` object is initialized with a new field corresponding to the host address. 

Without this logic, there is no guarantee that the host address is not `localhost`, and its neglect can lead to careless mistakes that [may compromise security](https://stackoverflow.com/a/49206688).

The reason the `host` field is added to the `Container` object is to save future lookups (acting like a cache).


src/core/container.rs
```diff
@@ -54,6 +91,34 @@ where
...
+            let host_url = Url::parse(&host).expect("failed to parse url");
+            match host_url.scheme() {
+                "https" | "http" | "tcp" => {
+                    return host_url.host_str().unwrap().to_string();
+                }
+                _ => (),
+            }
+        }
+        if Path::new("/.dockerenv").exists() {
+            let container = docker_client.run(Alpine);
```
bsideup: TODO: this should be globally cached to avoid `docker run` for each new container

bsideup:
FYI this PR follows this class from the Java version:
https://github.com/testcontainers/testcontainers-java/blob/1.12.4/core/src/main/java/org/testcontainers/dockerclient/DockerClientConfigUtils.java

bsideup:
@4meta5 make sure you also change the tests. Also, consider adding a step on Travis that runs the tests inside a container

thomaseizinger:
Hi @4meta5,

thanks for working on this!
Would you mind elaborating on the design a bit more? It seems counter-intuitive to me that this functionality is on the `Container` struct. I would have rather seen it on the `Cli` struct as that one is responsible for the communication with the docker host.

The `Container` instances are created at the end of `run`, hence whatever the value of `DOCKER_HOST` is could be set in the constructor of `Container` then.

Additionally, if all that functionality is inside the `Cli` struct, I believe you don't have to abuse the `Image` trait to create a private, purpose-specific `Alpine` image to do that work because you have access to the internals of `Cli` (some refactoring may be required there to get access to everything you need).

Accoring to [this](https://superuser.com/a/1021925/376882) SO answer, the use of `.dockerenv` is deprecated. I've never dealt with it, do we need to build support for this?

bsideup:
@thomaseizinger

_Disclaimer: I am one of the http://github.com/testcontainers/testcontainers-java maintainers._

>  It seems counter-intuitive to me that this functionality is on the Container struct

Although it is the same for every container at the moment, it may change in future (e.g. direct container-to-container communication), and it is aligned with what we have in the Java version of Testcontainers (see `getContainerIpAddress()`)

> I've never dealt with it, do we need to build support for this?

We do, since otherwise there is no other reliable way of detecting the in-Docker mode, in which we have to use the gateway's IP because `localhost` points to container's localhost.

> According to this SO answer, the use of .dockerenv is deprecated

The SO answer is talking mostly about the init binary and its deprecation since LXC is no longer in use.

The file is there:
https://github.com/moby/moby/blob/b3e9f7b13b0f0c414fa6253e1f17a86b2cff68b5/daemon/initlayer/setup_unix.go#L29

And some Docker's internals also rely on the presence of this file:
https://github.com/docker/docker-install/blob/36b78b2c802c0fc8cc997af8123acba384239916/verify-docker-install#L14
https://github.com/docker/libnetwork/blob/d0ae17dcfaa1f21e3b0f5d55bba4239f08489640/drivers/bridge/setup_bridgenetfiltering.go#L160

Although if you know a better approach to detect the in-container mode, we will be more than happy to use it instead :)

bsideup:
FYI it is also possible to use `docker network inspect bridge` instead of running a container with `ip|awk`:
https://github.com/testcontainers/testcontainers-java/pull/2151

4meta5:
Just getting back to this now. For context, the intent of this PR is to provide methods to detect the in-container mode by introducing methods to set and get the host address on the `Container`. I still need to
- [ ] change existing tests based on new `Container` code
- [ ] update travis to add a step to run the tests in a container

I'll also use `docker network inspect bridge` instead of running `ip | awk` because I think it is more readable. @thomaseizinger does this sound good?

bsideup:
> I'm still a bit confused about how to test that this works in-container

you need to.... run the tests in-container ;)
```shell
$ docker run -it --rm -v $PWD:$PWD -w $PWD -v /var/run/docker.sock:/var/run/docker.sock -v $HOME/.cargo:/root/.cargo rust cargo test
```

src/core/container.rs
4meta5: This should be done with [`lazy_static`](https://docs.rs/lazy_static/1.2.0/lazy_static/) but I am unable to call the trait method `run` on a generic docker container inside of this macro.

4meta5:
> you need to.... run the tests in-container ;)

Ok, I added the command from this comment as a step to the travis build to run the tests in-container. All the tests pass and the code compiles on my local machine.

I added a TODO to cache the `docker_container.run(Alpine)` command with [`lazy_static`](https://docs.rs/lazy_static/1.4.0/lazy_static/) but I cannot figure out how to call a trait method within this macro.

@thomaseizinger I'd like to know what should be done before this can be merged.

.travis.yml
```diff
@@ -19,6 +19,7 @@ matrix:
         - cargo fmt -- --check
         - cargo build --all-targets
         - cargo test
+        - docker run -it --rm -v $PWD:$PWD -w $PWD -v /var/run/docker.sock:/var/run/docker.sock -v $HOME/.cargo:/root/.cargo rust cargo test
```
4meta5: I feel like this change just made the CI build fail. Maybe it doesn't belong here or maybe I should place it elsewhere.

thomaseizinger:
Sorry for getting back to this so late.

> @thomaseizinger I'd like to know what should be done before this can be merged.

I am confused about what we are actually solving here. You've mentioned that the original motivating issue was https://github.com/testcontainers/testcontainers-rs/issues/129?

On the issue itself, we've already discussed an implementation idea. I am not sure if @wngr already started on implementing something here.

Regarding the proposed changes, I'd like to discuss, why this functionality needs to be on the container struct.

@bsideup can you elaborate more on the following sentence:

> Although it is the same for every container at the moment, it may change in future (e.g. direct container-to-container communication).

Container-to-container communication works based on docker networks right? How is the host address related to this?
I thought docker containers can access each other through the convention of `{container_name}:{internal_port}`?

Why is it not enough if we pass the value of DOCKER_HOST into the container (or default to 127.0.0.1 if it is not set)?

To me it sounds like the functionality of determining the host address of a container is functionality that should be a concern of the docker client and not the container itself. The rest of the library follows the pattern of:
1) exposing certain functionality on the `Docker` trait (start, stop, delete, etc).
2) calling these APIs from within the `Container` struct to create a more ergonomic experience

Following that, we would need new functionality on the `Docker` trait a la:

```
fn host_address(&self, id: &str)
```

We can then call this from within the `Container` struct where we pass the id of the container in.

Implementing the main functionality within the specific client implementation should allow you to re-use certain pieces (like running a container), without needing to (ab)-use the `Image` trait for this. This should also remove the need for even having the `Alpine` struct.

wngr:
I actually took a start at this last week. The approach I followed was to add getting the ip address per container `id` to the Docker trait, so pretty much as @thomaseizinger suggested :
```diff
@@ -167,12 +191,22 @@ impl Docker for Cli {
             .spawn()
             .expect("Failed to execute docker command");
     }
+    fn ip_address(&self, id: &str) -> String {
+        if Self::running_inside_docker() {
+            let info = self.get_container_info(id);
+            info.network_settings.ip_address
+        } else {
+            "127.0.0.1".to_string()
+        }
+    }
 }
```
This is fed by the output of `docker inspect`. Some care needs to be taken re mounting volumes (which was recently added to this crate). Furthermore, it would be nice to make sure that the spawned containers are actually on the same network as the container, that is running the test and thus is actually orchestrating the test containers' lifecycle (which I haven't done yet).

I suppose the rationale behind adding that to the `Container` trait is just a different API idea behind `testcontainers-java`, which seems a bit different from `testcontainers-rs`.

-------

BTW, to figure out whether the code is being executed in a container or not, I used the following:
```diff
diff --git a/src/clients/cli.rs b/src/clients/cli.rs
index 7f129aa..d4965b1 100644
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -28,6 +28,13 @@ pub struct Cli {
 }
 
 impl Cli {
+    fn running_inside_docker() -> bool {
+        // https://stackoverflow.com/questions/20010199/how-to-determine-if-a-process-runs-inside-lxc-docker#comment100259376_20012536
+        std::path::Path::new("/.docker-env").exists()
+            || std::fs::read_to_string("/proc/1/cgroup")
+                .map(|cgroup| cgroup.contains("docker") || cgroup.contains("lxc"))
+                .unwrap_or(false)
+    }
     fn register_container_started(&self, id: String) {

```

4meta5:
Sounds good, I like your approach and am comfortable closing this PR in favor of @wngr 's way of doing things. Thanks for explaining how things are organized in `testcontainers-rs` :)

Issue #140, opened 2019-12-08
Debugging Travis Build Breaking


4meta5:
I can't make sense of why the build in https://github.com/testcontainers/testcontainers-rs/pull/139 is breaking. Here is the first error when `cargo build` is run:

```rust
$ cargo build

    Updating crates.io index

   Compiling smallvec v1.0.0

   Compiling percent-encoding v2.1.0

error[E0658]: use of unstable library feature 'alloc': this library is unlikely to be stabilized in its current form or name (see issue #27783)

  --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/smallvec-1.0.0/lib.rs:38:1

   |

38 | extern crate alloc;

   | ^^^^^^^^^^^^^^^^^^^
...
```

Note that this branch compiles and all tests passed when `cargo build` and `cargo test` are run on my local machine

D4nte:
This is most likely due to the difference of rust version used.
The `rust-toolchain` file contains `stable`

We can see in the CI:
```
$ curl -sSf https://build.travis-ci.org/files/rustup-init.sh | sh -s -- --default-toolchain=$TRAVIS_RUST_VERSION --profile=minimal -y
...
$ rustc --version
rustc 1.32.0 (9fda7c223 2019-01-16)
$ rustup --version
rustup 1.20.2 (13979c968 2019-10-16)
$ cargo --version
cargo 1.32.0 (8610973aa 2019-01-02)
```

So it looks like Travis is installing an older version of Rust. Latest stable being 1.39.0.

Let me delete Travis cache and see if it helps.

D4nte:
@thomaseizinger do you know why travis is running two builds? one for 1.32.0 and one for 1.36.0? 

https://travis-ci.org/testcontainers/testcontainers-rs/builds/622306051?utm_source=github_status&utm_medium=notification

4meta5:
added a commit to #139 -- this can be closed once/if that PR is merged

thomaseizinger:
> The current (undocumented) MSRV (Minimal Supported Rust Version) is 1.32.0:
> 
> https://github.com/testcontainers/testcontainers-rs/blob/b6f9dbe82478f28f5c5b46686bcc4dfe422fd9ea/.travis.yml#L4
> 
> From my point of view, it is if fine to change it to 1.36.0 (current stable is 1.39.0) so feel free to do a update to update the Travis config.

Yes it is fine to change that :)
[...]

PR #141, opened 2019-12-13
Allow tests to be run inside a docker container

wngr:
This adds the `ip_address` method to `Docker` trait in order to to query the ip address of the spawned container. This is derived from a `docker inspect` call.

This also introduces a new usage pattern when used from the host system: 
Rather than relying on a port mapping to the host, the presumably known target port can be directly queried by using the ip address of the container in the internal docker network.
Let me know whether this actually breaks any workflows I didn't think of.
Note: This relies on the fact, that the default docker bridge network is configured for all involved containers.

I looked into shiplift to create a new `Client` impl, but refrained from doing so, as the current API is completely sync, and it felt rather heavy pulling in shiplift w/ a runtime. However, I'm very much interested in a follow-up discussion, where this crate could offer an additional async API, or a way forward with the sync API.
Futhermore, with the approach presented in this PR, the orchestrating container needs to have the `docker` cli available. When adding more and more features, it might make sense eventually to use shiplift to interface with the docker daemon, and by this also get rid of this requirement..

Closes https://github.com/testcontainers/testcontainers-rs/pull/129
Related https://github.com/testcontainers/testcontainers-rs/pull/139

I also bumped the version in `.travis.yml` #140.

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -28,6 +28,14 @@ pub struct Cli {
 }
 
 impl Cli {
+    fn running_inside_docker() -> bool {
+        // https://stackoverflow.com/questions/20010199/how-to-determine-if-a-process-runs-inside-lxc-docker#comment100259376_20012536
+        std::path::Path::new("/.docker-env").exists()
+            || std::fs::read_to_string("/proc/1/cgroup")
+                .map(|cgroup| cgroup.contains("docker") || cgroup.contains("lxc"))
+                .unwrap_or(false)
+    }
+
     fn register_container_started(&self, id: String) {
         let mut lock_guard = match self.container_startup_timestamps.write() {
             Ok(lock_guard) => lock_guard,
@@ -78,8 +86,15 @@ impl Cli {
             command.arg("-e").arg(format!("{}={}", key, value));
         }
 
-        for (orig, dest) in image.volumes() {
-            command.arg("-v").arg(format!("{}:{}", orig, dest));
+        if Self::running_inside_docker()
+            && image.volumes().into_iter().collect::<Vec<_>>().len() > 0
+        {
+            // TODO: Panic here or just log and continue?
+            panic!("Running inside Docker: Unable to map volumes.");
+        } else {
+            for (orig, dest) in image.volumes() {
+                command.arg("-v").arg(format!("{}:{}", orig, dest));
+            }
         }
 
         command
@@ -88,6 +103,24 @@ impl Cli {
             .arg(image.descriptor())
             .args(image.args())
             .stdout(Stdio::piped())
+    }
+
+    fn get_container_info(&self, id: &str) -> ContainerInfo {
+        let child = Command::new("docker")
+            .arg("inspect")
+            .arg(id)
+            .stdout(Stdio::piped())
+            .spawn()
+            .expect("Failed to execute docker command");
+
+        let stdout = child.stdout.unwrap();
+
+        let mut infos: Vec<ContainerInfo> = serde_json::from_reader(stdout).unwrap();
+
+        let info = infos.remove(0);
+
+        log::trace!("Fetched container info: {:#?}", info);
+        info
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/clients/cli.rs
```diff
@@ -130,22 +163,14 @@ impl Docker for Cli {
...
-
-        info.network_settings.ports.into_ports()
+        let info = self.get_container_info(id);
+        let ports = info.network_settings.ports.into_ports();
+        if Self::running_inside_docker() {
+            // Return transparent port mapping here?!
+            ports
+        } else {
+            ports
+        }
```
wngr: Open question, whether this API should be changed in any way.

src/clients/cli.rs
```diff
@@ -78,8 +86,15 @@ impl Cli {
             command.arg("-e").arg(format!("{}={}", key, value));
         }
 
-        for (orig, dest) in image.volumes() {
-            command.arg("-v").arg(format!("{}:{}", orig, dest));
+        if Self::running_inside_docker()
+            && image.volumes().into_iter().collect::<Vec<_>>().len() > 0
+        {
+            // TODO: Panic here or just log and continue?
+            panic!("Running inside Docker: Unable to map volumes.");
```
wngr: How should this be handled?

bsideup: This is very restrictive. It is still possible to map volumes if they mapped correctly

src/clients/cli.rs
```diff
@@ -167,12 +192,24 @@ impl Docker for Cli {
...
     }
+    /// This transparently returns the ip address of the container `id`,
+    /// as reported by `docker inspect`.
+    /// Note: The user needs to make sure that the physical or virtual node,
+    /// e.g. wherever the orchestrating code is running, is able to route
+    /// to this IP address. In general, this is achieved by relying on the
+    /// default docker bridge network.
+    fn ip_address(&self, id: &str) -> String {
+        let info = self.get_container_info(id);
+        info.network_settings.ip_address
```
bsideup: if container is connected to a network that is different from the tests' one, you will not be able to use this IP.

This is why in testcontainers-java we always connect to the gateway's IP

bsideup:
> Rather than relying on a port mapping to the host, the presumably known target port can be directly queried by using the ip address of the container in the internal docker network.
> Let me know whether this actually breaks any workflows I didn't think of.

From testcontainers-java experience I can say that this breaks in **many** environments. This is why in the previous PR ( #139 ) we did it differently (but the way it works in **every** case)

bsideup:
Also, I am a bit disappointed that you decided not not listen to the feedback based on testcontainers-java and will let you discover all the fun yourself :)

thomaseizinger:
> From testcontainers-java experience I can say that this breaks in many environments. This is why in the previous PR ( #139 ) we did it differently (but the way it works in every case)

@bsideup are you only talking about not using the port mapping or also determining the ip address via `docker inspect`?

thomaseizinger:
> Also, I am a bit disappointed that you decided not not listen to the feedback based on testcontainers-java and will let you discover all the fun yourself :)

Just to clarify, I am very open to allow `testcontainers-rs` to become as mature as `testcontainers-java`! My main concern in #139 was purely implementation-wise.

If the approach in #139 is superior, maybe we can take the best of both worlds and implement #139 with a design similar to this PR?

I am unfortunately not too familiar with the internals of docker, hence I don't exactly know the constraints we are facing here. Would it be too much to ask to create a structured write-up of what our constraints are for implementing this?

thomaseizinger:
> I looked into shiplift to create a new Client impl, but refrained from doing so, as the current API is completely sync, and it felt rather heavy pulling in shiplift w/ a runtime. However, I'm very much interested in a follow-up discussion, where this crate could offer an additional async API, or a way forward with the sync API.

@wngr That is a good point! Now that async/await is maturing in Rust and we have things like `#[tokio::test]` and `#[async_std::test]`, we can at some point migrate the API to be async which then should allow us to fairly easily implement a client on top of shiplift. There is also https://github.com/fussybeaver/bollard btw.

wngr:
Sorry for being passive on this PR, vacation days took its toll on me :-)..

-------
@bsideup
> Also, I am a bit disappointed that you decided not not listen to the feedback based on testcontainers-java and will let you discover all the fun yourself :)

I'm honestly confused now, whether #139 and the PR presented here are actually about the same thing. The scenario I want to realize with the proposed approach is to be able to run the actual testing code inside a docker container itself. Thus, I have to get the container's IP address. You rightfully noted, that this assumes that both containers are on the same network (TODO!).
I'd greatly appreciate if you could elaborate on what you mean by ..
> This is why in testcontainers-java we always connect to the gateway's IP
.. and how the scenario I described above can be done in a more robust way.

rkudryashov:
I have just encountered an issue that I think is related to this thread:

```
---- test_sign_in stdout ----
thread 'test_sign_in' panicked at 'Failed to execute docker command: Os { code: 2, kind: NotFound, message: "No such file or directory" }', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/testcontainers-0.9.1/src/clients/cli.rs:102:21
Panic in Arbiter thread.
```

It would be great to make this working! Are there any news on the PR?

thomaseizinger:
> It would be great to make this working! Are there any news on the PR?

Nothing other than what you can see on this thread unfortunately :)

From my perspective, I think efforts stalled mainly due to confusion of what usecases we actually want to support :grimacing: 

Like I said [above](https://github.com/testcontainers/testcontainers-rs/pull/141#issuecomment-569583378), it would be very helpful if we could have a write-up of which different usecases we want to solve and what the traps around this are. @bsideup mentioned that they discovered several of those in `testcontainers-java` :)

PR #154, opened 2020-03-24
Include dev branch in travis config

thomaseizinger:


thomaseizinger:
I will merge this regardless of the tests failing because we already have a PR that fixes the tests but it needs this PR to actually run the tests.

PR #159, opened 2020-03-26
Allow custom options on docker run

ufoscout:
This PR adds a `run_with_options` method that allows running a docker container with custom options instead of the currently hardcoded ones (i.e. "-d" "-P").
This permits, for example, to start a container on a predetermined port.

E.g.:
```rust
    let fixed_port = get_the_port_from_somewhere();

    let _node = docker.run_with_options(
        vec!["-d", "-p", &format!("{}:5432", fixed_port)],
        images::postgres::Postgres::default(),
    );
```

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -82,9 +82,18 @@ impl Cli {
             command.arg("-v").arg(format!("{}:{}", orig, dest));
         }
 
+        if let Some(ports) = image.ports() {
+            for port in &ports {
+                command
+                    .arg("-p")
+                    .arg(format!("{}:{}", port.local, port.internal));
+            }
+        } else {
+            command.arg("-P"); // expose all ports
+        }
+
         command
             .arg("-d") // Always run detached
-            .arg("-P") // Always expose all ports
             .arg(image.descriptor())
             .args(image.args())
             .stdout(Stdio::piped())
@@ -226,6 +235,8 @@ impl Ports {
 
 #[cfg(test)]
 mod tests {
+    use crate::core::Port;
+    use crate::images::generic::GenericImage;
     use crate::{Container, Docker, Image};
 
     use super::*;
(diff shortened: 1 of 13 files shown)


ufoscout:
@thomaseizinger 
I understand your point but not completely. The proposed method is in the `Docker` trait, isn't that trait explicitly for Docker as stated by its name?

In my case, I need to start an image with an exact port mapping because for some integration tests I cannot use a random port.

thomaseizinger:
> The proposed method is in the Docker trait, isn't that trait explicitly for Docker as stated by its name?

Yes it is for `Docker` but in "you can talk to the docker daemon through this" and not in "we call the CLI under the hood".

Docker runs as a daemon process on your machine and provides an HTTP API. The `docker` CLI just makes HTTP requests if you call it.

To (eventually) support scenarios where the docker daemon runs on a remote machine, we need to maintain an interface that is not specific to the docker CLI.

> In my case, I need to start an image with an exact port mapping because for some integration tests I cannot use a random port.

That is a use-case I am happy to support. I would suggest to add an API for specifying port mappings on a particular instance of an image before it is passed to the `Docker` trait impl.

ufoscout:
@thomaseizinger 
I added a method on the image itself. Let me know your thoughts.

thomaseizinger:
Cool, thanks! That looks good :)

For some weird reason, the CI build doesn't show up :thinking: 
I will try and investigate this on the weekend!

thomaseizinger:
Small status update: It seems that the CI integration with this repository is so old that it still uses the legacy "status" API of GitHub and not the new checks API. It was originally configured to use the old `travis-ci.org` hosting (for open-source repositories). For now I've requested access to the new GitHub app from the testcontainers organization admins.

In the long run, we could just migrate away from travis and use GitHub actions instead :)

**tl;dr:** Once the new integration with Travis is approved, the status checks should show up and we can merge this PR!

ufoscout:
@thomaseizinger any progress on this one?

thomaseizinger:
> @thomaseizinger any progress on this one?

Sorry for being so slow on this :S

Travis CI is really annoying unfortunately. I would like to migrate to GitHub actions but haven't found the time yet.

PR #162, opened 2020-05-02
Update postgres to 0.17

vladinator1000:
Wow, this project is awesome!

Here's a PR that updates the postgres dependency to match the latest one.

I ran the tests on my machine and everything but the dynamodb one passed, it looks unrelated though 🤷‍♀️

```
---- dynamodb_local_create_table stdout ----
thread 'dynamodb_local_create_table' panicked at '
        expected: result[ok]
         but was: result[error]<HttpDispatch(HttpDispatchError { message: "operation was canceled: connection closed before message completed" })>
```

### DIFF
--- a/tests/images.rs
+++ b/tests/images.rs
@@ -1,7 +1,6 @@
 use bitcoincore_rpc::RpcApi;
 use bson::{self, doc, Document};
 use mongodb::{self, Client};
-use postgres::{Connection, TlsMode};
 use redis::Commands;
 use rusoto_core::{HttpClient, Region};
 use rusoto_credential::StaticProvider;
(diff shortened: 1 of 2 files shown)


tests/images.rs
```diff
@@ -180,21 +180,21 @@ fn generic_image() {
...
+        db
+    );
+
+    let mut conn = Client::connect(connection_string, NoTls).unwrap();
+    let rows = conn.query("SELECT 1 + 1", &[]).unwrap();
+
+    for row in rows {
+        let first_column: i32 = row.get(0);
+        assert_eq!(first_column, 2);
+    }
```
thomaseizinger, line 197: If for some reason, we would be returning 0 rows, the test would not have any assertions. We are only returning a single row so I don't think a loop is needed here :)

vladinator1000: ahh good catch, will change that back

vladinator1000:
Fixed the row stuff, but this test fails consistently on rustc 1.45.0-nightly (7f65393b9 2020-05-01)
```
---- dynamodb_local_create_table stdout ----
thread 'dynamodb_local_create_table' panicked at '
        expected: result[ok]
         but was: result[error]<HttpDispatch(HttpDispatchError { message: "operation was canceled: connection closed before message completed" })>
```

PR #174, opened 2020-06-12
Entrypoint customization

tumdum:
Extend `Image` so that it is possible to customize `entrypoint` used by image.

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -306,7 +306,7 @@ mod tests {
         type Args = Vec<String>;
         type EnvVars = HashMap<String, String>;
         type Volumes = HashMap<String, String>;
-        type EntryPoint = void::Void;
+        type EntryPoint = std::convert::Infallible;
 
         fn descriptor(&self) -> String {
             String::from("hello-world")
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -241,7 +241,7 @@ impl Ports {
 mod tests {
     use crate::core::Port;
     use crate::images::generic::GenericImage;
-    use crate::{Container, Docker, Image};
+    use crate::{Container, Docker, Image, Void};
 
     use super::*;
 
@@ -306,6 +306,7 @@ mod tests {
         type Args = Vec<String>;
         type EnvVars = HashMap<String, String>;
         type Volumes = HashMap<String, String>;
+        type EntryPoint = Void;
 
         fn descriptor(&self) -> String {
             String::from("hello-world")
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -82,6 +82,10 @@ impl Cli {
             command.arg("-v").arg(format!("{}:{}", orig, dest));
         }
 
+        if let Some(entrypoint) = image.entrypoint() {
+            command.arg("--entrypoint").arg(entrypoint);
+        }
+
         if let Some(ports) = image.ports() {
             for port in &ports {
                 command
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -241,7 +241,7 @@ impl Ports {
 mod tests {
     use crate::core::Port;
     use crate::images::generic::GenericImage;
-    use crate::{Container, Docker, Image, Void};
+    use crate::{Container, Docker, Image};
 
     use super::*;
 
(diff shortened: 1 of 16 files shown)


tests/images.rs
```diff
@@ -223,6 +223,39 @@ fn generic_image() {
     assert_eq!(first_column, 2);
 }
 
+#[test]
+fn generic_image_with_custom_entrypoint() {
+    let docker = clients::Cli::default();
+    let msg = images::generic::WaitFor::message_on_stdout("server is ready");
+
+    let generic = images::generic::GenericImage::new("tumdum/simple_web_server:latest")
```
thomaseizinger, line 231: Is this image going to be supported for longer?

I wouldn't want to depend on an image for our tests that breaks or is removed eventually :)

It would be nice to depend on one of the standard images like alpine for these tests.

tests/images.rs
```diff
@@ -223,6 +223,39 @@ fn generic_image() {
...
+        "foo",
+        reqwest::blocking::get(&format!("http://127.0.0.1:{}", port))
+            .unwrap()
+            .text()
+            .unwrap()
+    );
+
+    let generic = images::generic::GenericImage::new("tumdum/simple_web_server:latest")
+        .with_wait_for(msg)
+        .with_entrypoint(&"/bar");
```
thomaseizinger, line 246: Having to pass a reference to a string slice is a bit odd. That is effectively a double reference. Any chance we can fix that?

src/clients/cli.rs
```diff
@@ -302,6 +306,7 @@ mod tests {
         type Args = Vec<String>;
         type EnvVars = HashMap<String, String>;
         type Volumes = HashMap<String, String>;
+        type EntryPoint = void::Void;
```
thomaseizinger, line 309: Do you mind using `std::convert::Infallible` here? From my understanding, that should at some point be aliased to the never type (`!`).

tests/images.rs
tumdum: I've created it just for this pull request, source is here: https://git.sr.ht/~ttt/simple_web_server

tests/images.rs
tumdum: That reference wasn't needed.

src/clients/cli.rs
tumdum: Ok

tests/images.rs
thomaseizinger: Thanks!

If it is specifically designed for this purpose, it would be nice if we could colocate the source with the test.

Do you mind including it in this repository? I was thinking of using a cargo buildscript to build (and tag) the container at build time so it is available for the test!

tumdum: @thomaseizinger Building container in build script can lead to a situation where after building, user removes all local containers and later reruns tests only to find out that they no longer pass. Is this an desired behaviour?

thomaseizinger: That is a very rare situation no?

By default, build scripts are run on every invocation anyway unless we make use of some of the ["change detection"](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed) mechanisms.
Given that container builds through docker are idempotent, this should be relatively fast.

We would have to opt out from the buildscript in non-test cases. No sure if that is possible.

thomaseizinger: From doing some research, it seems like test buildscripts are not possible. However, what we can do is move all the tests to a separate crate and include the buildscript in there :)

The crate itself doesn't need to be released obviously!

tumdum: > move all the tests to a separate crate and include the buildscript in there :)

You want to have this change done in this pull request?

thomaseizinger: > > move all the tests to a separate crate and include the buildscript in there :)
> 
> You want to have this change done in this pull request?

I saw you licensed the image source as A-GPL. To my knowledge, that is not compatible with the MIT or Apache2 license in this repository.
If you are ok with me taking the source of the docker image and putting it into this repository despite the license then we can merge the PR as is! I guess we would have to drop A-GPL then? Not sure about this to be honest.

If you are keen to introduce a testing crate as described above then even better but it is not a requirement from my side :)

tumdum: > If you are ok with me taking the source of the docker image and putting it into this repository despite the license then we can merge the PR as is! I guess we would have to drop A-GPL then? Not sure about this to be honest.

I'm completely fine with that - you can use that code as if it was licensed under either MIT or Apache2 license. 

Do you need me to change license in that repo or is my written permission here enough?

thomaseizinger: That is good enough, thank you :)

I am happy to merge this as is then!

thomaseizinger: Thanks for your contribution and patience :)

PR #177, opened 2020-06-23
Update deps

thomaseizinger:


otavio:
Awesome. I check it again and there are some more deps to update. 

Maybe, you could do the remaining updates? I'd be interested in a new release as it is using some very old dependencies and end duplicating crates.

PR #184, opened 2020-06-24
Allow containers to be started once stopped

### LINKED ISSUE #164: Allow docker client to start container
Using `Docker::stop` to stop a container is nice for testing failure conditions. It would be nice if there was a `Docker::start` so that I can restart the container and test that the issue goes away.

LLFourn:
To be able to test connection dropping. Also wait for commands to stop.

Resolves #164.

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -169,12 +169,25 @@ impl Docker for Cli {
     }
 
     fn stop(&self, id: &str) {
-        Command::new("docker")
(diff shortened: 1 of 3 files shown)


thomaseizinger:
bors cancel

thomaseizinger:
Just realized that this PR was targeting master instead of dev. Please rebase against `dev`! :)

LLFourn:
rebased against dev. Good to see that it closes an issue as well :D

PR #185, opened 2020-06-24
Misc improvements to postgres image

LLFourn:
- Allow passing in version (note 12 is the stable verison now but kept it at 11 for backwards compat).
- Remove password in favour of just trusting any connection
- Remove strange empty struct so you can actually pass in arguments

### DIFF
--- a/src/images/postgres.rs
+++ b/src/images/postgres.rs
@@ -7,6 +7,7 @@ pub struct Postgres {
     arguments: PostgresArgs,
     env_vars: HashMap<String, String>,
     ports: Option<Vec<Port>>,
+    version: u8,
 }
 
 #[derive(Default, Debug, Clone)]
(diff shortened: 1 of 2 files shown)


src/images/postgres.rs
```diff
@@ -39,12 +31,12 @@ impl Postgres {
 }
 
 impl Image for Postgres {
-    type Args = PostgresArgs;
```
thomaseizinger, line 42: The idea of `PostgresArgs` is to provide type-safety in terms of what you can pass to the image. So whatever arguments you want to pass, they should be modeled using fields in this struct and serialized properly using the `IntoIterator` implementation. See the other images for inspiration.

The reasoning is/was to provide a discoverable interface to users versus them having to read up on the documentation of the image and type the arguments themselves, possibly getting them wrong. With the struct, we can at least provide some tests and make sure what we provide works :)

If you need more flexibility over what you want to pass, I would suggest using `GenericImage` instead, which allows you to do whatever you want :D

thomaseizinger:
@LLFourn Please rebase against latest dev, not master :)

src/images/postgres.rs
LLFourn: hmmm this is strange to me since I would have thought that the right thing to do here would be to just have more parameters on the `Postgres` itself. When starting the image these could be translated into the concrete cli args etc. Then the args could just be free unsafe args the consumer wants to add (like the `env_args` params). 

But ok I see that what I really want is `GenericImage` probably. Thanks for the tip. I will remove all of those changes.

LLFourn:
I did the rebase against dev and made the note. I think this is ok to merge now.

src/images/postgres.rs
thomaseizinger: > Then the args could just be free unsafe args the consumer wants to add (like the `env_args` params).

Yeah I am not particularly happy about that design as it has the same issues in regards to type safety. Env variables are a lot less common though so I didn't bother too much for now :)

> hmmm this is strange to me since I would have thought that the right thing to do here would be to just have more parameters on the Postgres itself.

We could do that as long as we force each image to return an iterator of strings for the arguments. Not sure if it makes a huge difference though. We could be able to get rid of one associated type.

PR #200, opened 2020-09-02
Fix `Cli::rm` not blocking until rm is successful

### LINKED ISSUE #193: Containers don't stop when run from VSCode / Rust Analyzer
I'm not sure if this is a problem with testcontainers or Rust Analyzer, but I've noticed that if I run the tests from within VSCode then my containers don't ever shut down. If I run the exact same test from the command line, with the exact command line that VSCode is using then everything works exactly as expected.

This is the case regardless of whether the test passes or fails.

Pressing this "Run Test" button:
image

Comes up with:
```
> Executing task in folder service: cargo test --test get_by_id -- get_unknown_by_id --exact --nocapture <
```

And running that exact command in the terminal then correctly shuts things down.

I only noticed this when I found I had 30 containers still running from the last 2 days.

Cheers

### LINKED ISSUE #199: Containers don't stop when run from bazel test
I'm currently building my code with bazel instead of cargo. Using testcontainers, the containers appear to be left alive after the test finishes, I.E. they don't seem to be being dropped and cleaned up. Note that manually `drop`ing the `container` and `docker` variables before exiting the function does not appear to fix it.

> version: testcontainers v0.10.0
> OS: Ubuntu 20.04 inside WSL2 host
> Docker engine: 19.03.12

Code:
```rust
    use testcontainers as tc;
    use testcontainers::Docker;

    #[must_use]
    fn setup_container(client: &tc::clients::Cli) -> tc::Container<'_, tc::clients::Cli, tc::images::redis::Redis> {
        client.run(tc::images::redis::Redis::default())
    }

    #[test]
    fn test_new() -> RedisResult<()> {
        let docker = tc::clients::Cli::default();
        let container = setup_container(&docker);
        let port = container.get_host_port(6379).unwrap();
        let mut cache = TensorCache::new(Url::parse(&format!{"redis://localhost:{}/", port}).unwrap())?;

        let test_str = "are you there?";
        let response: String = redis::cmd("PING").arg(test_str).query(&mut cache.conn)?;
        assert_eq!(response, test_str);
        Ok(())
    }
````

`docker ps`:
image
P.S. There are actually two test cases, one of which I omitted in the above code, which is why you observe two docker containers at a time. I've been able to reproduce it with only the test case above, so this should not be the cause of the bug.

Probably related:
#193, #106

### PR DESCRIPTION
**Pull request by TheButlah**
Fixes #199

Potentially fixes - #193, #106

Changed:
- `Cli::rm()` method now blocks until command finishes executing.

Added:
- Test case to ensure that `Cli::rm()` command actually fails instead of silently succeeding.

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -162,14 +162,15 @@ impl Docker for Cli {
     }
 
     fn rm(&self, id: &str) {
-        Command::new("docker")
+        let output = Command::new("docker")
             .arg("rm")
             .arg("-f")
             .arg("-v") // Also remove volumes
             .arg(id)
             .stdout(Stdio::piped())
-            .spawn()
+            .output()
             .expect("Failed to execute docker command");
+        assert!(output.status.success(), "Failed to remove docker container");
     }
 
     fn stop(&self, id: &str) {
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -167,10 +167,8 @@ impl Docker for Cli {
             .arg("-f")
             .arg("-v") // Also remove volumes
             .arg(id)
-            .stdout(Stdio::piped())
             .output()
             .expect("Failed to execute docker command");
-        log::error!("{:?}", output);
         assert!(output.status.success(), "Failed to remove docker container");
     }
 
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -170,6 +170,7 @@ impl Docker for Cli {
             .stdout(Stdio::piped())
             .output()
             .expect("Failed to execute docker command");
+        log::error!("{:?}", output);
         assert!(output.status.success(), "Failed to remove docker container");
     }
 
@@ -403,5 +404,13 @@ mod tests {
         assert!(!format!("{:?}", command).contains(r#"-P"#));
         assert!(format!("{:?}", command).contains(r#""-p" "123:456""#));
         assert!(format!("{:?}", command).contains(r#""-p" "555:888""#));
+    }
+
+    #[test]
+    #[should_panic(expected = "Failed to remove docker container")]
+    fn cli_rm_command_should_panic_on_invalid_container() {
+        let docker = Cli::default();
+        docker.rm("!INVALID_NAME_DUE_TO_SYMBOLS!");
+        unreachable!()
     }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: rust
 matrix:
   include:
-    - rust: 1.36.0 # test for the MSRV when included as a library
+    - rust: 1.41.0 # test for the MSRV when included as a library
       cache: cargo
       script: cargo build
     - rust: 1.43.0 # our tests need this to compile certain dependencies

### CODE REVIEW

**Comment by TheButlah** [comment]
Looks like bors is frozen, not sure what's up with that

**Comment by TheButlah** [comment]
I don't know whats going on with CI - I think its an issue with the CI system, not the PR. LMK If this PR needs any revision!

File: src/clients/cli.rs [hunk 1]
```diff
@@ -162,14 +162,16 @@ impl Docker for Cli {
...
             .arg("rm")
             .arg("-f")
             .arg("-v") // Also remove volumes
             .arg(id)
             .stdout(Stdio::piped())
-            .spawn()
+            .output()
             .expect("Failed to execute docker command");
+        log::error!("{:?}", output);
+        assert!(output.status.success(), "Failed to remove docker container");
```
thomaseizinger, line 174: I don't think the debug log is going to very useful as it contains two `Vec<u8>` :sweat_smile: 

Reading up on the docs of [`spawn`](https://doc.rust-lang.org/std/process/struct.Command.html#method.spawn), it seems like stdin, stdout and stderr are inherited meaning if the process fails and the test is run with `--no-capture` then one would see the stderr output and can investigate the failure. That means we can probably also remove line 170 :)

thomaseizinger (approved):
Thanks for looking into this @TheButlah!

Patch looks good, made one suggestion regarding the debug log.

I've also deleted the Travis cache and kicked it again, let's see if it works this time :)

src/clients/cli.rs
TheButlah: Ok, I've made the changes. LMK if it looks good.

TheButlah:
bors merge

TheButlah:
Welp I tried. @thomaseizinger can you have bors merge it?

thomaseizinger:
Sure thing, off we go!

bors r+

TheButlah:
Looks like the build on rust 1.36 didn't work (except it did 2 days ago...?). The problem originates in `/home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/subtle-2.3.0/src/lib.rs:510`, which is not my code. @thomaseizinger looks like a problem with CI, not on my end

thomaseizinger:
> Looks like the build on rust 1.36 didn't work (except it did 2 days ago...?). The problem originates in `/home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/subtle-2.3.0/src/lib.rs:510`, which is not my code. @thomaseizinger looks like a problem with CI

Uff, seems like one of our dependencies broke their MSRV again without issuing at least a minor version bump.
The orphan rules were relaxed with Rust 1.41: https://blog.rust-lang.org/2020/01/30/Rust-1.41.0.html

I am afraid that means we need to raise the Rust version we are supporting. Do you mind editing the CI script to use Rust 1.41?

The reason it worked two days ago is because we don't have a lockfile in this repo (given that we are a library). The release must have happened since then :)

TheButlah:
Wow - I had no idea that the following:
```rust
impl<T> From<BetterVec<T>> for Vec<T> {
    //...
}
```
used to be disallowed as recently as January 2020! What a time to be alive.

I've bumped the minimum supported rust version to 1.41. Let me know if everything looks good.

thomaseizinger:
bors merge

thomaseizinger:
> I've bumped the minimum supported rust version to 1.41. Let me know if everything looks good.

Everything looks good, thanks for sticking with this! :)

PR #202, opened 2020-09-04
Issue 201

invis87:
add possibility to set network where container will be started

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -74,6 +74,10 @@ impl Cli {
     fn build_run_command<'a, I: Image>(image: &I, command: &'a mut Command) -> &'a mut Command {
         command.arg("run");
 
+        if let Some(network) = image.network() {
+            command.arg(format!("--network={}", network));
+        }
+
         for (key, value) in image.env_vars() {
             command.arg("-e").arg(format!("{}={}", key, value));
         }
@@ -320,6 +324,7 @@ mod tests {
         type EnvVars = HashMap<String, String>;
         type Volumes = HashMap<String, String>;
         type EntryPoint = std::convert::Infallible;
+        type Network = std::convert::Infallible;
 
         fn descriptor(&self) -> String {
             String::from("hello-world")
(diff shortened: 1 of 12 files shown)


thomaseizinger (changes requested):
Hi @invis87,

thanks for the contribution!
I am happy to merge this although I might give the library a cleanup in the near future as the current design causes a lot of duplication between all the images.

Before we can merge this, please rebase onto the `dev` branch. You seem to have branched off master and hence have unnecessary commits in your PR history.
In addition, a test would be nice that makes sure this feature is actually working!

invis87:
Hi!
Okey, I will rebase :)
Also, looks like for my integration tests I need to create docker network to be able to access one service from another (it is possible to do this through `host.docker.internal`, but not on all platforms). Is it ok if I add network creation to pub trait Docker?

thomaseizinger:
> Hi!
> Okey, I will rebase :)
> Also, looks like for my integration tests I need to create docker network to be able to access one service from another (it is possible to do this through `host.docker.internal`, but not on all platforms). Is it ok if I add network creation to pub trait Docker?

I'd assume it is our only option if we want to expose the functionality of docker as is. Not sure if we should but maybe we can do something clever if we zoom out a bit:

I would assume you have a usecase where you want to start several containers and they should all be on the same network? What if we make the network a configuration parameter of the client - i.e. `Cli` - and all containers started through that client instance are automatically connected to that network?

invis87:
About my case - in integration test I start several containers and most of them do not care about the others. But one container cares. And there is no way to give him connection path in `testcontainers-rs` except through host machine (by `host.docker.internal`) which does not work on Linux for now.

So docker network is the only way I see.

I think your idea - about `Cli` parameter is good and simple. But what if some one want to create two containers in one network and two in another one? Creating to `docker` instances seems strange. From my point of view method to attach image to network is better, like `postgres.with_network(net: &Network)` but I don't know how to implement `Drop` in that case. Looks like `docker.create_network` and save struct somewhere inside.

invis87:
Also, to have access from one container to another in the same net we need to explicitly set container name. Which is also not possible now :(

invis87:
test added

thomaseizinger:
Hi @invis87 !

PR #203, opened 2020-09-07
Create docker network

invis87:
Not finished.
Just to show the idea.
@thomaseizinger what do you think? I found tricky moment - network should be removed only after all containers. But looks that we can't control the order.

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -74,6 +74,10 @@ impl Cli {
     fn build_run_command<'a, I: Image>(image: &I, command: &'a mut Command) -> &'a mut Command {
         command.arg("run");
 
+        if let Some(network) = image.network() {
+            command.arg(format!("--network={}", network));
+        }
+
         for (key, value) in image.env_vars() {
             command.arg("-e").arg(format!("{}={}", key, value));
         }
@@ -320,6 +324,7 @@ mod tests {
         type EnvVars = HashMap<String, String>;
         type Volumes = HashMap<String, String>;
         type EntryPoint = std::convert::Infallible;
+        type Network = std::convert::Infallible;
 
         fn descriptor(&self) -> String {
             String::from("hello-world")
@@ -402,5 +407,18 @@ mod tests {
         assert!(!format!("{:?}", command).contains(r#"-P"#));
         assert!(format!("{:?}", command).contains(r#""-p" "123:456""#));
         assert!(format!("{:?}", command).contains(r#""-p" "555:888""#));
+    }
+
+    #[test]
+    fn cli_run_command_should_include_network() {
+        let image = GenericImage::new("hello");
+        let image = image.with_network("awesome-net");
+
+        let mut docker = Command::new("docker");
+        let command = Cli::build_run_command(&image, &mut docker);
+
+        println!("Executing command: {:?}", command);
+
+        assert!(format!("{:?}", command).contains(r#"--network=awesome-net"#));
     }
 }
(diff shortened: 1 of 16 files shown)


thomaseizinger (commented):
I like the general idea but it still forces each image to specify the network separately :thinking: 

What if we make the network an argument of the docker client (i.e. `Cli`)? We could make it optional and if it is present, we create the network on startup.

After that, all images started through this client will be created with that network.

Something like:

```rust
let client = clients::Cli::default().with_network("awesome-net");

client.run(some_image); // some_image is automatically connected to "awesome-net"
```

That would allow us to remove the "network" field from the image trait. What do you think?

invis87:
As I wrote here: https://github.com/testcontainers/testcontainers-rs/pull/202#issuecomment-688110023 your solution is nice and simple, "But what if some one want to create two containers in one network and two in another one?". Your solution is not flexible in my opinion.

By the way I already use my branch in tests and it works nice). To be sure that network removed last I use rust scopes:
```rust
fun test(...) {
  let _network = docker.create_network(&ch_net_cfg);
  {
    //create containers here
  }
//here containers aldready dropped
}
```

thomaseizinger:
> As I wrote here: [#202 (comment)](https://github.com/testcontainers/testcontainers-rs/pull/202#issuecomment-688110023) your solution is nice and simple, "But what if some one want to create two containers in one network and two in another one?". Your solution is not flexible in my opinion.

If we want to start things on two different networks, we could always create two client instances.

Alternatively, we could extend the `Docker#run` function to accept more parameters than just the image. That would keep the actual `Docker` instance stateless. I am thinking of something like:

```rust
let node = docker.run(Postgres::default(), RunArgs::default().with_name("postgres"));
```

Thoughts?

> By the way I already use my branch in tests and it works nice). To be sure that network removed last I use rust scopes:
> 
> ```rust
> fun test(...) {
>   let _network = docker.create_network(&ch_net_cfg);
>   {
>     //create containers here
>   }
> //here containers already dropped
> }
> ```

That is a neat idea! I am just a bit worried about the bloat on the `Image` trait that keeps happening. Having to implement a getter and setter for the network and other things on each image doesn't seem like a sustainable design :/

invis87:
> Alternatively, we could extend the `Docker#run` function to accept more parameters than just the image. That would keep the actual `Docker` instance stateless. I am thinking of something like:
> 
> ```rust
> let node = docker.run(Postgres::default(), RunArgs::default().with_name("postgres"));
> ```
> 
> Thoughts?

Great idea, I like it!
But currently I don't have time to refactor. Maybe a little bit later.

thomaseizinger:
> > Alternatively, we could extend the `Docker#run` function to accept more parameters than just the image. That would keep the actual `Docker` instance stateless. I am thinking of something like:
> > ```rust
> > let node = docker.run(Postgres::default(), RunArgs::default().with_name("postgres"));
> > ```
> > 
> > 
> > Thoughts?
> 
> Great idea, I like it!
> But currently I don't have time to refactor. Maybe a little bit later.

All good!
Let me know if you don't get around to it at all :)

invis87:
> > > Alternatively, we could extend the `Docker#run` function to accept more parameters than just the image. That would keep the actual `Docker` instance stateless. I am thinking of something like:
> > > ```rust
> > > let node = docker.run(Postgres::default(), RunArgs::default().with_name("postgres"));
> > > ```
> > > 
> > > 
> > > Thoughts?
> > 
> > 
> > Great idea, I like it!
> > But currently I don't have time to refactor. Maybe a little bit later.
> 
> All good!
> Let me know if you don't get around to it at all :)

I create separate branch (cause my production service depends on current one): https://github.com/testcontainers/testcontainers-rs/pull/205

thomaseizinger:
Hi @invis87 !

PR #205, opened 2020-09-18
Create docker network args

invis87:
Add ability to set `name`, `network` and some custom parameters for `docker run` command

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -1,4 +1,4 @@
-use crate::core::{self, Container, Docker, Image, Logs, Network, NetworkConfig};
+use crate::core::{self, Container, Docker, Image, Logs, Network, NetworkConfig, RunArgs};
 use std::collections::HashMap;
 use std::sync::RwLock;
 use std::time::Instant;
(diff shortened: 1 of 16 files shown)


invis87:
Conflict because you change `rm` function to `assert` output, but I need to `wait` on it:
```
            .wait()
            .expect("Failer to remove docker container");
```
to be able to remove Network after. Otherwise, network can't be deleted with running containers inside it.

thomaseizinger:
> Conflict because you change `rm` function to `assert` output, but I need to `wait` on it:
> 
> ```
>             .wait()
>             .expect("Failed to remove docker container");
> ```
> 
> to be able to remove Network after. Otherwise, network can't be deleted with running containers inside it.

Yes, that was changed in PR #200 which fixed a few long outstanding bugs :)
That shouldn't affect you though I think. Calling `.output` implies `.wait` according to the docs: https://doc.rust-lang.org/std/process/struct.Command.html#method.output

> Executes the command as a child process, **waiting for it to finish** and collecting all of its output.

thomaseizinger (commented):
Thank you @invis87 !

I took this PR as a base and built https://github.com/testcontainers/testcontainers-rs/pull/206 on top of it. Would that design work for you?

The idea was to keep the `Docker` and `Image` traits slimmer by creating the network on startup of the container if it doesn't exist yet. We remove the network once the client is dropped.

invis87:
> Thank you @invis87 !
> 
> I took this PR as a base and built #206 on top of it. Would that design work for you?
> The idea was to keep the `Docker` and `Image` traits slimmer by creating the network on startup of the container if it doesn't exist yet. We remove the network once the client is dropped.

A little refactor will be needed for my code, but I think it will work.

Hope it will be possible to add Network configuration in the future (if someone need it), cause docker networks is very elastic

thomaseizinger:
> > Thank you @invis87 !
> > I took this PR as a base and built #206 on top of it. Would that design work for you?
> > The idea was to keep the `Docker` and `Image` traits slimmer by creating the network on startup of the container if it doesn't exist yet. We remove the network once the client is dropped.
> 
> A little refactor will be needed for my code, but I think it will work.

That is good to hear! Thanks!

> Hope it will be possible to add Network configuration in the future (if someone need it), cause docker networks is very elastic

I am happy to explore which other functionality can/should be added! Having said that, I am not planning to expose every feature of docker through the API of testcontainers. At least not if it comes at the expense of usability or adds verbosity to standard usecases :)

thomaseizinger:
Hi @invis87 !

PR #206, opened 2020-09-21
Allow containers to be named and assigned to a network on startup

thomaseizinger:
Docker::run_args allows users to specify a docker network and a name
for the container. The network will be created on-demand and clean-up
in the end when the client is dropped.

Co-authored-by: Aleksandrov Vladimir <invis87@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
--   `Docker#run_with_args` method. This allows naming a container and assigning it to a specific docker network. The docker
+-   `Docker::run_with_args` method. This allows naming a container and assigning it to a specific docker network. The docker
 network will be created if it doesn't exist yet. Once the client is dropped, the network will be removed again if it
 has previously been created. A network that already existed will not be removed.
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,13 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+-   `Docker#run_with_args` method. This allows naming a container and assigning it to a specific docker network. The docker
```
D4nte, line 12:
```suggestion
-   `Docker::run_with_args` method. This allows naming a container and assigning it to a specific docker network. The docker
```
Why using a hash symbol?

src/clients/cli.rs
```diff
@@ -195,6 +223,53 @@ impl Docker for Cli {
     }
 }
 
+impl Drop for Cli {
```
D4nte, line 226: Are you testing that dropping the cli stops the container?

D4nte (approved):

goller:
Just tried the branch; works great !

src/clients/cli.rs
thomaseizinger: Dropping the client does not stop the container, dropping the container stops the container :)

Rust lifetimes guarantee that the containers are dropped before the client.

CHANGELOG.md
thomaseizinger: Because I've wrote too much Java in my life.

PR #207, opened 2020-09-21
Added address_type default to bech 32 to bitcoincore

### LINKED ISSUE #161: Allow to specify addresstype for bitcoind image
Bitcoind allows to specify the default type for addresses generated through `getnewaddress` with the `-addresstype` option.

We should allow users to configure this option in the args and default to bech32.

ywalterh:
This is to address issue: https://github.com/testcontainers/testcontainers-rs/issues/161

Passed all existing test including coblox_bitcoincore_getnewaddress which caught my typo in types. Let me know if I need to add a test for difference address type.

Fixes #161.

### DIFF
--- a/src/images/coblox_bitcoincore.rs
+++ b/src/images/coblox_bitcoincore.rs
@@ -56,7 +56,6 @@ impl RpcAuth {
     fn generate_salt() -> String {
         let mut buffer = [0u8; 16];
         thread_rng().fill(&mut buffer[..]);
-
         encode(buffer)
     }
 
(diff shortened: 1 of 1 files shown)


src/images/coblox_bitcoincore.rs
```diff
@@ -93,6 +92,7 @@ pub struct BitcoinCoreImageArgs {
     pub accept_non_std_txn: Option<bool>,
     pub rest: bool,
     pub fallback_fee: Option<f64>,
+    pub address_type: String,
```
thomaseizinger, line 95: Given that there is only a limited set of address-types, we could model this as an enum!

The valid options are:

- `legacy`
- `p2sh-segwit`
- `bech32`

thomaseizinger (commented):
Hi @ywalterh!

Thanks for the contribution!
One comment but looks good otherwise :)

src/images/coblox_bitcoincore.rs
ywalterh: Great feedback! I've implemented it and commit the new change.

thomaseizinger (approved):
Thank you!

bors r+

Issue #204, opened 2020-09-10
Any interest in breaking out some of this crates functionality into a general purpose docker client api?


TheButlah:
I'm fairly unfamiliar with Rust's options for docker clients, but from what I can see based on a cursory search of crates.io, I think this library's api is a lot more ergonomic than the alternatives out there. I know that this crate is not intended for production use or use outside of integration tests, but is there any code that could be extracted into another crate and used as the start of a general purpose docker client library? Any thoughts on what that might look like?

thomaseizinger:
Initially when I created this lib, I didn't think that we would be stuck with a `docker` CLI based client for this long :D

I actually intend to at some point switch over to a HTTP-based one to allow for greater flexibility and also async support.

Generally, I don't think the code in this crate is a good candidate for being extracted:

1. It is very minimal, and could easily be reproduced if you need a simple client.
2. There is no error handling. We have lots of unwraps everywhere which is really only possible because we only want to be used in tests. This is unacceptable for any kind of production code that needs to interact with docker.

I would actually like to go the other way round: Replace the CLI client with a proper HTTP client. Options I know of:

- https://github.com/softprops/shiplift
- https://github.com/fussybeaver/bollard

ywalterh:
Moving to HTTP client definitely would make features like exposing host Docker daemon is running on much easier. But it'll change the way testcontainers framework starts up. Mac / Linux uses unix socket and Windows don't. Docker binary is a good abstraction to dodge these problems. But for long term I believe it's a worthwhile change. 

I had some experience porting CLI lib to native Docker API. I can take a stab at it using either shiplift / bollard. (Or report back neither are feature complete). After the change:

- Only Mac / Linux would have good unix socket default 
- Windows and remote docker host would need to provide Host / Port either via code level API or DOCKER_HOTS env
- Features depending `docker cp` and `docker log` might require extra work in the library to handle

Let me know if I can help.

thomaseizinger:
> I had some experience porting CLI lib to native Docker API. I can take a stab at it using either shiplift / bollard. (Or report back neither are feature complete). After the change:
> - Only Mac / Linux would have good unix socket default 
> - Windows and remote docker host would need to provide Host / Port either via code level API or DOCKER_HOTS env
> - Features depending `docker cp` and `docker log` might require extra work in the library to handle

[...]

PR #216, opened 2020-09-30
Use shiplift HTTP Docker client library instead of CLI

ywalterh:
Followed the discussion of issue #204 

- [x] Added shiplift library implementing basic Docker trait
- [x] Make shiplift implementation feature equivilent
- [x] Replace Cli in library and pass all test
- [x] Figure out Drop pattern

- [x] Implement GenericImageAsync
- [ ] Implement a canned image in Async trait
- [x] Add more integration test to shiplift client

Worklog:
- Creating and running container working, using tokio::runtime to get around the fact tha shiplift is mostly async
- Created DockerAsync trait and all associated type in async
- Figured out how to handle logging in async.
- Use futures::executor::block_on to run Drop rm


ywalterh:
Hey @thomaseizinger , shiplift library is pretty good so far. (At least the git repo is, 0.6.0 is pretty outdated)

I have a directional question: shiplift is async by nature and testcontainers-rs aren't doing that yet. Do you think I should start making the Docker trait async or use tokio::runtime to hack around it to emulate what Cli module is doing right now.

I attached some WIP code, the `Docker` trait is very helpful.

thomaseizinger:
> I have a directional question: shiplift is async by nature and testcontainers-rs aren't doing that yet. Do you think I should start making the Docker trait async or use tokio::runtime to hack around it to emulate what Cli module is doing right now.

Yes, I think making the trait async is the way to go. The migration of existing code while certainly possible would be quite tedious.

Maybe we can start with adding a `DockerAsync` trait that mimics the current interface but has all its functions async? Users could then choose which trait to import and we can deprecate the old one eventually (or maybe just leave it).

For async-traits, you will need https://docs.rs/async-trait/0.1.41/async_trait/.

thomaseizinger:
Also, thanks for picking this up, I am excited to land this :)

ywalterh:
> > I have a directional question: shiplift is async by nature and testcontainers-rs aren't doing that yet. Do you think I should start making the Docker trait async or use tokio::runtime to hack around it to emulate what Cli module is doing right now.
> 
> Yes, I think making the trait async is the way to go. The migration of existing code while certainly possible would be quite tedious.
> 
> Maybe we can start with adding a `DockerAsync` trait that mimics the current interface but has all its functions async? Users could then choose which trait to import and we can deprecate the old one eventually (or maybe just leave it).
> 
> For async-traits, you will need https://docs.rs/async-trait/0.1.41/async_trait/.

Got it, I'll get started on async trait route then. It'll save me a lot of dancing around with log handling too. I think it's a good approach.

src/clients/shiplift.rs
```diff
@@ -252,20 +258,6 @@ mod tests {
...
-    #[test]
-    fn shiplift_run_command_should_include_name() {
-        let image = HelloWorld::default();
-        let run_args = RunArgs::default().with_name("hello_container");
-
-        let shiplift = Shiplift::new();
-        shiplift.run_with_args(image, run_args);
-
-        //TODO add assert maybe receive the container and then inspect the id
+        tokio_test::block_on(shiplift.run(image));
```
thomaseizinger, line 261: If you depend on `tokio` with the `macros` feature enabled, you can write `#[tokio::test]` instead of `#[test]` and make the test `fn` itself `async` so you can just use `.await` :)

src/core/container_async.rs
```diff
@@ -0,0 +1,62 @@
+use crate::core::{DockerAsync, ImageAsync};
+
+/// Represents a running docker container using async trait.
+///
+/// Containers have a [`custom destructor`][drop_impl] that removes them as soon as they go out of scope:
```
thomaseizinger, line 5: This might actually be a problem because Rust doesn't have async drop yet :(

I think we might need to hack around this by blocking the thread in the `Drop` impl. Not sure what the current best practice here is.

src/clients/shiplift.rs
ywalterh: Yup! I was looking for that syntax, and didn't dig deeper. 

Btw, do you think this will introduce some inconvenience to the users of the testcontainer crate? To use Async Docker trait you'll have to depend on tokio crate.

thomaseizinger: Yes, it might create some inconvenience which is why I was thinking of keeping both APIs :)

src/core/container_async.rs
ywalterh: Unfortunately after extensive reading and researching it's still a really hard problem and not even sure if rust wants to support it. I could do something crazy but rust async is designed to block people from doing such crazy thing. Like doing a `block_on` inside a Runtime context. I've tried a few and no luck.

Any suggestion on which route we should take? Is there a way to design the code around needing the `Drop`?

**The** article: https://boats.gitlab.io/blog/post/poll-drop/
Discussion:
- https://users.rust-lang.org/t/what-is-the-idiomatic-way-of-cleaning-up-async-resources-in-rust/48878
- https://stackoverflow.com/questions/59940976/panic-running-async-code-while-dropping-rust-future

src/core/container_async.rs
```diff
@@ -71,4 +72,53 @@ where
...
+        if keep_container {
+            self.stop().await
+        } else {
+            self.rm().await
+        }
+        println!("finished dropping");
+    }
+}
+
+// use futures::executor::block_on;
```
thomaseizinger, line 109: I think that is a good call, we don't have any other option at this stage!

ywalterh: :( unfortunately it doesn't work yet. It just stalled and never poll for the future underneath. I'll try more things to see if I can get a reliable Drop from an async test. I put more info on the comment above.

Journey continues.

thomaseizinger: I wonder if that is the case because the whole test fn is a future and is being polled by another executor and we may stall the executor thread with this function.

Worst case users have to manually call `rm`/`stop` within async contexts for now, that would a bit sad though :(

ywalterh: Ha! You are right, `#[tokio::test]` does treat the whole test as a future. `tokio_test::block_on` inside normal test on the other hand will work with a `Drop` method containing a small runtime of its own. But I don't like the feeling of restricting how the test should be written, if anything using `#[tokio::test]` is more favorable like you mentioned in another comment.  I'll keep digging. There might be other ways like spawning another thread and joining.

ywalterh: After adding `#[tokio::test(threaded_scheduler)]` to test, `block_on` in Drop started doing the right thing and container is getting cleaned up!

I think it's a small price to pay to use the async API. I haven't read up much on threaded_scheduler and I think they just changed the API to `core_thread` and `max_thread`.

thomaseizinger: If it works with `threaded_scheduler` then the initial assumption was right: `block_on` blocks the executor thread which means the future that actually does the `Container#rm` cannot make any progress because it is being executed within the same thread.

To make this easier to use, we might want to expose our own proc-macro at some point for this so we can do:

```rust
#[testcontainers::async_test]
async fn test() {

}
```

which would always use a threaded runtime.

I wonder if there could be any race-conditions where with bad luck, the future that does `Container#rm` still ends up being on the same thread as the one that we block and so things don't work _sometimes_. I am not sure if tokio is smart enough to move futures between executor threads if it noticed that one of them is blocked.

ywalterh: Yeah, it does sound like a valid concern. Even thought Drop technically happens after all previous `.await` calls, there's no guaranteed the cleanup future gets pulled last. Let me clean it up a bit and try it on a slower VM and try limiting it to one core.

thomaseizinger: Another way of trying this might be to just execute `cargo test` (for a single test) a thousand times or something :D
To increase the chances of it happening (if it ever happens), you also might want to consider setting the number of executor threads to 1 or possibly 2. I don't know if tokio also uses the current thread for executing futures or spawns new ones.

ywalterh: Haha, that's exactly what I did and I ran it on my slower Linux VM. I didn't encounter any stalled and every run passed and all containers are removed. Tokio test API has newer control over this by allowing `core_thread` and `max_thread` which works as well. We might just be safer to implement `#[testcontainers::async_test]` trait, and fingers crossed tokio doesn't remove `threaded_scheduler`. The version I'm using didn't mention the macro at all. 

For now I'll start adding a more complicated reference implementation like redis to see if we are seeing other issue.

thomaseizinger: > I didn't encounter any stalled and every run passed and all containers are removed.

That is good news! Thanks for testing!

src/clients/shiplift.rs
```diff
@@ -6,7 +6,10 @@ use shiplift::{ContainerOptions, LogsOptions, NetworkCreateOptions, RmContainerO
 use std::fmt;
 
 pub struct Shiplift {
-    client: Docker,
+    // XXX should this be private? Need advice
```
thomaseizinger, line 9: Ideally yes, if we manage to not expose any `shiplift` types in our public API then changes to their API will not affect our users :)

thomaseizinger: Which means we could potentially make `patch` / `minor` version releases even if shiplift released a breaking change.

thomaseizinger: We might also consider renaming this client to `Http` or something and completely hide that we are using shiplift under-neath :thinking:

tests/shiplift_client.rs
```diff
@@ -76,3 +79,50 @@ async fn should_wait_for_at_least_one_second_before_fetching_logs_shiplift() {
...
+    let container_details = docker
+        .client
+        .containers()
+        .get(container.id())
+        .inspect()
+        .await
+        .unwrap();
+
+    assert_that!(container_details.host_config.publish_all_ports).is_equal_to(true);
+}
```
thomaseizinger, line 129: You could move these tests into the `src/clients/shiplift.rs` file, then you have access to private fields :)

ywalterh: That's a good point. I was gonna ask if there's a pattern of leaving tests actually running container out of unit test. If you have no concern I'll move them back to `shiplift.rs`

src/clients/shiplift.rs
ywalterh: Totally agree, making it private. Yeah Shiplift is a temporary name and it does sound awfully specific. I like `Http` or maybe just `Api` or `Engine` (referring https://docs.docker.com/engine/api/). I can clean it up once we agree on the name.

tests/shiplift_client.rs
thomaseizinger: The only difference between tests in `tests/` and tests within the module is that tests in `tests/` can only access the public API and are hence a good way of testing the lib in a way users will have to deal with it. For example:

- does the module structure make sense (in the `use` statements)?
- are all types accessible that need to be accessible?

In this case, we are testing an implementation detail of the `Shiplift` client, so it is totally fine to do that within the actual module.

I am not fussed about the "unit-test go into the module and integration-test go into `tests/" policy. In my experience, it is usually quite hard to even find an agreement on what a unit test actually is :)

Sidenote: Excellent talk on this topic: https://www.youtube.com/watch?v=EZ05e7EMOLM

Generally, it would be ideal if as many tests as possible - be it unit or integration tests - are as fast as possible. Starting containers is slow, so if you can test things in an effective way without starting containers that is great. If starting containers is necessary to ensure things actually work, that is also ok :)

ywalterh: Thanks for the thorough explanation! I'm with you on unit test / integration distinction sometimes unnecessary. Rust makes it very easy to organize tests so we can just focus on getting good coverage. These tests are moved back into `shiplift.rs`.

Since the API is a wrapper around shiplift, it's hard to test its functionalities without starting a real container. These tests are running blazing fast (at least on my Mac), so shouldn't be a concern for now.  :)

src/clients/shiplift.rs
ywalterh: Changed to `Http`

ywalterh:
@thomaseizinger ready for some constructive feedback!

src/core/container_async.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// Returns the id of this container.
+    pub fn id(&self) -> &str {
+        &self.id
+    }
+
+    /// Gives access to the log streams of this container.
+    pub async fn logs<'a>(&'a self) -> LogsAsync<'a> {
+        let id_clone: String = String::from(&self.id);
+        self.docker_client
+            .logs(Box::leak(id_clone.into_boxed_str()))
```
thomaseizinger, line 67: This `Box::leak` is a bit concerning. I think we have a memory leak here if I am reading the [docs](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.leak) correctly.

Feel free to change the signature of `.logs` if you don't want to mess around with manual dropping here.

src/core/image_async.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// rarely instantly available as soon as they are started. Most of them take some time to boot
+    /// up.
+    ///
+    /// Implementations MUST return a async function signalling the current thread that the
+    /// passed-in container is ready to be interacted with. The container instance provides
+    /// access to logs of the container.
+    ///
+    /// Most implementations will very likely want to make use of this to wait for a particular
+    /// message to be emitted.
+    async fn wait_until_ready<D: DockerAsync + Sync>(
```
thomaseizinger, line 82: This is the only function within the `ImageAsync` trait that is actually async. I think we should refactor this whole "waiting" concept in a way that works for blocking and async clients. That would allow us to reuse the same image definition for both blocking and async clients. Currently, we need to redefine every image with the `ImageAsync` trait to use it with an async client :/

One idea on how to achieve this would be to remove some flexibility in what images can do here. In particular, there seem to be primarily three things that images do to determine that they are ready:

- Wait for a message on stdout
- Wait for a message on stderr
- Wait for an arbitrary amount of time

I think we can model that be having a function that returns a `Vec<WaitFor>` where `WaitFor` is:

```rust
enum WaitFor {
	Message { message: String, stream: Stream },
	Time { ms: u64 }
}
```

Cargo.toml
```diff
@@ -18,6 +18,10 @@ rand = "0.7"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 sha2 = "0.9"
+shiplift =  {git = "https://github.com/softprops/shiplift" }
```
thomaseizinger, line 21: Let's hope they are cutting a release soon! We can't release with `git` dependencies so if we merge this before hand, we will need to do some manual work to exclude this code on every release. That shouldn't be too hard to so it is not actually a blocker for merging this PR! 

GitFlow for the win! :grin:

thomaseizinger: To make this easier, could you squash all of the commits into a single one so I can just revert that when we make releases?

thomaseizinger (changes requested):
First, thank you so much for working on this @ywalterh! :)
Using an HTTP-based docker client is something I've envisioned from the very beginning! It is nice to see it happening now!

I've left several comments with many of them being fairly small.

There is one bigger topic regarding refactoring how `wait_until_ready` works. I believe it would be worthwhile try and get rid of the `ImageAsync` trait so that we don't have to redefine all images and they are automatically compatible with several clients.

Let me know if you are still keen to work on this! If you do, it would be good to have this refactoring as a separate commit at the beginning of this PR (or maybe even as a completely separate PR and we rebase this on top of that) so that I can easily revert the async stuff in the releases branches until `shiplift` cuts a release!

ywalterh:
> First, thank you so much for working on this @ywalterh! :)
> Using an HTTP-based docker client is something I've envisioned from the very beginning! It is nice to see it happening now!
> 
> I've left several comments with many of them being fairly small.
> 
> There is one bigger topic regarding refactoring how `wait_until_ready` works. I believe it would be worthwhile try and get rid of the `ImageAsync` trait so that we don't have to redefine all images and they are automatically compatible with several clients.
> 
> Let me know if you are still keen to work on this! If you do, it would be good to have this refactoring as a separate commit at the beginning of this PR (or maybe even as a completely separate PR and we rebase this on top of that) so that I can easily revert the async stuff in the releases branches until `shiplift` cuts a release!

No problem at all. Sorry I was on Thanksgiving break and didn't reply to this sooner. Let me address these issues first and look into the feasibility of removing `ImageAsync`.

thomaseizinger:
Closing this in favor of https://github.com/testcontainers/testcontainers-rs/pull/268.

Thank you for all the hard work @ywalterh !

PR #218, opened 2020-10-13
Create zookeeper image with him test.

devcdcc:
Hi, this is my first contribution and i would like to contribute with kafka image for rust, but kafka needs for zookeeper, this is the zookeeper image and tomorrow should i will to publish kafka.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,3 +33,5 @@ spectral = "0.6"
 reqwest = { version = "0.10", features = ["blocking"] }
 json = "0.12"
 tokio = { version = "0.2", features = ["macros"] }
+env_logger = "0.7"
+zookeeper = "0.5"
--- a/src/images/zookeeper.rs
+++ b/src/images/zookeeper.rs
@@ -1,0 +1,86 @@
+use crate::core::Port;
+use crate::{Container, Docker, Image, WaitForMessage};
+use std::collections::HashMap;
+
+const CONTAINER_IDENTIFIER: &str = "zookeeper";
+const DEFAULT_TAG: &str = "3.6.2";
+
+#[derive(Debug, Default, Clone)]
+pub struct ZookeeperArgs;
+impl IntoIterator for ZookeeperArgs {
+    type Item = String;
+    type IntoIter = ::std::vec::IntoIter<String>;
+
+    fn into_iter(self) -> <Self as IntoIterator>::IntoIter {
+        vec![].into_iter()
+    }
+}
+
+#[derive(Debug)]
+pub struct Zookeeper {
+    tag: String,
+    arguments: ZookeeperArgs,
+    ports: Option<Vec<Port>>,
+}
+
+impl Default for Zookeeper {
+    fn default() -> Self {
+        Zookeeper {
+            tag: DEFAULT_TAG.to_string(),
+            arguments: ZookeeperArgs {},
+            ports: None,
+        }
+    }
+}
+impl Image for Zookeeper {
+    type Args = ZookeeperArgs;
+    type EnvVars = HashMap<String, String>;
+    type Volumes = HashMap<String, String>;
+    type EntryPoint = std::convert::Infallible;
+    fn descriptor(&self) -> String {
+        format!("{}:{}", CONTAINER_IDENTIFIER, &self.tag)
+    }
+
+    fn wait_until_ready<D: Docker>(&self, container: &Container<'_, D, Self>) {
+        container
+            .logs()
+            .stdout
+            .wait_for_message("Started AdminServer")
+            .unwrap();
+    }
+
+    fn args(&self) -> <Self as Image>::Args {
+        self.arguments.clone()
+    }
+
+    fn volumes(&self) -> Self::Volumes {
+        HashMap::new()
+    }
+
+    fn env_vars(&self) -> Self::EnvVars {
+        HashMap::new()
+    }
+
+    fn ports(&self) -> Option<Vec<Port>> {
+        self.ports.clone()
+    }
+
+    fn with_args(self, arguments: <Self as Image>::Args) -> Self {
+        Zookeeper { arguments, ..self }
+    }
+}
+impl Zookeeper {
+    pub fn with_tag(self, tag_str: &str) -> Self {
+        Zookeeper {
+            tag: tag_str.to_string(),
+            ..self
+        }
+    }
+
+    pub fn with_mapped_port<P: Into<Port>>(mut self, port: P) -> Self {
+        let mut ports = self.ports.unwrap_or_default();
+        ports.push(port.into());
+        self.ports = Some(ports);
+        self
+    }
+}
--- a/tests/images.rs
+++ b/tests/images.rs
@@ -1,3 +1,8 @@
+#![deny(unused_mut)]
+extern crate env_logger;
+extern crate log;
+extern crate zookeeper;
+
 use bitcoincore_rpc::RpcApi;
 use mongodb::{bson, Client};
 use redis::Commands;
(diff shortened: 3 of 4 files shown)


Cargo.toml
```diff
@@ -33,3 +33,5 @@ spectral = "0.6"
 reqwest = { version = "0.10", features = ["blocking"] }
 json = "0.12"
 tokio = { version = "0.2", features = ["macros"] }
+env_logger = "0.7"
+zookeeper = "0.5"
```
thomaseizinger, line 37: Missing a newline at the end of the file :)

src/images/zookeeper.rs
```diff
@@ -0,0 +1,88 @@
+use crate::core::Port;
+use crate::{Container, Docker, Image, WaitForMessage};
+use std::collections::HashMap;
+
+const CONTAINER_IDENTIFIER: &str = "bitnami/zookeeper";
```
thomaseizinger, line 5: Any reason we can't use the official zookeeper image? https://hub.docker.com/_/zookeeper

src/images/zookeeper.rs
```diff
@@ -0,0 +1,88 @@
...
+    arguments: ZookeeperArgs,
+    ports: Option<Vec<Port>>,
+}
+
+impl Default for Zookeeper {
+    fn default() -> Self {
+        Zookeeper {
+            tag: DEFAULT_TAG.to_string(),
+            arguments: ZookeeperArgs {},
+            ports: None,
```
thomaseizinger, line 31: Does zookeeper have a default port? If yes we can include it here to make usage easier!

tests/images.rs
```diff
@@ -322,3 +329,45 @@ pub fn free_local_port() -> Option<u16> {
...
+        }
+    }
+    let zk_urls = zk_server_urls(host_port);
+    println!("connecting to {}", zk_urls);
+
+    let zk = ZooKeeper::connect(&*zk_urls, Duration::from_secs(15), LoggingWatcher).unwrap();
+    zk.add_listener(|zk_state| println!("New ZkState is {:?}", zk_state));
+
+    let auth = zk.add_auth("digest", vec![1, 2, 3, 4]);
+    println!("authenticated -> {:?}", auth);
```
thomaseizinger, line 363: Are these printlns really necessary? I would prefer to keep the tests free of such output!

tests/images.rs
```diff
@@ -322,3 +329,45 @@ pub fn free_local_port() -> Option<u16> {
...
+        match env::var(key) {
+            Ok(val) => val,
+            Err(_) => format!("localhost:{}", port),
+        }
+    }
+    let zk_urls = zk_server_urls(host_port);
+    println!("connecting to {}", zk_urls);
+
+    let zk = ZooKeeper::connect(&*zk_urls, Duration::from_secs(15), LoggingWatcher).unwrap();
+    zk.add_listener(|zk_state| println!("New ZkState is {:?}", zk_state));
```
thomaseizinger, line 360: What do we need the listener for in the test?

tests/images.rs
```diff
@@ -322,3 +329,45 @@ pub fn free_local_port() -> Option<u16> {
...
+#[test]
+fn zookeeper_check_connection() {
+    struct LoggingWatcher;
+    impl Watcher for LoggingWatcher {
+        fn handle(&self, e: WatchedEvent) {
+            info!("{:?}", e)
+        }
+    }
+    use std::env;
+    use std::time::Duration;
```
thomaseizinger, line 342: These can go at the top of the file.

tests/images.rs
```diff
@@ -322,3 +329,45 @@ pub fn free_local_port() -> Option<u16> {
         .map(|addr| addr.port())
         .ok()
 }
+
+#[test]
+fn zookeeper_check_connection() {
```
thomaseizinger, line 334: This test is doing quite a lot. Any chance we can trim does down to the very essentials that are required to assert that things are working?

thomaseizinger (changes requested):
Looks good overall!

I a few comments inline :)

src/images/zookeeper.rs
devcdcc: Hi, @thomaseizinger  sure. Could i separate zookeeper tests in another file?

tests/images.rs
```diff
@@ -331,43 +331,33 @@ pub fn free_local_port() -> Option<u16> {
...
-    use std::time::Duration;
-    let _ = pretty_env_logger::try_init();
-    let free_local_port = free_local_port().unwrap();
+fn zookeeper_check_existence_directories() {
+    let default_zookeeper_port = 2181;
+    let free_local_port = free_local_port().unwrap_or(default_zookeeper_port);
     let docker = clients::Cli::default();
-    let image = images::zookeeper::Zookeeper::default().with_mapped_port((free_local_port, 2181));
+    let image = images::zookeeper::Zookeeper::default()
+        .with_mapped_port((free_local_port, default_zookeeper_port));
```
thomaseizinger, line 339: You don't need to specify this again here, just calling `node.get_host_port` should be sufficient!

src/images/zookeeper.rs
```diff
@@ -0,0 +1,90 @@
+use crate::core::Port;
+use crate::{Container, Docker, Image, WaitForMessage};
+use std::collections::HashMap;
+
+const CONTAINER_IDENTIFIER: &str = "zookeeper";
+const DEFAULT_TAG: &str = "3.5.8";
```
thomaseizinger, line 6: Why not the latest tag `3.6.2`?

src/images/zookeeper.rs
```diff
@@ -0,0 +1,90 @@
...
+    arguments: ZookeeperArgs,
+    ports: Option<Vec<Port>>,
+}
+
+impl Default for Zookeeper {
+    fn default() -> Self {
+        Zookeeper {
+            tag: DEFAULT_TAG.to_string(),
+            arguments: ZookeeperArgs {},
+            ports: Some(vec![Port {
```
thomaseizinger, line 32: So actually, we don't even need this! Sorry to send you on a wild goose chase here :sweat_smile: 

Looking at the source of the docker image (https://hub.docker.com/layers/zookeeper/library/zookeeper/3.6.2/images/sha256-a6ba8cd94070ecd6d30f0c2c0911dc04c7d6c7eef77e35460d8e43cdb7a0bd33?context=explore), it automatically exposes the port `2181`.

That combined with [this](https://github.com/testcontainers/testcontainers-rs/blob/3d80acb27677a5897a5a07a8d5276024a8979d93/src/clients/cli.rs#L101-L109) functionality in testcontainers makes things "just work" if people are using `Container#get_host_port`.

If we specify the port here, we force docker to use 2181 on the host as well which means we can only have one of those images running at the same time.

In conclusion, we should actually not default this here. I guess we still want to keep the `with_mapped_port` functionality but I think that would actually fit better onto the [`RunArgs`](https://github.com/testcontainers/testcontainers-rs/blob/3d80acb27677a5897a5a07a8d5276024a8979d93/src/core/docker.rs#L22) that we introduced recently.

thomaseizinger (changes requested):
Thanks!

Two more things:
1. As commented inline, I think we should actually not set the default port.
2. Please squash all your commits into one after that!

src/images/zookeeper.rs
devcdcc: i just thought  in kafka compatibility, i was wrong, we can specify  the version later :)

src/images/zookeeper.rs
devcdcc: yeah, I got it, thanks.

devcdcc:
> Thanks!
> 
> Two more things:
> 
> 1. As commented inline, I think we should actually not set the default port.
> 2. Please squash all your commits into one after that!

devcdcc:
@thomaseizinger sorry i did a miss click and close the PR XD
And don't worry about suggestion, this is some of my first rust code.

thomaseizinger (approved):
Thanks for the contribution @devcdcc !

devcdcc:
Thanks for suggestions @thomaseizinger , i am newbie in rust and i would like to continue contributing to open source projects :D

PR #229, opened 2020-10-31
create kafka image and few core utilities (gateway, ip_address).

devcdcc:
Hi, there is an image for Kafka.
Could someone review that?
Thanks.  :)

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -171,21 +171,7 @@ impl Docker for Cli {
     }
 
     fn ports(&self, id: &str) -> crate::core::Ports {
-        let child = Command::new("docker")
-            .arg("inspect")
-            .arg(id)
-            .stdout(Stdio::piped())
-            .spawn()
-            .expect("Failed to execute docker command");
-
-        let stdout = child.stdout.unwrap();
-
-        let mut infos: Vec<ContainerInfo> = serde_json::from_reader(stdout).unwrap();
-
-        let info = infos.remove(0);
-
-        log::trace!("Fetched container info: {:#?}", info);
-
+        let info = Cli::network_settings(id);
         info.network_settings.ports.into_ports()
     }
 
@@ -533,5 +519,22 @@ mod tests {
 
         // client has been dropped, should clean up networks
         assert!(!network_exists("awesome-net"))
+    }
+}
+
+impl Cli {
+    fn network_settings(id: &str) -> ContainerInfo {
+        let child = Command::new("docker")
+            .arg("inspect")
+            .arg(id)
+            .stdout(Stdio::piped())
+            .spawn()
+            .expect("Failed to execute docker command");
+
+        let stdout = child.stdout.unwrap();
+        let mut infos: Vec<ContainerInfo> = serde_json::from_reader(stdout).unwrap();
+        let info = infos.remove(0);
+        log::trace!("Fetched container info: {:#?}", info);
+        info
     }
 }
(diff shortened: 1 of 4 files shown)


tests/images.rs
```diff
@@ -380,3 +381,56 @@ fn orientdb_exists_database() {
 
     assert!(!exists);
 }
+
+#[test]
+fn kafka_publish_and_read_message() {
+    let free_local_port = free_local_port().unwrap();
+    let docker = clients::Cli::default();
+    let zk_image =
+        images::zookeeper::Zookeeper::default().with_mapped_port((free_local_port, 2181));
```
thomaseizinger, line 390: I don't know much about zookeeper and kafka but if these images are supposed to talk to each other, you might want to utilize the functionality for docker networks that was added recently: https://github.com/testcontainers/testcontainers-rs/pull/206

tests/images.rs
```diff
@@ -380,3 +381,56 @@ fn orientdb_exists_database() {
...
+    let zk_node = docker.run(zk_image);
+    let zk_host_port = zk_node.get_host_port(2181).unwrap();
+    let zk_host_address = zk_node.get_container_gateway();
+    let zk_urls = format!("{}:{}", zk_host_address, zk_host_port);
+    let kafka_image = images::kafka::Kafka::default()
+        .with_env_var("KAFKA_CFG_AUTO_CREATE_TOPICS_ENABLE", "true")
+        .with_env_var("ALLOW_ANONYMOUS_LOGIN", "yes")
+        .with_env_var("ALLOW_PLAINTEXT_LISTENER", "yes")
+        .with_env_var("KAFKA_CFG_AUTO_CREATE_TOPICS_ENABLE", "true")
+        .with_env_var("KAFKA_CFG_ZOOKEEPER_CONNECT", &*zk_urls);
```
thomaseizinger, line 400: Would it make sense to set these env vars by default within the image?

tests/images.rs
```diff
@@ -380,3 +381,56 @@ fn orientdb_exists_database() {
...
+            "bash",
+            "-c",
+            &*consumer_command,
+        ])
+        .output()
+        .unwrap();
+    assert_eq!(
+        &*String::from_utf8(consumer.stdout).unwrap().trim_end(),
+        message.clone()
+    );
```
thomaseizinger, line 435: Most of the other tests in here utilize some kind of client library to test the container functionality. Ideally, these tests also act as a form of documentation how to best use these images. Shelling out to bash doesn't feel like the way people would usually use this :)

src/clients/cli.rs
```diff
@@ -189,6 +189,43 @@ impl Docker for Cli {
         info.network_settings.ports.into_ports()
     }
 
+    fn gateway(&self, id: &str) -> String {
```
thomaseizinger, line 192: Two things regarding these functions:

1. I am not totally convinced we need them and would prefer if we keep the API surface of this traits as small as possible. Like I commented on the test, I believe using docker networks and default ports should make it possible to let the zookeeper and kafka image interact without having to expose this. That would also have the benefit that we provide a better out-of-the-box experience compared to what users currently have to do in a test to set things up properly.

2. IF it turns out that we still need these functions then we should at least reduce the duplication within the implementation. Currently, we have the code for calling `inspect` three times.

thomaseizinger (changes requested):
Hi @devcdcc !

Thank you for contributing! Please see the inline comments :)

devcdcc:
> Hi @devcdcc !
> 
> Thank you for contributing! Please see the inline comments :)

I am working on it, i just need to include the kafka driver.

thomaseizinger:
Closing because stale.

PR #238, opened 2020-12-24
[images] Allow custom version for Postgres

lloydmeta:
Adds a helper to allow customising the major version of the Psql image.


thomaseizinger (approved):
Thanks for the contribution!

bors r+

PR #245, opened 2021-01-15
Upgrade tokio

rishflab:
Upgrade tokio and request

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ rusoto_sqs = "0.45"
 spectral = "0.6"
 reqwest = { version = "0.10", features = ["blocking"] }
 json = "0.12"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -29,9 +29,9 @@ rusoto_credential = "0.45"
 rusoto_dynamodb = "0.45"
 rusoto_sqs = "0.45"
 spectral = "0.6"
-reqwest = { version = "0.10", features = ["blocking"] }
+reqwest = { version = "0.1", features = ["blocking"] }
```
thomaseizinger, line 32: This should be version 0.11.

thomaseizinger (changes requested):
Do the two commits work in isolation? If not, please squash them into one.

PR #246, opened 2021-01-25
Add udp and sctp support to port mappings

### LINKED ISSUE #234: Allow port forwarding for udp ports
I'm trying to test a DNS server I build and for this to work I need to forward `upd` ports from inside the container to my host system.
The docker cli syntax to forward udp ports is `-p 5353:53/udp`. Currently, this is not possible with testcontainer, because `Port` is defined as:
```
pub struct Port {
    pub local: u16,
    pub internal: u16,
}
```
I'm willing to provide a PR for this, but I think we should discuss possible implementations first, because it will most probably result in a breaking change.

kunerd:
Resolves #234

### DIFF
--- a/src/core/container.rs
+++ b/src/core/container.rs
@@ -1,5 +1,5 @@
 use crate::{
-    core::{docker::Docker, env::Command, image::WaitFor},
+    core::{docker::Docker, ports::{Ports, MapToHostPort}, env::Command, image::WaitFor},
     Image,
 };
 use std::{fmt, marker::PhantomData};
@@ -128,13 +128,17 @@ impl<'d, I> Container<'d, I> {
     /// This method panics if the given port is not mapped.
     /// Testcontainers is designed to be used in tests only. If a certain port is not mapped, the container
     /// is unlikely to be useful.
-    pub fn get_host_port(&self, internal_port: u16) -> u16 {
+    pub fn get_host_port<T>(&self, internal_port: T) -> T
+    where
+        T: fmt::Debug,
+        Ports: MapToHostPort<T>
+    {
         self.docker_client
             .ports(&self.id)
-            .map_to_host_port(internal_port)
+            .map_to_host_port(&internal_port)
             .unwrap_or_else(|| {
                 panic!(
-                    "container {} does not expose port {}",
+                    "container {:?} does not expose port {:?}",
                     self.id, internal_port
                 )
             })
--- a/src/core/container.rs
+++ b/src/core/container.rs
@@ -1,5 +1,10 @@
 use crate::{
-    core::{docker::Docker, ports::{Ports, MapToHostPort}, env::Command, image::WaitFor},
+    core::{
+        docker::Docker,
+        env::Command,
+        image::WaitFor,
+        ports::{MapToHostPort, Ports},
+    },
     Image,
 };
 use std::{fmt, marker::PhantomData};
@@ -131,7 +136,7 @@ impl<'d, I> Container<'d, I> {
     pub fn get_host_port<T>(&self, internal_port: T) -> T
     where
         T: fmt::Debug,
-        Ports: MapToHostPort<T>
+        Ports: MapToHostPort<T>,
     {
         self.docker_client
             .ports(&self.id)
--- a/src/core/ports.rs
+++ b/src/core/ports.rs
@@ -12,7 +12,7 @@ struct Udp(u16);
 pub struct Ports {
     tcp: HashMap<Tcp, Tcp>,
     udp: HashMap<Udp, Udp>,
-    sctp: HashMap<Sctp, Sctp>
+    sctp: HashMap<Sctp, Sctp>,
 }
 
 impl Ports {
@@ -21,7 +21,8 @@ impl Ports {
         let udp = HashMap::new();
         let sctp = HashMap::new();
 
-        ports.into_iter()
+        ports
+            .into_iter()
             .filter_map(|(internal, external)| {
                 // internal is ')8332/tcp', split off the protocol ...
                 let mut iter = internal.split('/');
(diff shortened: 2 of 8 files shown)


kunerd:
Maybe a better solution for the PortMapping would look like this:
```
enum Protocol {
    Tcp,
    Udp,
    Sctp
}

struct Ports {
    host,
    container
}

struct PortMapping {
    protocol: Protocol,
    ports: Ports
}

**Comment by thomaseizinger** [comment]
> Do we really need the lookup of the mapped host port for an internal port?
> [https://github.com/testcontainers/testcontainers-rs/blob/dev/src/core/docker.rs#L82](https://github.com/testcontainers/testcontainers-rs/blob/dev/src/core/docker.rs?rgh-link-date=2021-01-25T21%3A40%3A02Z#L82)

Yes we need that otherwise we don't know what ports the OS assigned to us if we start the containers with `-P`.

Also, have you tried running the tests locally with `cargo test`? You seem to have some compilation errors :)

**Comment by kunerd** [comment]
I ok, now this makes sense. I could imagine two interface to get the mapped host port.  Either something like:
```
enum Port {
     Tcp(u16),
     Udp(u16),
     Sctp (u16)
}
....
pub fn map_to_host_port(&self, internal_port: u16) -> Option<Port> {
}
```
or 
```
pub fn map_to_host_port_tcp(&self, internal_port: u16) -> Option<Port> {
}
pub fn map_to_host_port_udp(&self, internal_port: u16) -> Option<Port> {
}
```

**Comment by thomaseizinger** [comment]
Thanks for your proposal!

I think it will be valuable if the default case (a TCP port mapping) is easily accessible. As such, what do you think about an API along the lines of:

```rust
enum Protocol {
	Tcp,
	Udp,
	Sctp
}

pub fn map_to_host_port(&self, internal_port: u16) -> Option<(u16, Protocol)> {
    ...
}
```

A caller could then simply discard the protocol if they don't care about it (or if they know that it is always a specific one like this:

```rust
let (port, _) = container.map_to_host_port(8080).unwrap();
```

This actually got me thinking, do we ever care about which protocol this port was mapped on? I feel like in the context of testcontainers, we actually have an expectation in that regard.

The postgres container for example will always expose a TCP port mapping on port 5432. The information of `Protocol::Tcp` that would be returned by the above function is pretty useless. I already know that from the context of the test.

What I am trying to say is: Can we maybe just leave the mapping API as is and simply search for the mapped port in the `PortMapping` response and discard the protocol information?

kunerd:
> What I am trying to say is: Can we maybe just leave the mapping API as is and simply search for the mapped port in the `PortMapping` response and discard the protocol information?
That would be nice, but unfortunately it's possible to bind different protocols to the same port. Something as `-p 8080:80 -p 8080:80 udp` is totally valid. I will have another look. I also updated my post above, because I meant returning an `Option` with a `u16` instead of a `Port`:
``` rust
pub fn map_to_host_port_tcp(&self, internal_port: u16) -> Option<u16> {
    ...
}
```

thomaseizinger:
> > What I am trying to say is: Can we maybe just leave the mapping API as is and simply search for the mapped port in the `PortMapping` response and discard the protocol information?
> 
> That would be nice, but unfortunately it's possible to bind different protocols to the same port. Something as `-p 8080:80 -p 8080:80 udp` is totally valid. I also updated my post above, because I meant the `get_host_port` function and it should return an `Option` with a `u16` instead of a `Port`.

That is interesting! Thanks for sharing!

I think we might be able to achieve a nice API with some trait magic:

Imagine `get_host_port` with a signature like:

```rust
fn get_host_port<P>(&self, internal_port: P) -> u16 where P: Into<(u16, Protocol)> {
	let (port, protocol) = internal_port.into();
	// ...
}
```

We can then implement the following traits:

```rust
impl Into<(u16, Protocol)> for &str {
	fn into(self) -> (u16, Protocol) {
		// parse self as '{port}/{udp,tcp,sctp}'
		// panic in parsing code if necessary, we are only within tests so panicking is fine
		// return tuple of port and protocol
	}
}

impl Into<(u16, Protocol)> for u16 {
	// default to TcpProtocol for implementations of port
}
```

The user can then call the function in various ways:

```rust
container.get_host_port(8080);
container.get_host_port("8080/tcp");
container.get_host_port("8080/udp");
```

kunerd:
Hi, the current solution should be fully backward compatible. I think it still needs some refactoring and some test, but then it should be ready to go. Would like to here your thoughts.

src/core/container.rs
```diff
@@ -69,7 +69,11 @@ where
     /// This method does **not** magically expose the given port, it simply performs a mapping on
     /// the already exposed ports. If a docker image does not expose a port, this method will not
     /// be able to resolve it.
-    pub fn get_host_port(&self, internal_port: u16) -> Option<u16> {
+    pub fn get_host_port<T>(&self, internal_port: T) -> Option<u16>
+    where
+        T: Into<u16> + Copy,
```
thomaseizinger, line 74: If this is only `Into<u16>`, how are we going to map a non-TCP port? As far as I understand, the user needs to able to specify the protocol here.

thomaseizinger (commented):
Thanks for working on this!

Please note that parts of the codebase changed quite a bit with the recent work on an async interface and the shiplift integration. Can you rebase this PR onto latest `dev` please?

I've also adapted the parsing code for the ports a bit so it is hopefully easier to adapt to extract the protocol.

I think there are two things we need to still adapt here:

1. Users need to be able to specify the protocol when resolving the port. Currently the trait bound is only `Into<u16>`.
2. We should also adapt the API of `RunArgs` such that users can configure arbitrary port mappings there.

Let me know how the rebase goes. If it is too cumbersome, it might be easier to just start again from current dev! Sorry for the inconvenience.

src/core/container.rs
thomaseizinger: We might need to create our own trait here due to coherence rules.

kunerd: The `Into<u16>` is misleading here, it only is necessary for the debug print, I think it would be better to switch to `{:?}` there. The next line is the important one, because it let's us forward the call to the type specific implementation of `map_to_host_port`. You can try it in this Playground example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=100daefbe4f085548bfc3c620f522b82
I should even change the return value into `Option<T>`, so that the user couldn't accidentally mix up different port types - missed that here, because it was a bit late as a have done the commit.

kunerd:
Now, only additional documentation and some test are missing.

kunerd:
I have tried it with my original use case and it seems to work as expected. But, I think the `RunArgs::with_mapped_port()` needs to be more ergonomic to use.

src/core/container.rs
```diff
@@ -128,13 +133,17 @@ impl<'d, I> Container<'d, I> {
     /// This method panics if the given port is not mapped.
     /// Testcontainers is designed to be used in tests only. If a certain port is not mapped, the container
     /// is unlikely to be useful.
-    pub fn get_host_port(&self, internal_port: u16) -> u16 {
+    pub fn get_host_port<T>(&self, internal_port: T) -> T
```
thomaseizinger, line 136: How does one go about calling this API? From what I can see, everything that implements `MapToHostPort` is actually private and can't be used by users outside of the library.

To make sure this doesn't happen, can you add an example in `examples/` (dir doesn't exist yet). Rust examples are compiled as part of the tests so that will be a good way of ensuring people can actually use the API :)

src/core/ports.rs
```diff
@@ -1,18 +1,33 @@
...
+        let tcp = HashMap::new();
+        let udp = HashMap::new();
+        let sctp = HashMap::new();
+
+        ports
             .into_iter()
             .filter_map(|(internal, external)| {
-                // internal is '8332/tcp', split off the protocol ...
-                let internal = internal.split('/').next()?;
+                // internal is ')8332/tcp', split off the protocol ...
```
thomaseizinger, line 27: A typo got into this comment :)

```suggestion
                // internal is '8332/tcp', split off the protocol ...
```

thomaseizinger (changes requested):
Thank you!

I think we still need to iron out a few details before this can merge :)

DDtKey:
Closing due to inactivity and outdated state. 
Feel free to open a new PR, but keep in mind that we also going to deprecate existing `cli` client

PR #250, opened 2021-01-27
Allow using podman

ctron:
This change allows using podman: https://podman.io/

`podman` is a drop in CLI replacement for `docker`, so the change should be minimal.

My main usecase for this is that `docker` fails to run Postgres images in GitHub actions. When the container starts, it shuts down the database and then execs into the main database process. This gets stuck when run in Docker, probably because GitHub actions already runs inside Docker.

Switching to podman, which is available by default on Ubuntu 20.04, and also available by default on GitHub actions workers, fixes the problem.

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -1,6 +1,7 @@
 use crate::core::{self, Container, Docker, Image, Logs, RunArgs};
 use std::{
     collections::HashMap,
+    ffi::OsString,
     process::{Command, Stdio},
     sync::RwLock,
     thread::sleep,
@@ -13,7 +14,7 @@ const ZERO: Duration = Duration::from_secs(0);
 /// Implementation of the Docker client API using the docker cli.
 ///
 /// This (fairly naive) implementation of the Docker client API simply creates `Command`s to the `docker` CLI. It thereby assumes that the `docker` CLI is installed and that it is in the PATH of the current execution environment.
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct Cli {
     /// The docker CLI has an issue that if you request logs for a container
     /// too quickly after it was started up, the resulting stream will never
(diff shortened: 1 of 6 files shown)


ctron:
Just to be clear, the default is still `docker`. However, you can manually switch to `podman` by creating a client using `Cli::podman()`.

src/clients/cli.rs
```diff
@@ -25,9 +26,42 @@ pub struct Cli {
...
+    fn default() -> Self {
+        Self::docker()
+    }
 }
 
 impl Cli {
+    /// Create a new client, providing a custom `docker` command.
+    pub fn new<S>(command: S) -> Self
+    where
+        S: Into<OsString>,
```
thomaseizinger, line 42: What is the need for `OsString`? Why not just a normal `String`?

thomaseizinger (changes requested):
This looks mostly good to me, thank you for the contribution!

Two things:

- I would like to have the use of `OsString` clarified. As far as I understand, a normal `String` should be good enough for our purposes.
- Can you please add an entry to the CHANGELOG.md that documents this new feature?

src/clients/cli.rs
ctron: The `new` function of `Command` is defined as:

```rust
    pub fn new<S: AsRef<OsStr>>(program: S) -> Command {
        Command { inner: imp::Command::new(program.as_ref()) }
    }
```

I used an `OsString`, rather than `OsStr`, in order to clone it and be able to simply pass it into the `new` function, as it already is an `OsString`. This way testcontainers doesn't make an decision on the data type, as it only passes through the value from the user to the `Command`.

ctron:
I rebased and added a line in the changelog

ctron:
It looks like the travis job failed to an issue downloading the cache. I guess this simply needs to be re-started.

src/clients/cli.rs
thomaseizinger: Thanks for elaborating!

I saw that `String` also implements `AsRef<OsStr>` so I guess it doesn't matter as much when converting from a Rust String to an `OsString`. The other direction seems more troublesome because the `OsString` could contain non-UTF8 bytes.

Happy to stick to `OsString` here.

thomaseizinger:
> It looks like the travis job failed to an issue downloading the cache. I guess this simply needs to be re-started.

Yes, Travis has been causing problems for a while now but I haven't had the time lately to make the switch to GH actions. I am trying to do that today.

thomaseizinger:
I've switched everything over to GitHub actions. Can you please rebase onto dev again?

thomaseizinger:
@ctron I hope you don't mind that I rebased your branch and added a test :)

PR #253, opened 2021-02-01
Replace `wait_until_ready` with declarative `ready_conditions`

### LINKED ISSUE #39: Provide re-usable abstraction for additional waiting period
With the addition of the dynamodb image, we now have two images that require an "additional wait period" setting. The first one is bitcoind.

In both cases, this is exposed through an environment variable.

We should provide an abstraction that allows images to easily opt-into this behavior without implementing it completely themselves.

thomaseizinger:
Making this feature declarative has several advantages:

- It reduces the amount of code one needs to write for a new image.
- It avoids the need for an `ImageAsync` trait once we introduce that,
see discussions in #216 for details. cc @ywalterh
- It allows for a re-usable abstraction of waiting for a specific
amount of time, which fixes #39.

### DIFF
--- a/src/core/image.rs
+++ b/src/core/image.rs
@@ -1,4 +1,4 @@
-use crate::core::{Container, Docker};
+use std::{env::var, time::Duration};
 
 /// Represents a docker image.
 ///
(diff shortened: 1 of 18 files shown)


src/core/image.rs
```diff
@@ -63,18 +64,12 @@ where
     /// suddenly changed.
     fn descriptor(&self) -> String;
 
-    /// Blocks the current thread until the started container is ready.
+    /// Returns a list of conditions that need to be met before a started container is considered ready.
     ///
```
thomaseizinger, line 68: TODO: Clarify that the order of these conditions is important.

luckysori (approved):
Nice change!

ywalterh:
Nice, I'll rebase to try it out soon.

PR #258, opened 2021-02-02
Make `get_host_port` panic if the port cannot be mapped

### LINKED ISSUE #226: [RFC] `Container#get_host_port` should panic if the port was not mapped
Having panicking APIs is more convenient and this library is designed to be used within tests only so panicked should be fine.

thomaseizinger:
Closes #226.

@D4nte @luckysori any thoughts on why this change could be a really bad idea?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - How images express when a container is ready: Instead of implementing `wait_until_ready`, images now need to implement `ready_conditions` which returns a list of `WaitFor` instances.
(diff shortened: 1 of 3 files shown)


luckysori (approved):
> any thoughts on why this change could be a really bad idea?

Panicking seems a little aggressive, but it's true that it's unlikely that you can recover in a test if a port that is expected to be mapped is not.

thomaseizinger:
> > any thoughts on why this change could be a really bad idea?
> 
> Panicking seems a little aggressive, but it's true that it's unlikely that you can recover in a test if a port that is expected to be mapped is not.

Yeah if I want to make the API more ergonomic I can only either panic or assign a default value and the default value sounds even worse 😅
I just wanted to confirm that there isn't an obvioue usecase where you could recover.

PR #261, opened 2021-02-07
Upgrade default bitcoin image version to 0.21.0

bonomat:
With the upgrade to 0.21.0 a new startup command was introduced which allows us to execute a script when the startup sequence was completed. We simply echo `bitcoind startup sequence completed.`.
This allows us to remove `-debug` which keeps the bitcoind log _clean_.
Note: this release also removed the default wallet.

### DIFF
--- a/src/images/coblox_bitcoincore.rs
+++ b/src/images/coblox_bitcoincore.rs
@@ -5,6 +5,8 @@ use rand::{thread_rng, Rng};
 use sha2::Sha256;
 use std::{collections::HashMap, fmt};
 
+const BITCOIND_STARTUP_MESSAGE: &str = "bitcoind startup sequence completed.";
+
 #[derive(Debug)]
 pub struct BitcoinCore {
     tag: String,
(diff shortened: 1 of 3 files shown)


src/images/coblox_bitcoincore.rs
```diff
@@ -197,7 +198,7 @@ impl Image for BitcoinCore {
 
     fn ready_conditions(&self) -> Vec<WaitFor> {
         vec![
-            WaitFor::message_on_stdout("Flushed wallet.dat"),
+            WaitFor::message_on_stdout("bitcoind startup sequence completed."),
```
thomaseizinger, line 201: To make the relationship between these constants clearer, can we extract this into a non-pub `const` and use `format!` above to configure the command?

thomaseizinger (changes requested):
Thank you!

Can you also include a line in the CHANGELOG please with a link to `bitcoind`'s changelog?

thomaseizinger (approved):
Thanks, looks good!

D4nte (approved):

thomaseizinger:
@bonomat can you rebase this please?

bonomat:
🤦‍♂️rebased against master

PR #264, opened 2021-02-08
Fixing some failing tests

bonomat:
This fixes some failing tests: 

1. 58ff5c939c0520bcd38b774b231a3fe7ce45ef6c: Orientdb randomly fails in CI but not locally (tested on MacOS). 

I'll record issues for all the ignored tests once we are happy with this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,12 +28,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   We now wait for 2 seconds unconditionally after the specified message has been found.
 - Support for the `KEEP_CONTAINERS` env variable.
   The functionality of `KEEP_CONTAINERS=true` is superseded by `TESTCONTAINERS=keep`.
-  
+
 ### Fixed
 
-- Removing a docker container did not error if failed. This was fixed by asserting the daemon's response instead of 
-  just the status code: If a docker container was removed correctly using `rm -f -v <ID>` <ID> is printed on stdout. 
+- Removing a docker container did not error if failed. This was fixed by asserting the daemon's response instead of
+  just the status code: If a docker container was removed correctly using `rm -f -v <ID>` <ID> is printed on stdout.
   <ID> can either be the container name or its ID which is used within testcontainer-rs.
+- Fixed clippy warnings of camel case names containing a capitalized acronym.
 
 ## [0.12.0] - 2021-01-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   We now wait for 2 seconds unconditionally after the specified message has been found.
 - Support for the `KEEP_CONTAINERS` env variable.
   The functionality of `KEEP_CONTAINERS=true` is superseded by `TESTCONTAINERS=keep`.
+  
+### Fixed
+
+- Removing a docker container did not error if failed. This was fixed by asserting the daemon's response instead of 
+  just the status code: If a docker container was removed correctly using `rm -f -v <ID>` <ID> is printed on stdout. 
+  <ID> can either be the container name or its ID which is used within testcontainer-rs.
 
 ## [0.12.0] - 2021-01-27
 
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -277,7 +277,15 @@ impl Docker for Cli {
             .arg(id)
             .output()
             .expect("Failed to execute docker command");
-        assert!(output.status.success(), "Failed to remove docker container");
+        let error_msg = "Failed to remove docker container";
+        assert!(output.status.success(), error_msg);
+        // The container's id is printed on stdout if it was removed successfully.
+        assert!(
+            String::from_utf8(output.stdout)
+                .expect("Could not decode daemon's response.")
+                .contains(id),
+            error_msg
+        );
     }
 
     fn stop(&self, id: &str) {
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -278,12 +278,13 @@ impl Docker for Cli {
             .output()
             .expect("Failed to execute docker command");
         let error_msg = "Failed to remove docker container";
-        assert!(output.status.success(), error_msg);
+        assert!(output.status.success(), "{}", error_msg);
         // The container's id is printed on stdout if it was removed successfully.
         assert!(
             String::from_utf8(output.stdout)
                 .expect("Could not decode daemon's response.")
                 .contains(id),
+            "{}",
             error_msg
         );
     }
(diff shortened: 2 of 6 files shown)


thomaseizinger:
> [`b81dd8c` (#264)](https://github.com/testcontainers/testcontainers-rs/pull/264/commits/b81dd8cd90e25168d197a6d6fb0e82f4ecb0a4f3): Removing an non-existing container on MacOS does not fail but prints an error message on stdout.
>        E.g. removing non-existing container with id/name `foo` would print: `Error: No such container: foo`

It should panic because it should exit with a non-zero exit code and we panic on that. I am curious why that wouldn't work on MacOS?

> ce45d24 (#264): Podman might not exist on every system. I'm ignoring the test if it does not exist.

Docker might also not exist on every system but we expect it to be installed if you are hacking on testcontainers so you can run the tests. I don't think we should be making an exception for podman just because it is less popular.

Could you deal with that locally please?

Alternatively we could feature-gate it but it feels wrong because it makes things only harder for users at no benefit.

> eaddf9e (#264): Orientdb randomly fails in CI but not locally (tested on MacOS).

Fine with me.
We should write up maintenance-guidelines where we ask contributors to be ready to fix bugs in these tests if they occur otherwise this is not going to scale I am afraid :/

bonomat:
> It should panic because it should exit with a non-zero exit code and we panic on that. I am curious why that wouldn't work on MacOS?

I verified it locally, it does not return a non-zero exit code. Unfortunately I could not find any documentation/man page for this. 

> Alternatively we could feature-gate it but it feels wrong because it makes things only harder for users at no benefit.

I never heard of `podman` before. I'll go with feature-gate.

thomaseizinger:
> I never heard of `podman` before. I'll go with feature-gate.

Neither have I before I was approached with the feature request of adding it. However, that doesn't make what I wrote above any less true: A feature-gate only has downsides from the PoV of a library user. As long as they don't want to use podman, they don't need to care about it but if they want to, they suddenly have to enable a feature flag.

I think it is acceptable to require certain things to be installed when you are _working_ on the library itself. If you don't want to install it, there are several ways of how to deal with the failing test:

- ignore it locally but don't include it in commits
- run tests without bailing early on failure and make sure only the podman one fails
- delete the file locally and don't include that in commits

thomaseizinger:
> I verified it locally, it does not return a non-zero exit code. Unfortunately I could not find any documentation/man page for this.

Really? That is super weird. The CLI is cross-platform AFAIK. Just to double check, the output of

```
docker rm foobar; echo $?
```
is
```
Error: No such container: foobar
0
```

on your machine?

bonomat:
> > I verified it locally, it does not return a non-zero exit code. Unfortunately I could not find any documentation/man page for this.
> 
> Really? That is super weird. The CLI is cross-platform AFAIK. Just to double check, the output of
> 
> ```
> docker rm foobar; echo $?
> ```
> 
> is
> 
> ```
> Error: No such container: foobar
> 0
> ```
> 
> on your machine?

There we go, found the issue: 
If you do `docker rm -f -v foo` (which we do in the code)
I get: 
image

While if you don't use `-f` the it returns a non-zero value: 

image

Is that different on Ubuntu?

bonomat:
Please excuse, my comment might have come out in the wrong way. 

> [...] As long as they don't want to use podman, they don't need to care about it but if they want to, they suddenly have to enable a feature flag.
> [...] 

Personally I think a feature flag for this is the way to go as it is an additional feature. Do I see it correctly that we did not release a version with `prodman`. If so, the change would not be sudden for users. 

I'm also happy to just drop this commit. Eitherway is fine with me. I'd like to get a successful build of this library if possible :D

thomaseizinger:
> [snip]
> Is that different on Ubuntu?

On Manjaro it certainly isn't. Can't speak about Ubuntu :D

How did that ever work? :thinking: 

We do need `-f` to remove running containers. Can we patch the code such that instead of asserting on the exit status, we assert on the stderr text to know whether or not it worked? As in, `panic` if the stderr text starts with `Error: No such container`. That should hopefully work everywhere then.

> Personally I think a feature flag for this is the way to go as it is an additional feature. Do I see it correctly that we did not release a version with `prodman`. If so, the change would not be sudden for users.
> 
> I'm also happy to just drop this commit. Eitherway is fine with me. I'd like to get a successful build of this library if possible :D

You are right that it is conceptually a _feature_!

One of the main goals of testcontainers is to keep the configuration surface as minimal as possible to maximize ease of use. As such, there has to be some kind of benefit for users if they have to deal with added complexity. With feature flag, that benefit is usually faster compile-times due to a smaller dependency tree.

But in this case, having to worry about feature-flags brings no benefit to the user. Instead it brings benefits to the people working on the library by them not having to install `podman` and I think that this trade-off is simply not worth it.

As such, my suggestion is to revert the changes to the podman test. It should already be possible to successfully _build_ the library on MacOS. To run the tests, you can either install (and annoyingly configure) podman or trust that if this is the only test that fails but passes in CI that everything is good to go!

Thanks for taking the time to investigate and fix these tests :)

bonomat:
Sorry for the delay: 

> How did that ever work? 🤔
😅 
Does the test run for you? 

> We do need -f to remove running containers. Can we patch the code such that instead of asserting on the exit status, we assert on the stderr text to know whether or not it worked? As in, panic if the stderr text starts with Error: No such container. That should hopefully work everywhere then.

Can definitely do that. 👍

> As such, my suggestion is to revert the changes to the podman test. 

Will do this 👍

thomaseizinger:
>  Does the test run for you? It seems to be working on ubuntu (see github actions)

Whatever worked, it seems to work no longer :sweat_smile: 

https://github.com/testcontainers/testcontainers-rs/pull/267/checks?check_run_id=1924476222#step:5:549

bonomat:
Yay, all green 🥳

CHANGELOG.md
```diff
@@ -28,12 +28,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-  
+
 ### Fixed
 
-- Removing a docker container did not error if failed. This was fixed by asserting the daemon's response instead of 
-  just the status code: If a docker container was removed correctly using `rm -f -v <ID>` <ID> is printed on stdout. 
+- Removing a docker container did not error if failed. This was fixed by asserting the daemon's response instead of
+  just the status code: If a docker container was removed correctly using `rm -f -v <ID>` <ID> is printed on stdout.
   <ID> can either be the container name or its ID which is used within testcontainer-rs.
+- Fixed clippy warnings of camel case names containing a capitalized acronym.
```
thomaseizinger, line 37: Would have been good to mention the actual breaking changes here, i.e. the renaming of `ElasticMq` etc.

Not a blocker for merging though, will adapt in a future PR!

Thanks!

thomaseizinger (approved):
bors r+

PR #269, opened 2021-02-18
Simplify `Image` trait

### LINKED ISSUE #263: Remodel `Image` trait
Get rid of:

- associated type for `Args` by introducing a dedicated string-iterator type that can be constructed from a `vec` and whatever else we need
- associated types for `Volumes` and `Envs` by introducing a dedicated (string, string)-iterator type
- `with_args` function: nothing in the lib uses that
- `with_entrypoint` function: nothing in the lib uses that either
- get rid of `with_args` and `args` by providing dedicated overrides for `Docker::run` where we construct `Args::default` for them or where they can pass their own instance based on the associated type.
- also consider removing fixed `Default` bound and add it to the fn on the `Docker` trait

Make overriding `volumes` and `envs` optional by returning an empty iterator.

thomaseizinger:
Resolves #263.


thomaseizinger:
Superseded by #295.

PR #278, opened 2021-03-12
Add EventStoreDB image.

YoEight:


src/images/eventstoredb.rs
```diff
@@ -0,0 +1,102 @@
...
+    env_vars: HashMap<String, String>,
+}
+
+impl EventStoreDB {
+    pub fn insecure_mode(mut self) -> Self {
+        self.env_vars
+            .insert("EVENTSTORE_INSECURE".to_string(), "true".to_string());
+        self.env_vars.insert(
+            "EVENTSTORE_ENABLE_ATOM_PUB_OVER_HTTP".to_string(),
+            "true".to_string(),
```
thomaseizinger, line 32: I guess these could actually use `self.add_env_var` to avoid some of the duplication?

src/images/eventstoredb.rs
```diff
@@ -0,0 +1,102 @@
...
+    type EnvVars = HashMap<String, String>;
+    type Volumes = HashMap<String, String>;
+    type EntryPoint = std::convert::Infallible;
+
+    fn descriptor(&self) -> String {
+        format!("{}:{}", CONTAINER_IDENTIFIER, &self.tag)
+    }
+
+    fn ready_conditions(&self) -> Vec<WaitFor> {
+        vec![WaitFor::message_on_stdout("SPARTA")]
```
thomaseizinger, line 74: Is this a joke or does it actually emit "SPARTA" when the image is ready? :laughing:

src/images.rs
```diff
@@ -1,6 +1,7 @@
 pub mod coblox_bitcoincore;
 pub mod dynamodb_local;
 pub mod elasticmq;
+pub mod eventstoredb;
```
thomaseizinger, line 4: We usually name these modules after the Docker image name which in this case is `eventstore`.

thomaseizinger (commented):
Thank you for contributing and sorry for the late review!

thomaseizinger:
Closing because stale.

Issue #105, opened 2019-09-10
Investigate why we depend on OpenSSL


thomaseizinger:
Apparently, depending on testcontainers brings in a dependency on openssl. That shouldn't be the case because we don't do any networks requests (yet).

Investigate and fix.

D4nte:
It's due to web3 and to bitcoin_rpc_client that brings in json_client (that brings in tls)

thomaseizinger:
Interesting! Those should be devDependencies I think and hence should not
affect downstream users.

D4nte:
Yep, actually it does not :) see comit-network/comit-rs#1376

thomaseizinger:
Closing then :)

PR #280, opened 2021-03-24
Removed openssl dep

aig787:
Looks like a dependency on openssl creeped in with the addition of shiplift. This assumed #105 still holds and that openssl is not required. It still built without the feature enabled.


thomaseizinger (commented):
Thanks for the contribution!

Can you squash the two commits into one please?

aig787:
I tried a force push but remote rejected it earlier, seems fixed now. 🤷‍♂️ must have been something on my end

thomaseizinger (approved):

thomaseizinger:
bors retry

PR #282, opened 2021-04-01
Add Kafka image confluentinc/cp-kafka

saitoy1:
This PR adds a Kafka image [confluentinc/cp-kafka](https://hub.docker.com/r/confluentinc/cp-kafka/) to the crate. I was aware that there was an existing #229 and wanted to help cross the finish line, but decided to start from scratch for the following reasons:
- Other client libraries for `testcontainers`, e.g. Java, Python, and NodeJS, use `confluentinc/cp-kafka` whereas #229 uses `bitnami/kafka`. For consistency with other client libraries, this PR has chosen `confluentinc/cp-kafka`.
- #229 shells out to bash execute `kafka-console-consumer` and `kafka-console-consumer` during testing. The test in this PR uses the `rdkafka` crate to interact with a Kafka broker, illustrating how that can be done without bash.
- #229 has been inactive for the last couple of months and I needed the Kafka image urgently.

That said, I have no intention of devaluing or competing with #229, so please feel free to take whichever PR that makes sense to the crate.

Finally, for adding tests, I have noticed that `tests/images.rs` is gradually becoming a monolith of tests for various images. Maybe that's actually intended, but I had hard time figuring out where to place a test for the Kafka image. Furthermore, the test for the Kafka image is longer and I did not feel comfortable squeezing it in `tests/images.rs`. I thought it may be easier for someone looking for tests as a form of documentation to locate them if tests are organized by images. To this end, I have created a separate test file `tests/images/kafka.rs` just to contain the test for the Kafka image and `tests/tests.rs` to tell Cargo where to find the test. If this is not something the crate wants in the long term, I'm happy to move the test for the Kafka image to `tests/images.rs` and remove `tests/images/kafka.rs` and `tests/tests.rs`.

### DIFF
--- a/src/images/kafka.rs
+++ b/src/images/kafka.rs
@@ -1,0 +1,123 @@
+use crate::{core::WaitFor, Image};
+use std::collections::HashMap;
+
+const CONTAINER_IDENTIFIER: &str = "confluentinc/cp-kafka";
+const DEFAULT_TAG: &str = "6.1.1";
+
+pub const KAFKA_PORT: u16 = 9093;
+const ZOOKEEPER_PORT: u16 = 2181;
+
+#[derive(Clone, Debug, Default)]
+pub struct KafkaArgs;
+
+impl IntoIterator for KafkaArgs {
+    type Item = String;
+    type IntoIter = ::std::vec::IntoIter<String>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        vec![
+            "/bin/bash".to_owned(),
+            "-c".to_owned(),
+            format!(
+                r#"
+echo 'clientPort={}' > zookeeper.properties;
+echo 'dataDir=/var/lib/zookeeper/data' >> zookeeper.properties;
+echo 'dataLogDir=/var/lib/zookeeper/log' >> zookeeper.properties;
+zookeeper-server-start zookeeper.properties &
+. /etc/confluent/docker/bash-config &&
+/etc/confluent/docker/configure &&
+/etc/confluent/docker/launch"#,
+                ZOOKEEPER_PORT
+            ),
+        ]
+        .into_iter()
+    }
+}
+
+#[derive(Debug)]
+pub struct Kafka {
+    arguments: KafkaArgs,
+    env_vars: HashMap<String, String>,
+    tag: String,
+}
+
+impl Kafka {
+    pub fn with_tag<T: Into<String>>(self, tag: T) -> Self {
+        Self {
+            tag: tag.into(),
+            ..self
+        }
+    }
+}
+
+impl Default for Kafka {
+    fn default() -> Self {
+        let mut env_vars = HashMap::new();
+
+        env_vars.insert(
+            "KAFKA_ZOOKEEPER_CONNECT".to_owned(),
+            format!("localhost:{}", ZOOKEEPER_PORT),
+        );
+        env_vars.insert(
+            "KAFKA_LISTENERS".to_owned(),
+            format!("PLAINTEXT://0.0.0.0:{},BROKER://0.0.0.0:9092", KAFKA_PORT),
+        );
+        env_vars.insert(
+            "KAFKA_LISTENER_SECURITY_PROTOCOL_MAP".to_owned(),
+            "BROKER:PLAINTEXT,PLAINTEXT:PLAINTEXT".to_owned(),
+        );
+        env_vars.insert(
+            "KAFKA_INTER_BROKER_LISTENER_NAME".to_owned(),
+            "BROKER".to_owned(),
+        );
+        env_vars.insert(
+            "KAFKA_ADVERTISED_LISTENERS".to_owned(),
+            format!(
+                "PLAINTEXT://localhost:{},BROKER://localhost:9092",
+                KAFKA_PORT
+            ),
+        );
+        env_vars.insert("KAFKA_BROKER_ID".to_owned(), "1".to_owned());
+        env_vars.insert(
+            "KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR".to_owned(),
+            "1".to_owned(),
+        );
+
+        Self {
+            arguments: KafkaArgs::default(),
+            env_vars,
+            tag: DEFAULT_TAG.to_owned(),
+        }
+    }
+}
+
+impl Image for Kafka {
+    type Args = KafkaArgs;
+    type EnvVars = HashMap<String, String>;
+    type Volumes = HashMap<String, String>;
+    type EntryPoint = std::convert::Infallible;
+
+    fn descriptor(&self) -> String {
+        format!("{}:{}", CONTAINER_IDENTIFIER, &self.tag)
+    }
+
+    fn ready_conditions(&self) -> Vec<WaitFor> {
+        vec![WaitFor::message_on_stdout("Creating new log file")]
+    }
+
+    fn args(&self) -> <Self as Image>::Args {
+        self.arguments.clone()
+    }
+
+    fn env_vars(&self) -> Self::EnvVars {
+        self.env_vars.clone()
+    }
+
+    fn volumes(&self) -> Self::Volumes {
+        HashMap::new()
+    }
+
+    fn with_args(self, arguments: <Self as Image>::Args) -> Self {
+        Self { arguments, ..self }
+    }
+}
--- a/tests/kafka.rs
+++ b/tests/kafka.rs
@@ -1,0 +1,75 @@
+use futures::StreamExt;
+use rdkafka::{
+    config::ClientConfig,
+    consumer::{stream_consumer::StreamConsumer, Consumer},
+    producer::{FutureProducer, FutureRecord},
+    Message,
+};
+use std::time::Duration;
+
+use testcontainers::{clients, core::RunArgs, images::kafka};
+
+#[tokio::test]
+async fn test_produce_and_consume_messages() {
+    let docker = clients::Cli::default();
+    let kafka_node = docker.run_with_args(
+        kafka::Kafka::default(),
+        RunArgs::default().with_mapped_port((kafka::KAFKA_PORT, kafka::KAFKA_PORT)),
+    );
+
+    let bootstrap_servers = format!("localhost:{}", kafka_node.get_host_port(kafka::KAFKA_PORT));
+
+    let producer = ClientConfig::new()
+        .set("bootstrap.servers", &bootstrap_servers)
+        .set("message.timeout.ms", "5000")
+        .create::<FutureProducer>()
+        .expect("Failed to create Kafka FutureProducer");
+
+    let consumer = ClientConfig::new()
+        .set("group.id", "testcontainer-rs")
+        .set("bootstrap.servers", &bootstrap_servers)
+        .set("session.timeout.ms", "6000")
+        .set("enable.auto.commit", "false")
+        .set("auto.offset.reset", "earliest")
+        .create::<StreamConsumer>()
+        .expect("Failed to create Kafka StreamConsumer");
+
+    let topic = "test-topic";
+    consumer
+        .subscribe(&[&topic])
+        .expect("Failed to subscribe to a topic");
+
+    let number_of_messages_to_produce = 5_usize;
+    let expected: Vec<String> = (0..number_of_messages_to_produce)
+        .map(|i| format!("Message {}", i))
+        .collect();
+
+    for (i, message) in expected.iter().enumerate() {
+        producer
+            .send(
+                FutureRecord::to(&topic)
+                    .payload(message)
+                    .key(&format!("Key {}", i)),
+                Duration::from_secs(0),
+            )
+            .await
+            .unwrap();
+    }
+
+    let mut message_stream = consumer.stream();
+    for produced in expected {
+        let borrowed_message = tokio::time::timeout(Duration::from_secs(10), message_stream.next())
+            .await
+            .unwrap()
+            .unwrap();
+
+        assert_eq!(
+            produced,
+            borrowed_message
+                .unwrap()
+                .payload_view::<str>()
+                .unwrap()
+                .unwrap()
+        );
+    }
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ mongodb = "2.0.0-alpha"
 orientdb-client = "0.5"
 postgres = "0.19"
 pretty_env_logger = "0.4"
+rdkafka = "0.26"
 redis = "0.20"
 reqwest = { version = "0.11", features = [ "blocking" ] }
 rusoto_core = "0.46"
--- a/src/images.rs
+++ b/src/images.rs
@@ -3,6 +3,7 @@ pub mod dynamodb_local;
 pub mod elasticmq;
 pub mod generic;
 pub mod hello_world;
+pub mod kafka;
 pub mod mongo;
 pub mod orientdb;
 pub mod parity_parity;


saitoy1:
I see the following error in `msrv`:
```
   Compiling socket2 v0.4.0
error[E0658]: `match` is not allowed in a `const fn`
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/socket2-0.4.0/src/lib.rs:156:9
    |
156 | /         match address {
157 | |             SocketAddr::V4(_) => Domain::IPV4,
158 | |             SocketAddr::V6(_) => Domain::IPV6,
159 | |         }
    | |_________^
    |
    = note: see issue #49146 <https://github.com/rust-lang/rust/issues/49146> for more information

error: aborting due to previous error
```
My understanding is that if `cargo build` as performed in `msrv` fails, it shouldn't be able to start running tests, but the `test` stage got a green check. Could you help me better understand what this step `msrv` does so that I can further look into this failure? Thanks.

src/images/kafka.rs
```diff
@@ -0,0 +1,112 @@
...
+use std::collections::HashMap;
+
+const CONTAINER_IDENTIFIER: &str = "confluentinc/cp-kafka";
+const DEFAULT_TAG: &str = "6.1.1";
+
+pub const DEFAULT_BROKER_LISTENER_PORT: u16 = 9092;
+pub const DEFAULT_LOCALHOST_LISTENER_PORT: u16 = 19092;
+
+#[derive(Clone, Debug, Default)]
+pub struct KafkaArgs(pub Vec<String>);
```
thomaseizinger, line 11: `KafkaArgs` should be a strongly-typed struct where we only convert to strings using `IntoIterator`. Checkout [this](https://github.com/testcontainers/testcontainers-rs/blob/df3805945219040d38c9112c0579b898249cdf3b/src/images/coblox_bitcoincore.rs#L100-L189) for some inspiration.

thomaseizinger (changes requested):
Hi @saitoy1!

Thank you very much for your contribution, it is most welcome!
Please see the comments in the review. Generally, we should be making as much use of the type system as possible. At the moment, there are a lot of strings being passed around and _using_ the `Kafka` struct is not very convenient because the caller has to still specify all sorts of stuff.

The main idea of the abstractions within testcontainers is that - if possible - just creating a `default()` instance of the image gets you something that is working. In other words, what we strive for is:

```rust
let docker = Cli::default();
let container = docker.run(Kafka::default());
```

Should give you a running container.

It might be tricky to achieve that because as far as I understand, kafka always depends on zookeeper but even in that case, there should be minimal amounts of configuration needed to get a running container. Hence, at the most, we should need to specify the zookeeper port and the rest happens by default.

Let me know if you need more input :)

saitoy1:
> Hi @saitoy1!
> 
> Thank you very much for your contribution, it is most welcome!
> Please see the comments in the review. Generally, we should be making as much use of the type system as possible. At the moment, there are a lot of strings being passed around and _using_ the `Kafka` struct is not very convenient because the caller has to still specify all sorts of stuff.
> 
> The main idea of the abstractions within testcontainers is that - if possible - just creating a `default()` instance of the image gets you something that is working. In other words, what we strive for is:
> 
> ```rust
> let docker = Cli::default();
> let container = docker.run(Kafka::default());
> ```
> 
> Should give you a running container.
> 
> It might be tricky to achieve that because as far as I understand, kafka always depends on zookeeper but even in that case, there should be minimal amounts of configuration needed to get a running container. Hence, at the most, we should need to specify the zookeeper port and the rest happens by default.
> 
> Let me know if you need more input :)

Thank you so much for your detailed feedback! That makes sense. Let me work on pushing as many configurations as possible to `Kafka::default` so the user does not need to specify them. I'll also make sure to address other feedback, too.

saitoy1:
> Hi @saitoy1!
> 
> Thank you very much for your contribution, it is most welcome!
> Please see the comments in the review. Generally, we should be making as much use of the type system as possible. At the moment, there are a lot of strings being passed around and _using_ the `Kafka` struct is not very convenient because the caller has to still specify all sorts of stuff.
> 
> The main idea of the abstractions within testcontainers is that - if possible - just creating a `default()` instance of the image gets you something that is working. In other words, what we strive for is:
> 
> ```rust
> let docker = Cli::default();
> let container = docker.run(Kafka::default());
> ```
> 
> Should give you a running container.
> 
> It might be tricky to achieve that because as far as I understand, kafka always depends on zookeeper but even in that case, there should be minimal amounts of configuration needed to get a running container. Hence, at the most, we should need to specify the zookeeper port and the rest happens by default.
> 
> Let me know if you need more input :)

Thank you @thomaseizinger for your detailed feedback! That makes sense. Let me work on pushing as many configurations as possible to `Kafka::default` so the user does not need to specify them. I'll also make sure to address other points, too.

tests/images/kafka.rs
```diff
@@ -0,0 +1,147 @@
...
+    // Pre-populates a `topic` before a producer sends the first message to the broker.
+    let topic = format!("topic-{}", Uuid::new_v4());
+    create_topic(&docker, &network, &topic);
+
+    let bootstrap_servers = format!(
+        "localhost:{}",
+        kafka_broaker_node.get_host_port(kafka::DEFAULT_LOCALHOST_LISTENER_PORT)
+    );
+
+    let producer: FutureProducer = ClientConfig::new()
```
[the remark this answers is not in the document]
saitoy1: Looks like it wouldn't compile without the hint:
```
error[E0282]: type annotations needed
   --> tests/images/kafka.rs:128:24
    |
103 |     let producer = ClientConfig::new()
    |         -------- consider giving `producer` a type
...
128 |         let producer = producer.clone();
    |                        ^^^^^^^^ cannot infer type
    |
    = note: type must be known at this point
```

tests/images/kafka.rs
```diff
@@ -0,0 +1,147 @@
...
+fn create_topic(docker: &clients::Cli, network: &Uuid, topic: &str) {
+    let topic_creation_command = format!("cub kafka-ready -b broker:{} 1 20 && \
+kafka-topics --create --if-not-exists --bootstrap-server broker:{} --partitions 1 --replication-factor 1 --topic {}",
+	kafka::DEFAULT_BROKER_LISTENER_PORT, kafka::DEFAULT_BROKER_LISTENER_PORT, topic);
+
+    // Once `topic` has been created, it is ok to destroy the container at the end of the function
+    // because it is meant to be a one-shot.
+    docker.run_with_args(
+        kafka::Kafka::default()
+            .with_wait_for(WaitFor::message_on_stdout("Created topic"))
```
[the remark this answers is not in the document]
saitoy1: Addressed in 0ae4d44. Originally, this `docker run` needed a different wait condition because this container was used specifically to populate a Kafka topic, printing different log messages to `stdout` compared to the one seen in a Kafka container when the brokers were getting started. That said, I removed the function that contained these lines of code as populating a topic turned out to be unnecessary for testing.

tests/images/kafka.rs
```diff
@@ -0,0 +1,147 @@
...
+            .with_args(kafka::KafkaArgs(vec![
+                "/bin/bash".to_owned(),
+                "-c".to_owned(),
+                topic_creation_command,
+            ])),
```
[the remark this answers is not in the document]
saitoy1: Addressed in 0ae4d44.

tests/images/kafka.rs
```diff
@@ -0,0 +1,147 @@
...
+    let _zk_node = docker.run_with_args(
+        zookeeper::Zookeeper::default(),
+        RunArgs::default()
+            .with_name("zookeeper")
+            .with_network(network),
+    );
+
+    let mut env_vars = HashMap::new();
+    env_vars.insert(
+        "KAFKA_ZOOKEEPER_CONNECT".to_owned(),
```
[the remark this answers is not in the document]
saitoy1: Addressed in 0ae4d44.

tests/images/kafka.rs
```diff
@@ -0,0 +1,147 @@
...
+        kafka::Kafka::default()
+            .with_env_vars(env_vars)
+            .with_wait_for(WaitFor::message_on_stdout("started")),
+        RunArgs::default()
+            .with_name("broker")
+            .with_network(network)
+            .with_mapped_port((
+                kafka::DEFAULT_LOCALHOST_LISTENER_PORT,
+                kafka::DEFAULT_LOCALHOST_LISTENER_PORT,
+            )),
```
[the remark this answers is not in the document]
saitoy1: So this is something I could not hide from the user completely after my best effort. When the user runs a Kafka container it will look like the following:
```
let kafka_node = docker.run_with_args(
    kafka::Kafka::default(),
    RunArgs::default().with_mapped_port((9093, 9093)),
);
```
The reason for the port mapping is that before the container runs, one of the required environment variables `KAFKA_ADVERTISED_LISTENERS` needs to specify the exposed port for a Kafka port but this exposed port is only known after the container has run. To work around this problem, we need to specify the port mapping so that the exposed port is known before the container runs.

I have noticed that other `testcontainers` client libraries seem to provide a way to obtain the exposed port for a given port (an [example](https://github.com/testcontainers/testcontainers-java/blob/master/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java#L56) in Java, [another](https://github.com/testcontainers/testcontainers-python/blob/bc969c53680606d7fff6a2e8350f5a6f234fcd38/testcontainers/kafka.py#L35) in Python)? If we can do the same in Rust we should be able to remove the port mapping completely.

tests/images/kafka.rs
```diff
@@ -0,0 +1,147 @@
...
+                    Duration::from_secs(0),
+                )
+                .await;
+
+            delivery_status
+        }
+    });
+
+    // Makes sure that messages above get published before consumption.
+    join_all(futures).await;
```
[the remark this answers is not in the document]
saitoy1: Agreed. Addressed in 0ae4d44.

tests/tests.rs
```diff
@@ -0,0 +1,3 @@
+mod images {
+    mod kafka;
+}
```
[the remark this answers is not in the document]
saitoy1: Addressed in 0ae4d44.

tests/images/kafka.rs
```diff
@@ -0,0 +1,147 @@
...
+            .stream()
+            .take(number_of_messages)
+            .try_for_each(|borrowed_message| {
+                // It is safe to unwrap as the producer has included payloads above.
+                let payload = borrowed_message.payload_view::<str>().unwrap().unwrap();
+                actual.push(payload.to_owned());
+                future::ready(Ok(()))
+            });
+
+    message_stream.await.unwrap();
```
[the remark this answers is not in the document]
saitoy1: Thanks for point it out. Addressed in 0ae4d44.

src/images/kafka.rs
saitoy1: Good point. Addressed in 0ae4d44.

saitoy1:
As of `1d1fd46`, I have addressed the code review feedback. After this change, we run `ZooKeeper` that's embedded in a confluent Kafka image as opposed to running it separately outside the Kafka container. By doing so, we are able to remove various configurations parameters we originally specified at container's startup.

I see that Java's `testcontainers` client library for a Kafka image allows the user to specify an [externally running ZooKeeper](https://github.com/testcontainers/testcontainers-java/blob/master/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java#L78-L81) container. For now, I have decided not to include that feature as I am unsure whether it is needed or not. That said, we should be able to support it if it's requested, much the way I implemented it when I first opened the PR.

tests/images/kafka.rs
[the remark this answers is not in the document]
thomaseizinger: I assume `KAFKA_ADVERTISED_LISTENERS` is only used internally within the container right?

The port within the container can always be 9093. This env variable should not matter for how docker maps the port onto the host system.

src/images/kafka.rs
```diff
@@ -0,0 +1,124 @@
...
+    type Item = String;
+    type IntoIter = ::std::vec::IntoIter<String>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        let mut args = vec![self.bash_pathname.into_os_string().into_string().unwrap()];
+        args.extend(self.options.into_iter());
+        args.push(self.commands);
+        args.into_iter()
+    }
+}
```
thomaseizinger, line 50: This looks fairly inconsistent to me:

The fields are public and therefore configurable. However, from the way they are populated within `default`, it doesn't seem like a user would get a working container if they override any this.

As a result, it looks like we don't need any of these fields and can just inline everything into the `IntoIterator` implementation. Generally, the way this is meant to be used is:

- the `XYZArgs` struct has strongly typed fields that - using the type system - model the full range of valid options.
- the `IntoIterator` implementation uses those fields and converts them into whatever command string necessary to run the container

It looks like none of what is stored within `KafkaArgs` makes sense to be configurable so concluding from that, all of this should just be inlined into the `IntoIterator` implementation and `KafkaArgs` becomes a zero-sized struct.

src/images/kafka.rs
```diff
@@ -0,0 +1,124 @@
...
+    arguments: KafkaArgs,
+    env_vars: HashMap<String, String>,
+    tag: String,
+}
+
+impl Kafka {
+    pub fn with_env_vars(mut self, env_vars: HashMap<String, String>) -> Self {
+        self.env_vars.extend(env_vars.into_iter());
+        self
+    }
```
thomaseizinger, line 63: Do we ever need to modify any environment variables? If not, this can go away.

src/images/kafka.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        env_vars.insert(
+            "KAFKA_LISTENER_SECURITY_PROTOCOL_MAP".to_owned(),
+            "BROKER:PLAINTEXT,PLAINTEXT:PLAINTEXT".to_owned(),
+        );
+        env_vars.insert("KAFKA_BROKER_ID".to_owned(), "1".to_owned());
+        env_vars.insert(
+            "KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR".to_owned(),
+            "1".to_owned(),
+        );
```
thomaseizinger, line 85: These look like they are always static so you can use [`phf`](https://docs.rs/phf/0.8.0/phf/) to store them in a private const and then just reference that const in the `fn env_vars()` function.

tests/kafka.rs
```diff
@@ -0,0 +1,80 @@
...
+use std::time::Duration;
+
+use testcontainers::{clients, core::RunArgs, images::kafka};
+
+#[tokio::test]
+async fn test_produce_and_consume_messages() {
+    let docker = clients::Cli::default();
+    let kafka_node = docker.run_with_args(
+        kafka::Kafka::default(),
+        RunArgs::default().with_mapped_port((kafka::KAFKA_PORT, kafka::KAFKA_PORT)),
```
thomaseizinger, line 17: Like I commented on the other one, I believe this is still not necessary because the `KAFKA_ADVERTISED_LISTENERS` env variable is actually scoped within the container and there, the listening port is always the default one.

How docker maps this port to the host port should not matter!

thomaseizinger: This means we can probably also get rid of the exported constant.

tests/kafka.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    let topic = "test-topic";
+    consumer
+        .subscribe(&[&topic])
+        .expect("Failed to subscribe to a topic");
+
+    let number_of_messages_to_produce = 5_usize;
+    let expected: Vec<String> = (0..number_of_messages_to_produce)
+        .map(|i| format!("Message {}", i))
+        .collect();
```
thomaseizinger, line 45: Personally, I usually prefer to use turbo-fish syntax on the actual function that requires the type parameters. The reason is that type-hints easily get stale if you change the expression whereas usage of turbo-fish syntax gives the compiler just enough information to continue with type inference by itself. That in turn makes future code changes easier because type inference automatically adapts to your new code.

```suggestion
    let expected = (0..number_of_messages_to_produce)
        .map(|i| format!("Message {}", i))
        .collect::<Vec<_>();
```

tests/kafka.rs
```diff
@@ -0,0 +1,80 @@
...
+        .collect::<Vec<_>>()
+        .await;
+
+    let number_of_messages_consumed = actual
+        .iter()
+        .zip(expected.iter())
+        .filter(|&(consumed, produced)| consumed.as_ref().unwrap() == produced)
+        .count();
+
+    assert_eq!(number_of_messages_to_produce, number_of_messages_consumed);
```
thomaseizinger, line 79: This assertion is not particularly robust.

Let's say something actually goes wrong and we don't receive as many messages as we publish. A stream with `take(number_of_messages_to_produce).collect()` will wait forever to receive `number_of_messages_to_produce` messages. Hence, we are never actually triggering the final assertion.

However, I guess we can't collect the whole stream because it will never actually end because there could always be more messages.

I think the most robust approach would be to do the assertion something like this (needs proper type conversions and error handling):

```rust
let mut produced_messages = consumer.stream();

for msg in expected {
	let produced = tokio::time::timeout(produced_messages.next(), Duration::from_secs(2)).unwrap();

	assert_eq!(msg, produced)
}
```

src/images/kafka.rs
```diff
@@ -0,0 +1,124 @@
...
+
+impl Default for KafkaArgs {
+    fn default() -> Self {
+        let starter_commands = format!(
+            "echo 'clientPort={}' > zookeeper.properties; \
+echo 'dataDir=/var/lib/zookeeper/data' >> zookeeper.properties; \
+echo 'dataLogDir=/var/lib/zookeeper/log' >> zookeeper.properties; \
+zookeeper-server-start zookeeper.properties & \
+export KAFKA_ZOOKEEPER_CONNECT='localhost:{}'; \
+export KAFKA_ADVERTISED_LISTENERS='PLAINTEXT://localhost:{},BROKER://localhost:9092'; \
```
thomaseizinger, line 25: Do these have to be part of this command string or can we specify them via the `env_args` infrastructure.

src/images/kafka.rs
```diff
@@ -0,0 +1,124 @@
...
+. /etc/confluent/docker/bash-config && \
+/etc/confluent/docker/configure && \
+/etc/confluent/docker/launch",
```
thomaseizinger, line 28: This might be more readable if you use a raw-string literal because you can use actual line-breaks without escaping.

thomaseizinger (changes requested):
Thank you for working in the feedback!

I've left a few more comments. Esp. the design of `KafkaArgs` still doesn't sit particularly well with me.

src/images/kafka.rs
saitoy1: That'd be a great idea, except that there seems to be an impedance mismatch between the return type of `env_vars` and `phf::Map`? Let's say we have the following static map:
```
type KafkaEnvVarsMap = phf::Map<&'static str, &'static str>;

static KAFKA_ENV_VARS: KafkaEnvVarsMap = phf_map! {
    "KAFKA_ZOOKEEPER_CONNECT" => "localhost:2181",
    "KAFKA_ADVERTISED_LISTENERS" => "PLAINTEXT://localhost:9093,BROKER://localhost:9092",
    "KAFKA_LISTENER_SECURITY_PROTOCOL_MAP" => "BROKER:PLAINTEXT,PLAINTEXT:PLAINTEXT",
    "KAFKA_BROKER_ID" => "1",
    "KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR" => "1",
};
```
The body of `env_vars` returning a reference to the map does not look like it can satisfy the trait bound for `Image::EnvVars`: `Default + IntoIterator<Item = (String, String)>`, plus `phf::Map` does not seem to support owned Strings as a key type. Do we want to use `lazy_static` or `once_cell::sync::Lazy` instead if `pfh::Map` does not work?

tests/kafka.rs
saitoy1: I'd like to remove the port mapping as much as you do but cannot think of a way to get rid of it. Let me share this with you to better illustrate the pain point . Say I drop the port mapping so that we run a Kafka container as
```
docker.run(kafka::Kafka::default());
```
We then need to obtain the location of a boot strap server (so we can later create a Producer and a Consumer with that info) and this will look like
```
let bootstrap_servers = format!("localhost:{}", kafka_node.get_host_port(9092));  # only exposes 9092 by default
```
bootstrap_servers would probably look like `localhost:18291` that is, it will be assigned whatever port number bound to the container port 9092.

Here comes the problem. Given the exposed port `18291`, `KAFKA_ADVERTISED_LISTENERS` needs to specify `PLAINTEXT://localhost:18291,BROKER://localhost:9093` (BROKER has 9093 unlike my code actual code, but that does not change what we're talking about here). It's explained [here](https://www.confluent.io/blog/kafka-client-cannot-connect-to-broker-on-aws-on-docker-etc/#scenario-4) in more details, but the idea is we need to specify the host port  bound to 9092 . If, instead, `KAFKA_ADVERTISED_LISTENERS` specified `PLAINTEXT://localhost:9092,BROKER://localhost:9093` then the Producer in the test code would fail to publish messages due to a timeout error. In our case, a Kafka is running within a container and the test code is running on a bare metal. For the test code, including the Producer and the Consumer, to be able to talk to the Kafka broker running in the container, `KAFKA_ADVERTISED_LISTENERS` needs to specify the exposed host port, but setting that port number for `KAFKA_ADVERTISED_LISTENERS` happens before `kafka_node.get_host_port(9092)` so we don't know what the port number to specify for `KAFKA_ADVERTISED_LISTENERS` while `kafka::Kafka::default()` runs.

To work around this problem, we seem to have to have a port mapping. This way, we know a priori what will be the exposed host port for 9092 and can specify that port number in `KAFKA_ADVERTISED_LISTENERS`. I hope I've made myself understood. Given that, do you think it's possible for us to remove the port mapping? Interestingly, Java and Python client libraries can obtain the exposed host port for a container port *while* the container is starting up.

tests/kafka.rs
saitoy1: It's a great suggestion. I think it's important to check timeout issues as part of verification. 81be19a addressed it and I intentionally use `unwrap` throughout. I learned that testing is one of the places where it makes sense to use `unwrap`. Let me know if we want to use alternatives like `expect`.

src/images/kafka.rs
saitoy1: Addressed in `81be19a`.

saitoy1: Addressed in 81be19a.

src/images/kafka.rs
saitoy1: Glad you asked. It turns out they don't, which is a pleasant surprise to me. I removed them from the command and moved them to `env_args` in 81be19a.

tests/kafka.rs
saitoy1: Makes sense. Converted all the type annotations to turbo-fish in 81be19a.

src/images/kafka.rs
saitoy1: Addressed in 81be19a.

src/images/kafka.rs
saitoy1: Thanks for the clarification. Now I see what you mean. Since they are rather mandatory & fixed values, hard-coding them in `IntoIterator` makes perfect sense to me. Addressed in 81be19a.

src/images/kafka.rs
thomaseizinger: Bummer! Sorry to send you onto this goose-chase. I should have checked better before making this suggestion.

Let's just leave it as is then :)

tests/kafka.rs
saitoy1: I've experimented and learned (a bit surprisingly) that the host port and container port need to be the same in order for the client running on a bare metal to talk to the Kafka broker running in a container.

If we started a container without a port mapping:
```
docker.run(kafka::Kafka::default());
```
and if we obtained the address of a boot strap server
```
let bootstrap_servers = format!("localhost:{}", kafka_node.get_host_port(9092));  # only exposes 9092 by default
```
`bootstrap_servers` would look like `localhost:18168` that is, it would be assigned an arbitrary port number bound to the container port 9092. Now, if we set `KAFKA_ADVERTISED_LISTENERS` to `PLAINTEXT://localhost:9092,BROKER://localhost:9093` (BROKER has 9093 unlike my code actual code, but that does not change what we're talking about), the Producer would not be able to talk to the Kafka broker any more (would issue a timeout error) and that's because 18168 is not specified in the port appearing in `KAFKA_ADVERTISED_LISTENERS` ([this](https://stackoverflow.com/questions/51514891/use-docker-compose-with-port-mapping-for-local-kafka-testing) can be related).

In sum, if there is a way to force the exposed host port number to be the same as the container port number, we can remove
```
with_mapped_port((kafka::KAFKA_PORT, kafka::KAFKA_PORT)),
```
FYI, this does not need to be `KAFKA_PORT`, any port number will work equally as long as we pass the same port numbers.

tests/images/kafka.rs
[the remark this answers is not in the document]
saitoy1: Please see https://github.com/testcontainers/testcontainers-rs/pull/282#discussion_r612841110

src/images/kafka.rs
saitoy1: No problem, thanks for the suggestion! While on this topic, I removed `with_env_vars` as per [this](https://github.com/testcontainers/testcontainers-rs/pull/282#discussion_r612096632) conversation but started feeling that the user not being able to configure Kafka settings through environment variables can be a bit too restrictive. Maybe we can leave `with_env_vars` removed for now and keep an eye on users' request in the future?

tests/kafka.rs
saitoy1: Here is my conclusion after looking into this a bit more. With current `testcontainers-rs`, I don't think it's feasible to remove `with_mapped_port` when running a Kafka container. BTW, I'll stand corrected; what I said above regarding a port mapping has to have the same port numbers externally and internally is incorrect. With additional environment variables for Kafka, that restriction can be lifted.

### Problem 1 - Kafka container needs to expose additional port as it is starting up 
When a Kafka client, whether a Producer or a Consumer, running on a bare metal interacts with a Kafka broker running in a container, the Kafka container needs to expose additional port, besides the default 9092, as explained [here](https://www.confluent.io/blog/kafka-client-cannot-connect-to-broker-on-aws-on-docker-etc/#scenario-4). Out of curiosity, I ran an integration test from the Python's `testcontainers` for its Kafka image and saw it exposes port 9093  by default:
```
$ docker ps
CONTAINER ID   IMAGE                         COMMAND                  CREATED         STATUS         PORTS                               NAMES
3e18d59476ff   confluentinc/cp-kafka:5.4.3   "sh -c 'while [ ! -f…"   3 seconds ago   Up 2 seconds   9092/tcp, 0.0.0.0:55079->9093/tcp   hungry_proskuriakova
```
testcontainers for Python and Java can specify an additional port mapping *while the container is starting up* so they can specify the exposed host port for 9093 within `KAFKA_ADVERTISED_LISTENERS` (please see [here](https://github.com/testcontainers/testcontainers-java/blob/dabc99c73eb9832ed2971c42412dd56bbd4d15be/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java#L56) and [here](https://github.com/testcontainers/testcontainers-python/blob/bc969c53680606d7fff6a2e8350f5a6f234fcd38/testcontainers/kafka.py#L46) for examples). With the Rust library, it does not look like we can do that because specifying port mapping [has been moved out](https://github.com/testcontainers/testcontainers-rs/commit/f95a5c22167b6b0093ec97479c6070fae134a498) from `Image` to `Docker`. As such, within `kafka::Kafka::default()`, we cannot write down the exposed port for 9093 within `KAFKA_ADVERTISED_LISTENERS` and we only get to know the exposed port for 9093 after `docker.run`.

### Problem 2 -  `KAFKA_ADVERTISED_LISTENERS` is not just used internally
Since the value specified for `KAFKA_ADVERTISED_LISTENERS` is used by the Kafka client running on a bare-metal even after the container has started up, it is not just internally used by the Kafka broker.

---
In the Rust library, if there were a way to specify the exposed port for 9093 within `Kafka::default` so that we could provide the following for `KAFKA_ADVERTISED_LISTENERS` (it's shown conceptually and obviously the syntax is incorrect),
```
PLAINTEXT://localhost:$(get_exposed_port(9093)),BROKER://localhost:9092
```
then we could remove the port mapping and the user could just run
```
let docker = Cli::default();
let container = docker.run(Kafka::default());
let bootstrap_servers = format!("localhost:{}", container.get_host_port(9093));
```

Kindly let me know what you think.

thomaseizinger: Very interesting, thank you for digging into this topic!

I don't fully understand how the Python and Java implementations achieve this. The port mapping and starting of the container happen at the same time, hence we don't know what the port is being mapped to until after the container is started.

At the point, we have already supplied the startup command to the container so we can no longer update it.

Sounds like a hard problem to solve so I am tempted to defer it for now and merge what we already have.

Let me know another quick pass on the code and see if we can go ahead with this!

thomaseizinger (approved):
This is good to be merged from my side if we **squash all commits** down into one!

As for the topic of kafka's listeners, I also read into this topic a bit. [This link](https://www.confluent.io/blog/kafka-listeners-explained/) was quite helpful and I now understand the problem better!

Looking at [dynamic configurations](https://docs.confluent.io/platform/current/kafka/dynamic-config.html#changing-broker-configurations-dynamically), I think we can solve this problem fairly elegantly.

We will need to adapt the `Image` API to have an (optional) callback that allows to run arbitrary commands within the container after it successfully started up. I'll have a play around with this once this is merged to see if we can - as a result of this - simplify the kafka usecase by using `kafka-configs` to update the `listeners` config property with the exposed port.

The tricky part in making this work is going to be about doing in a portable way between the sync and async `Docker` clients.

Anyway, thanks a lot for sticking to this PR and working in the feedback :)

tests/kafka.rs
saitoy1: Yeah, I think you put it succinctly. I'm good with what we have now as it should hopefully serve as a reasonable starting point.

saitoy1:
> This is good to be merged from my side if we **squash all commits** down into one!
> 
> As for the topic of kafka's listeners, I also read into this topic a bit. [This link](https://www.confluent.io/blog/kafka-listeners-explained/) was quite helpful and I now understand the problem better!
> 
> Looking at [dynamic configurations](https://docs.confluent.io/platform/current/kafka/dynamic-config.html#changing-broker-configurations-dynamically), I think we can solve this problem fairly elegantly.
> 
> We will need to adapt the `Image` API to have an (optional) callback that allows to run arbitrary commands within the container after it successfully started up. I'll have a play around with this once this is merged to see if we can - as a result of this - simplify the kafka usecase by using `kafka-configs` to update the `listeners` config property with the exposed port.
> 
> The tricky part in making this work is going to be about doing in a portable way between the sync and async `Docker` clients.
> 
> Anyway, thanks a lot for sticking to this PR and working in the feedback :)

Agreed, that link was helpful for me as well. I didn't know about the dynamic configurations but am curious to know how it can be used to address this. I'd love to review the code change when it's available.

Appreciate your feedback all along! It certainly helped make the PR a lot better. I think we can go ahead and merge what we have now.

thomaseizinger:
> I think we can go ahead and merge what we have now.

Sure! Do you mind squashing everything into one commit please?

saitoy1:
> > I think we can go ahead and merge what we have now.
> 
> Sure! Do you mind squashing everything into one commit please?

No problem. 48c2372 should contain all squashed commits. Please let me know if I didn't do it right.

thomaseizinger (approved):
> No problem. 48c2372 should contain all squashed commits. Please let me know if I didn't do it right.

Thank you!

For future reference, I consider [this](https://chris.beams.io/posts/git-commit/) to be an excellent guide for good commit messages :)

Thank you again for sticking to this PR and the contribution in general. I will record a ticket for optimizing the port mapping that we talked about!

bors r+

thomaseizinger:
bors retry

PR #288, opened 2021-05-13
Pull image if the image doesn't exist

### LINKED ISSUE #272: Shiplift client doesn't automatically pull image

tony84727:
Fix #272 
Pull the image when creating container returns 404 not found.

### DIFF
--- a/src/clients/http.rs
+++ b/src/clients/http.rs
@@ -4,9 +4,12 @@ use crate::{
 };
 use async_trait::async_trait;
 use futures::{executor::block_on, stream::StreamExt, TryStreamExt};
+use hyper::StatusCode;
+use shiplift::builder::ContainerOptionsBuilder;
+use shiplift::rep::ContainerCreateInfo;
 use shiplift::{
     rep::ContainerDetails, ContainerOptions, Docker, LogsOptions, NetworkCreateOptions,
-    NetworkListOptions, RmContainerOptions,
+    NetworkListOptions, PullOptions, RmContainerOptions,
 };
 use std::{
     fmt, io,
@@ -110,13 +113,32 @@ impl Http {
         }
 
         // create the container with options
-        let create_result = self
-            .inner
-            .shiplift
-            .containers()
-            .create(&options_builder.build())
-            .await;
-        let id = create_result.unwrap().id;
+        let create_result = self.create_container(&options_builder).await;
+        let id = {
+            match create_result {
+                Ok(container) => container.id,
+                Err(err) => match err {
+                    shiplift::Error::Fault { code, message: _ } => {
+                        if code != StatusCode::NOT_FOUND {
+                            panic!("{}", err);
+                        }
+                        {
+                            let mut options_builder = PullOptions::builder();
+                            options_builder.image(image.descriptor());
+                            let mut pulling =
+                                self.inner.shiplift.images().pull(&options_builder.build());
+                            while let Some(result) = pulling.next().await {
+                                if result.is_err() {
+                                    result.unwrap();
+                                }
+                            }
+                        }
+                        self.create_container(&options_builder).await.unwrap().id
+                    }
+                    _ => panic!("{}", err),
+                },
+            }
+        };
 
         self.inner
             .shiplift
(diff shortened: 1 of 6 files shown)


src/clients/http.rs
```diff
@@ -110,13 +113,32 @@ impl Http {
...
-            .create(&options_builder.build())
-            .await;
-        let id = create_result.unwrap().id;
+        let create_result = self.create_container(&options_builder).await;
+        let id = {
+            match create_result {
+                Ok(container) => container.id,
+                Err(err) => match err {
+                    shiplift::Error::Fault { code, message: _ } => {
+                        if code != StatusCode::NOT_FOUND {
```
thomaseizinger, line 122: Can we merge this pattern matching into one and thereby reduce indentation for this logic?

I am thinking of something like:

```rust
match create_result {
	Ok(c) => c.id,
	Err(shiplift::Error::Fault { code, .. }) if code == 404 => {
		// do image pull logic
	}
	e => panic!("Failed to create container: {}", e)
}
```

thomaseizinger (changes requested):
Thank you for contributing!

I left two small comments inline.

Additionally, it would be really great if we could somehow cover this in our integration tests. I am wondering how https://github.com/testcontainers/testcontainers-rs/blob/dev/tests/http_client.rs even works at the moment. Is `hello-world` simply present on the GitHub actions dev machines?

src/clients/http.rs
tony84727: Thank you! I don't even know we can write it like that. Fixed.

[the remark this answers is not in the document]
tony84727: Fixed

tony84727:
> I am wondering how https://github.com/testcontainers/testcontainers-rs/blob/dev/tests/http_client.rs even works at the moment. Is hello-world simply present on the GitHub actions dev machines?

I guess it's because of another integration test:
https://github.com/testcontainers/testcontainers-rs/blob/e36491f7e6fc84e0de15be148f87f1ded06e0f41/tests/podman.rs#L4-L7

docker cli will pull the image if the engine returns an image not found error.

> Additionally, it would be really great if we could somehow cover this in our integration tests.

Yes, but I'm not sure how to write the test for this case, as the test will need to be run with a clean docker engine.

Maybe we write an integration test that will delete the image at the beginning?

Another option might be using [docker in docker, dind](https://hub.docker.com/_/docker). Not sure if we can run dind by Github action though.

thomaseizinger:
> I guess it's because of another integration test:

Right, that makes sense!

> Yes, but I'm not sure how to write the test for this case, as the test will need to be run with a clean docker engine.
> Maybe we write an integration test that will delete the image at the beginning?

If we do that, we need make sure things run in sequence properly, otherwise we might be subject to race conditions. cargo will run all integration test files in sequence, so we can use that to have a dedicated test that uses shiplift's API to first delete the `hello-world` docker image and then use testcontainers to start it up.

tony84727:
> cargo will run all integration test files in sequence, so we can use that to have a dedicated test that uses shiplift's API to first delete the `hello-world` docker image and then use testcontainers to start it up.

Okay. added an integration that will delete `hello-world` image at the beginning and then test if we can spin up a `hello-world` container.

thomaseizinger:
> Okay. added an integration that will delete `hello-world` image at the beginning and then test if we can spin up a `hello-world` container.

Amazing, thank you!

A final request: can we squash all 4 commits into a single one please?

tony84727:
> A final request: can we squash all 4 commits into a single one please?

Sure, feel free to squash those commits. Should I squash commits then force push? Or you will use Github's squash merge?

thomaseizinger:
> Sure, feel free to squash those commits. Should I squash commits then force push? Or you will use Github's squash merge?

Wasn't planning on using squash merge but can do if you are okay with it. You will lose the signatures on your commits but I guess you are fine with that given that you are suggesting the option? :)

Just noticed that CI is not passing. Sorry for the late notice, this new feature of having to activate CI checks every time is annoying ...

tony84727:
> Wasn't planning on using squash merge but can do if you are okay with it. You will lose the signatures on your commits but I guess you are fine with that given that you are suggesting the option? :)

I think it's okay

> Just noticed that CI is not passing. Sorry for the late notice, this new feature of having to activate CI checks every time is annoying ...

CI seems to be failing on the `dev` branch. I've opened
https://github.com/testcontainers/testcontainers-rs/pull/289 to fix it.

Will deal with the formatting after the CI is fixed.

tony84727:
> CI seems to be failing on the dev branch. I've opened
#289 to fix it.
>
> Will deal with the formatting after the CI is fixed.

Fixed CI and formatting issue.

thomaseizinger:
Thank you!

PR #293, opened 2021-08-21
Support expose ports and add Google Cloud SDK image

yottabytt:
Hey @thomaseizinger 

Certain images (eg. [google/cloud-sdk](https://hub.docker.com/layers/google/cloud-sdk/353.0.0/images/sha256-fe242b7bb1334f7ce5fa56872347dd9819f3717040e65dec9478a566b2033f0e?context=explore)) don't have an `EXPOSE` layer. I think we can make use of `--expose` option in `docker run` command for those cases. I have added a method `expose_port` in `Image` trait. It has a default implementation. So, we can implement it only for the images where it is really required. 

At the moment, I only need bigtable emulator for my usecase. But when I looked at `testcontainers-java`, I thought I can add the rest too.

What do you think of this approach? 

I only focussed on `Docker` and `Cli`. Based on our discussion, I will iterate (`expose_port` to `expose_ports`, tests, changes in other files, etc.) this PR.

### DIFF
--- a/src/core/image.rs
+++ b/src/core/image.rs
@@ -96,6 +96,14 @@ where
     fn entrypoint(&self) -> Option<String> {
         None
     }
+
+    /// Returns the ports that needs to be exposed when a container is created.
+    ///
+    /// This method is useful when there is a need to expose some ports, but there is
+    /// no EXPOSE instruction in the Dockerfile of an image.
+    fn expose_ports(&self) -> Vec<u16> {
+        Default::default()
+    }
 }
 
 /// Represents a port mapping between a local port and the internal port of a container.
--- a/src/images.rs
+++ b/src/images.rs
@@ -2,6 +2,7 @@ pub mod coblox_bitcoincore;
 pub mod dynamodb_local;
 pub mod elasticmq;
 pub mod generic;
+pub mod google_cloud_sdk;
 pub mod hello_world;
 pub mod kafka;
 pub mod mongo;
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -160,7 +160,10 @@ impl Client {
                     .arg(format!("{}:{}", port.local, port.internal));
             }
         } else {
-            command.arg("-P"); // expose all ports
+            for port in image.expose_ports() {
+                command.arg(format!("--expose={}", port));
+            }
+            command.arg("-P"); // publish all exposed ports
         }
 
         command
--- a/src/images/google_cloud_sdk.rs
+++ b/src/images/google_cloud_sdk.rs
@@ -1,0 +1,182 @@
+use crate::{core::WaitFor, Image};
+use std::collections::HashMap;
+
+const CONTAINER_IDENTIFIER: &str = "google/cloud-sdk";
+const DEFAULT_TAG: &str = "353.0.0";
+
+const HOST: &str = "0.0.0.0";
+const DEFAULT_PORT: u16 = 80;
+pub const BIGTABLE_PORT: u16 = 8086;
+pub const DATASTORE_PORT: u16 = 8081;
+pub const FIRESTORE_PORT: u16 = 8080;
+pub const PUBSUB_PORT: u16 = 8085;
+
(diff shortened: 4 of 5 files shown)


thomaseizinger (commented):
Yes, happy to include expose but I would like it to be possible to expose multiple ports, not just one.

thomaseizinger:
Can we have the expose functionality in a separate commit please?

Also, I am trying to keep in mind https://github.com/testcontainers/testcontainers-rs/issues/263 with these kind of additions. It looks like this one is specific to each image though and hence could not actually go onto `RunArgs`.

src/core/image.rs
```diff
@@ -96,6 +96,10 @@ where
     fn entrypoint(&self) -> Option<String> {
         None
     }
+
+    fn expose_ports(&self) -> Option<Vec<u16>> {
+        None
+    }
```
thomaseizinger, line 102: Are we gaining anything from wrapping the `Vec` in an `Option`?

src/images/gcloud_emulators/bigtable.rs
```diff
@@ -0,0 +1,100 @@
+use crate::{core::WaitFor, Image};
+use std::collections::HashMap;
+
+const CONTAINER_IDENTIFIER: &str = "google/cloud-sdk";
```
thomaseizinger, line 4: Normally, we have one implementation of `Image` per docker image, not per application that is started from within the container.

I need to think about this but my initial tendency would be to only have one implementation of image and maybe `Args` be an enum with all the different things that one can start in that image?

src/images.rs
```diff
@@ -1,6 +1,7 @@
 pub mod coblox_bitcoincore;
 pub mod dynamodb_local;
 pub mod elasticmq;
+pub mod gcloud_emulators;
```
thomaseizinger, line 4: The naming should be after the docker image, in this case: `google_cloud_sdk`.

thomaseizinger (changes requested):
Thanks for working in the feedback.

As mentioned in the contribution guidelines, I would appreciate if commit messages would follow [this](https://chris.beams.io/posts/git-commit/) style :)

src/core/image.rs
yottabytt: Gotcha. I think we can go with zero-sized `Vec` for default cases.

src/images/gcloud_emulators/bigtable.rs
yottabytt: > I need to think about this but my initial tendency would be to only have one implementation of image and maybe Args be an enum with all the different things that one can start in that image?

Hmm, how we can then go about `ready_conditions()`, possibility of running an application with non-default `RunArgs` and so on?

src/core/image.rs
thomaseizinger: I think that should work yes.

src/images/gcloud_emulators/bigtable.rs
thomaseizinger: You could have different constructor functions on the struct that implements image and initialize `ready_conditions` as a field depending on what the user wanted to run.

yottabytt:
@thomaseizinger I would like to get this completed and merged first since its been open for a while (and #295 would take some more time). Regarding tests, since there is neither an official rust sdk nor well adapted client libs for all those services, I wasn't convinced to go with some random lib even though it may work. However, I felt like having some sort of tests that can depict the basic usage of the image. Hence, added tests around expose ports functionality. Thoughts?

thomaseizinger (approved):
Yep, lets get this in!

Thanks for the work and structuring the commits nicely :)

thomaseizinger:
The `Http` client is current lacking support for multiple things (see the open bugs), so I don't consider it ready yet anyway.

It will need support for the new expose feature at some point!

thomaseizinger:
Don't worry about the build failure, that will be fixed by #296 which is going to land before this.

thomaseizinger:
Feel free to merge this once the lint and kafka fix is in. Simply comment with `bors r+` .

bors d+

Issue #285, opened 2021-04-27
Problem: Kafka image requires manual port mapping


thomaseizinger:
https://github.com/testcontainers/testcontainers-rs/pull/282 introduced a `Kafka` image that _almost_ works out of the box.

As described in discussion there, we unfortunately need to provide a manual port mapping because we need to know the _external_ listening port of the kafka container before startup. [This](https://www.confluent.io/blog/kafka-listeners-explained/) guide explains the problem in more detail.

Ideally, we would want the kafka image to work without a manual port mapping for two reasons:

- It allows us to start multiple kafka containers without having to worry about the port
- It is easy to forget that this image needs a manual port mapping to function properly

There are likely several ways of how to solve this. However, the constraints around any possible solution are the following:

- The solution should be defined within the `Image` trait.
This is necessary because the `Image` trait is the canonical definition of the container we are going to start.
- The solution has to work with different `Docker` **and** `DockerAsync` implementations.
Due to the different signatures of the `Docker` and `DockerAsync` traits, we don't want to introduce this trait as a dependency to the `Image` trait. In fact, we [refactored](https://github.com/testcontainers/testcontainers-rs/pull/253) "waiting for a container" to a declarative solution to even be able to introduce the `DockerAsync` trait in the first place.

bsideup:
FYI it is possible to use random ports by postponing the process until container is started. Have a look at testcontainers-java's implementation:
https://github.com/testcontainers/testcontainers-java/blob/1.15.3/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java

thomaseizinger:
Thanks!

> FYI it is possible to use random ports by postponing the process until container is started.

I can't decide whether or not this is very clever or very hacky :joy: 

Jokes aside, definitely looks like a valid solution although I am more inclined to re-configure Kafka at runtime instead.

thomaseizinger:
The solution I described above would actually allow for both. What command an image wants to execute in a container would be completely opaque to the docker client. Hence, the command could also just start the actual process with the right container right away.

yottabytt:
@thomaseizinger until this gets solved, can't we use #282 ? Coz, i don't see `pub mod kafka` in `images.rs` from the latest release. I also see it was merged but the build was failing.

thomaseizinger:
[...]

PR #294, opened 2021-08-28
Solve kafka image's requirement of manual port mapping

yottabytt:
Hey @thomaseizinger 

Here is my initial take to the issue https://github.com/testcontainers/testcontainers-rs/issues/285 . I would like to get your feedback on this approach.

Note: I am building this on top of https://github.com/testcontainers/testcontainers-rs/pull/293 . 

TODO: 
- [ ] changes to other files esp. `container_async`


src/core/container.rs
```diff
@@ -86,10 +90,9 @@ where
         &self.image
     }
 
-    fn block_until_ready(&self) {
-        log::debug!("Waiting for container {} to be ready", self.id);
-
-        for condition in self.image.ready_conditions() {
+    pub fn block_until_ready(&self, ready_conditions: Vec<WaitFor>) {
```
thomaseizinger, line 93: I want as many of these functions as possible to remain private, otherwise any change to them is a breaking change.

src/core/container.rs
```diff
@@ -62,15 +63,18 @@ where
...
             id,
             docker_client: Box::new(docker_client),
             image,
             command,
+            ports,
             client_lifetime: PhantomData,
         };
 
-        container.block_until_ready();
+        container.block_until_ready(container.image().ready_conditions());
```
thomaseizinger, line 77: This is a weird indirection now. I think we may have to remove `container#block_until_ready` and instead put this functionality onto the `Docker` trait.

src/clients/cli.rs
```diff
@@ -27,6 +28,57 @@ impl Cli {
         self.run_with_args(image, RunArgs::default())
     }
 
+    pub fn run_and_exec<I: ImageExt>(&self, image: I) -> Container<'_, I> {
```
thomaseizinger, line 31: Why do we need a separate function for this? I think this should be default behaviour. We should keep the API minimal.

thomaseizinger (commented):
Thank you for working on this!

I've left some initial comments :)

src/core/image.rs
```diff
@@ -102,6 +103,32 @@ where
     }
 }
 
+pub trait ImageExt: Image
```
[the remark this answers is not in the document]
yottabytt: I started for a different reason i.e. to have an associated type. But, also was thinking I should get rid of this and make it contained within `Image` trait, which could also help us avoid `run_and_exec()`.

src/core/image.rs
```diff
@@ -102,6 +103,32 @@ where
...
+where
+    Self::ExecCommands: Default + IntoIterator<Item = ExecCommand>,
+{
+    type ExecCommands;
+
+    fn exec_after_start(&self) -> Self::ExecCommands;
```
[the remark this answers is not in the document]
yottabytt: I think you meant `ContainerDetails` type. I am not seeing anything inherently in terms of making things simpler by making use of that type. Plz correct me if I am missing anything.

yottabytt:
@thomaseizinger 👋  I think I have addressed almost all of your initial comments. Also, have added sufficient information in commit descriptions.

src/core/image.rs
```diff
@@ -81,6 +86,34 @@ where
     fn expose_ports(&self) -> Vec<u16> {
         Default::default()
     }
+
+    /// Returns the commands that needs to be executed after a container is started i.e. commands
+    /// to be run in a running container.
+    ///
+    /// This method is useful when certain re-configuration is required after the start
+    /// of container for the container to be considered ready for use in tests.
+    fn exec_after_start(&self) -> Option<Vec<ExecCommand>> {
```
thomaseizinger, line 95: What do you think of this function taking an instance of the `ContainerDetails` type?

Then an implementation could search for whatever data it needs and we would not have to maintain the `Substitution` abstraction and extend it to allow for other usecases than figuring out the host port for example.

I guess it is a trade-off between how many details we want to expose.

Overall, I would prefer if we pass data _into_ this function and let the function figure out how to construct the desired command than passing a template out.

We can either:

- use `ContainerDetails` and accept that this type is the part of our public API
- create our own type and keep things limited for now (to for example ports)

I don't mind much either way with a slight preference for the latter.

Adding our own type allows us to extend it with convenience functions like extracing a port mapping.

thomaseizinger: Do we need the `Option<Vec>`? What is the semantic difference between `None` and `Some(empty-vec)`?

src/images/kafka.rs
```diff
@@ -93,4 +96,31 @@ impl Image for Kafka {
...
+        vec![KAFKA_PORT]
+    }
+
+    fn exec_after_start(&self) -> Option<Vec<ExecCommand>> {
+        let mut commands = vec![];
+        let cmd = concat!(
+        "kafka-configs --alter ",
+        "--bootstrap-server 0.0.0.0:9092 --entity-type brokers --entity-name 1 ",
+        "--add-config advertised.listeners=[PLAINTEXT://localhost:{HOST_PORT},BROKER://localhost:9092]",
+        ).to_owned();
```
thomaseizinger, line 110: Why `concat!`? Would a raw string literal also do the job?

src/core/container.rs
```diff
@@ -162,6 +134,32 @@ where
...
+            let cmd = substitutions
+                .into_iter()
+                .fold(cmd, |cmd, (pattern, substitution)| match substitution {
+                    Substitution::HostPort { internal_port } => {
+                        cmd.replace(&pattern, &self.get_host_port(internal_port).to_string())
+                    }
+                    Substitution::Nothing => cmd,
+                });
+
+            log::debug!("Executing command {:?} in container {}", cmd, self.id());
```
thomaseizinger, line 154: Is the debug print of the command going to be pretty?

src/core/container.rs
```diff
@@ -162,6 +134,32 @@ where
         .unwrap_or_else(|_| panic!("container {} has missing or invalid bridge IP", self.id))
     }
 
+    pub fn exec(&self, exec_commands: Vec<ExecCommand>) {
```
thomaseizinger, line 137: I think this public API should just take a single `ExecCommand`.

For the `Image` it makes sense to return a list but any caller can run this in a loop themselves.

thomaseizinger (changes requested):
Thank you!

Some comments :)

yottabytt:
hey @thomaseizinger ! i think have addressed your comments !

PR #295, opened 2021-09-01
Remodel Image trait

yottabytt:
Attempt to resolve issue #263 

Cherry-picked and inspired from @thomaseizinger's excellent and big #269 PR.

I really love the fact that this enables avoiding tons of boiler plate code. 

TODO

- [x] CHANGELOG.md


src/clients/cli.rs
```diff
@@ -395,8 +395,6 @@ mod tests {
     }
 
     impl Image for HelloWorld {
-        type EntryPoint = std::convert::Infallible;
```
thomaseizinger, line 400: I definitely want this patch in as-is. Maybe move it to the front of the PR so rebasing is easier :)

thomaseizinger (commented):
Thank you for pushing this forward!

I have some concerns in regards to type-safety when it comes to the image args. Let's discuss on the inline comment!

src/core/image.rs
```diff
@@ -75,18 +62,24 @@ where
...
+    ///
+    /// 1. Similar to the Default implementation of an Image, the Default instance
+    /// of its arguments should be meaningful!
+    /// 2. Implementations should be conservative about which arguments they expose. Many times,
+    /// users will either go with the default arguments or just override one or two. When defining
+    /// the arguments of your image, consider that the whole purpose is to facilitate integration
+    /// testing. Only expose those that actually make sense for this case.
+    fn args(&self) -> Box<dyn Iterator<Item = String>> {
+        Box::new(vec![].into_iter())
+    }
```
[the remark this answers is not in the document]
yottabytt: > I think as a general theme, what we need to achieve is removing state from implementations of Image and make them as declarative as possible. That will cut down on the "getters" that users need to override. If we instead leverage composition on the call-site (i.e. when calling docker.run) 

This kinda makes sense. So, then, shouldn't we move everything else (`env_vars()` and may be `volumes()`?) apart from `descriptor()` and `ready_conditions()`out of `Image`? Also, I am curious should the method `exec_after_start()`  (a.k.a the method that cares of commands to run after starting the container) also be out of `Image`?

src/core/image.rs
```diff
@@ -26,70 +37,32 @@ where
...
     /// 1. Similar to the Default implementation of an Image, the Default instance
     /// of its environment variables should be meaningful!
     /// 2. Implementations should be conservative about which environment variables they expose. Many times,
     /// users will either go with the default ones or just override one or two. When defining
     /// the environment variables of your image, consider that the whole purpose is to facilitate integration
     /// testing. Only expose those that actually make sense for this case.
-    type EnvVars;
+    fn env_vars(&self) -> Box<dyn Iterator<Item = (&String, &String)> + '_> {
+        Box::new(std::iter::empty())
+    }
```
yottabytt, line 54: Regardless of where this method and `volumes()` ends up, shall we also then avoid iterators as their return types? Simply just `BTreeMap` or `HashMap`?

src/core/image.rs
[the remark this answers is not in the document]
thomaseizinger: Possibly. We might be onto something here with that viewpoint.
It depends on what the user may want to modify vs what is always static with the image.

As you've pointed out, `descriptor`, `ready_conditions` and `expose_ports` are _static_ and always bound to the image and its version.

But I think so are `volumes` and `env_vars`. The only thing that is different is that independently of the volumes and env vars needed by and image, users should have the ability to specify additional env vars and volumes when starting any image.

In a way, it is similar to docker's `Dockerfile` vs commandline arguments:

- A `Dockerfile` can declare a volume
- The `docker` CLI exposes `-v` to mount additional volumes

It sounds like `Image` needs to be able to describe everything that a `Dockerfile` can describe. Everything else needs to be configurable at the time when an actual image gets `run`.

`ready_conditions` and `exec_after_start` are special cases that don't exist within docker (although I wish docker had a way of expressing `ready_conditions` within a `Dockerfile`). I do see them as belonging to the `Image`. It is something we want to take care of for users of a particular `Image` implementation.

In an ideal world, the API we provide makes it impossible to misuse a particular image, i.e. you should always end up with a running / working container.

src/core/image.rs
thomaseizinger: Not too sure about exposing maps directly. The most UX friendly might be to create our own type, plus a macro or other convenient APIs to declare them effortlessly. Initially, I settled on iterators to be able to make more changes without them being breaking.

thomaseizinger: If it makes things easier, we can roll with exposing the maps directly though!

src/core/image.rs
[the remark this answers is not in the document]
yottabytt: > It depends on what the user may want to modify vs what is always static with the image.

> It sounds like Image needs to be able to describe everything that a Dockerfile can describe. Everything else needs to be configurable at the time when an actual image gets run.

Makes total sense.

> ready_conditions and exec_after_start are special cases that don't exist within docker (although I wish docker had a way of expressing ready_conditions within a Dockerfile). I do see them as belonging to the Image. It is something we want to take care of for users of a particular Image implementation. 

I was on the same page for `ready_conditions`. But for `exec_after_start`, I felt like we can also allow the users to run whatever command they want e.g. some highly specific `kafka-configs` command. But thinking now, I guess we shouldn't.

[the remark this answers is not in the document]
thomaseizinger: For arbitrary commands, we can expose `exec` on `Container` :)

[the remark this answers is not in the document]
yottabytt: Oh yeah, i was intending to mean `exec()` 😆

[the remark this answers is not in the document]
thomaseizinger: Yeah, I guess we can actually ship that as two separate features:

1. Allow users to run stuff from a container instance: `Container::exec`
2. Use `Container::exec` to implement `exec_after_start` and fix the kafka port mapping problem through that.

yottabytt:
@thomaseizinger just wanted to keep you posted that I am back for working on these PRs ❣️

yottabytt:
@thomaseizinger In general, pls let me know WDYT of this approach. I'll add a few more inline comments to explain and to enable discussion.

src/core/image.rs
```diff
@@ -106,6 +83,114 @@ where
     }
 }
 
+#[derive(Debug)]
+pub struct RunnableImage<I: Image> {
+    image: I,
+    image_args: I::Args,
+    image_tag: Option<String>,
```
yottabytt, line 90: I am kinda neutral in whether `image_tag` should be here or not. WDYT?

src/images/google_cloud_sdk.rs
```diff
@@ -11,60 +10,65 @@ pub const DATASTORE_PORT: u16 = 8081;
...
 }
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Copy)]
 pub enum Emulator {
     Bigtable,
     Datastore,
     Firestore,
     PubSub,
-    Help,
```
yottabytt, line 30: I did not like having this. By not having this, the image doesn't adhere to the req. that image args should have a meaningful default. And I think that's okay for images having many distinct services instead of just one. Just for the sake of some meaningful default, having such a not-so-meaningful (can also be considered as meaningless) did make much sense to me. WDYT?

src/clients/cli.rs
```diff
@@ -23,26 +24,22 @@ pub struct Cli {
...
 
 impl Cli {
-    pub fn run<I: Image>(&self, image: I) -> Container<'_, I> {
-        self.run_with_args(image, RunArgs::default())
-    }
-
-    pub fn run_with_args<I: Image>(&self, image: I, run_args: RunArgs) -> Container<'_, I> {
-        let mut docker = self.inner.command();
+    pub fn run<I: Image>(&self, image: impl Into<RunnableImage<I>>) -> Container<'_, I> {
+        let runnable = image.into();
```
thomaseizinger, line 28: Nit: If we would shadow the `image` variable here, the diff would be a lot less noisy. Also, I think `image` is still the primary concept of `RunnableImage` contrary to `runnable`.

src/clients/cli.rs
```diff
@@ -126,51 +123,51 @@ impl Client {
...
             command.arg("-P"); // publish all exposed ports
         }
 
         command
             .arg("-d") // Always run detached
-            .arg(image.descriptor())
-            .args(image.args())
-            .stdout(Stdio::piped())
+            .arg(runnable.image().descriptor())
+            .args(runnable.image_args().clone().into_iter())
```
thomaseizinger, line 167: Can we expose forwarding functions on `RunnableImage` instead of reaching into `image()` itself?

We will have to do this for all clients (currently only two), so making this reasonably ergonomic will bring a benefit.

src/clients/cli.rs
```diff
@@ -407,79 +405,64 @@ mod tests {
...
-        assert!(format!("{:?}", command).contains(r#"-d"#));
-        assert!(format!("{:?}", command).contains(r#"-P"#));
-        assert!(format!("{:?}", command).contains(r#""-v" "one-from:one-dest"#));
-        assert!(format!("{:?}", command).contains(r#""-v" "two-from:two-dest"#));
-        assert!(format!("{:?}", command).contains(r#""-e" "one-key=one-value""#));
-        assert!(format!("{:?}", command).contains(r#""-e" "two-key=two-value""#));
+        assert_eq!(
+            format!("{:?}", command),
+            r#""docker" "run" "-e" "one-key=one-value" "-e" "two-key=two-value" "-v" "one-from:one-dest" "-v" "two-from:two-dest" "-P" "-d" "hello-world""#
+        );
```
thomaseizinger, line 437: Having this refactoring of switching to a single assert in the tests in a separate patch would be amazing but I am also okay if you don't want to bother :)

tests/images.rs
```diff
@@ -17,10 +17,13 @@ use testcontainers::{core::WaitFor, *};
...
 
 #[test]
+#[ignore = "auth() needs to be worked on"]
 fn coblox_bitcoincore_getnewaddress() {
     let _ = pretty_env_logger::try_init();
     let docker = clients::Cli::default();
-    let node = docker.run(images::coblox_bitcoincore::BitcoinCore::default());
+    let node = docker.run(RunnableImage::from(
+        images::coblox_bitcoincore::BitcoinCore::default(),
+    ));
```
thomaseizinger, line 26: Can we define the `run` API in such a way that we don't need to type `RunnableImage::from`? Does this compile?

```rust
impl<I: Image> Into<RunnableImage<I>> for I {

}
```

I think that should allow users to pass instances of `Image` into `Docker::run`.

src/images/google_cloud_sdk.rs
thomaseizinger: Image args should have a meaningful default if possible. But we should not implement `Default` if we cannot provide a meaningful one though.

In this case, I think we should remove the `Default` impl for `CloudSdkArgs`.

src/images/google_cloud_sdk.rs
```diff
@@ -162,6 +125,51 @@ impl CloudSdk {
         }
     }
 
+    /// RunnableImage::from() uses default Image::Args, which is not
+    /// applicable for CloudSdk image. Hence, this helper method can be
+    /// used to set the correct args.
+    pub fn as_runnable(self) -> RunnableImage<Self> {
+        let args = self.arguments.clone();
+        RunnableImage::from(self).with_image_args(args)
+    }
```
thomaseizinger, line 134: 1. If we can get the `impl Into<RunnableImage<I>> for I` working, I believe we don't need this.
2. If we need this, we should adhere to naming conventions for conversion functions: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

src/core/image.rs
```diff
@@ -89,6 +83,114 @@ where
...
+        Self {
+            ports: Some(ports),
+            ..self
+        }
+    }
+}
+
+impl<I> From<I> for RunnableImage<I>
+where
+    I: Image + Default,
```
thomaseizinger, line 177: This bound is unnecessary I believe.

src/core/image.rs
```diff
@@ -89,6 +83,114 @@ where
...
+    }
+}
+
+impl<I: Image> RunnableImage<I> {
+    pub fn with_image_args(self, args: <I as Image>::Args) -> Self {
+        Self {
+            image_args: args,
+            ..self
+        }
+    }
```
thomaseizinger, line 134: I think we should change this into a constructor that also allows to pass in the `Image`:

```rust
    pub fn with_args(image: I, args: <I as Image>::Args) -> Self {
        Self {
			image: I,
            image_args: args,
            ..self
        }
    }
```

That is useful for cases where neither `Image`, nor `Image::Args` can implement `Default`.

src/core/image.rs
thomaseizinger: What we can add is:

```rust
impl Default for RunnableImage<I> where I: Image + Default, I::Args: Default { }
```

src/core/image.rs
```diff
@@ -89,6 +83,114 @@ where
...
+
+    pub fn volumes(&self) -> &BTreeMap<String, String> {
+        &self.volumes
+    }
+
+    pub fn ports(&self) -> &Option<Vec<Port>> {
+        &self.ports
+    }
+}
```
thomaseizinger, line 126: Do we need to expose all of these publicly? I would like the public API to be as small as possible so we have to release less breaking changes.

src/clients/http.rs
```diff
@@ -372,11 +358,10 @@ mod tests {
 
     #[tokio::test(flavor = "multi_thread")]
     async fn http_run_command_should_include_name() {
-        let image = GenericImage::new("hello-world");
         let docker = Http::new();
-
-        let run_args = RunArgs::default().with_name("hello_container");
-        let container = docker.run_with_args(image, run_args).await;
+        let image = GenericImage::new("hello-world");
+        let runnable = RunnableImage::from(image).with_name("hello_container");
```
thomaseizinger, line 363: What do you think of an extension trait to `Image` that allows one to call `with_name` and all of the other functions directly on `Image`?

It would enable an API such as:

```rust
let container = docker.run(GenericImage::new("hello-world").with_name("hello_container")).await;
```

src/core/container.rs
```diff
@@ -84,13 +84,13 @@ where
     /// [`Image`]: trait.Image.html
     /// [`arguments`]: trait.Image.html#associatedtype.Args
     pub fn image(&self) -> &I {
-        &self.image
+        self.runnable.image()
     }
```
thomaseizinger, line 88: By separating the `Args` from `Image`, I think we will need to expose `Image::Args` from a container as well.

That should also fix the ignored test with Bitcoin-Core's auth.

src/clients/http.rs
thomaseizinger: We can ship this as a follow-up feature though, not a requirement for me to merge this.

src/core/image.rs
thomaseizinger: Am I missing something or is it currently unused? Overriding the tag should be something that we can provide generically so yes, I think this should be in here.

Can you add docs to the `with_tag` function that explains that once that is used, there are no more guarantees that the container will actually work?

src/images/google_cloud_sdk.rs
```diff
@@ -11,60 +10,65 @@ pub const DATASTORE_PORT: u16 = 8081;
...
+                    Emulator::Firestore => "firestore",
+                    Emulator::PubSub => "pubsub",
+                }
+                .to_owned(),
+            );
+
+            args.push("start".to_owned());
+            if emulator == Emulator::Datastore {
                 args.push("--project".to_owned());
                 args.push(self.project);
```
thomaseizinger, line 63: Shouldn't `project be a field of the `Emulator::Datastore` variant if it is only used for that?

src/images/google_cloud_sdk.rs
```diff
@@ -11,60 +10,65 @@ pub const DATASTORE_PORT: u16 = 8081;
...
+                match emulator {
+                    Emulator::Bigtable => "bigtable",
+                    Emulator::Datastore => "datastore",
+                    Emulator::Firestore => "firestore",
+                    Emulator::PubSub => "pubsub",
+                }
+                .to_owned(),
+            );
+
+            args.push("start".to_owned());
```
thomaseizinger, line 60: This could just be a `vec!` no?

```suggestion
			let mut args = vec![
				"gcloud".to_owned(),
				"beta".to_owned(),
				"emulators".to_owned(),
				match emulator {
                    Emulator::Bigtable => "bigtable",
                    Emulator::Datastore => "datastore",
                    Emulator::Firestore => "firestore",
                    Emulator::PubSub => "pubsub",
                }.to_owned(),
                "start".to_owned()
			];
```

thomaseizinger (changes requested):
Thank you for working on this @yottabytt !

Made some comments. Overall this is looking pretty good :)

src/core/image.rs
yottabytt: Actually, based on what you think, I was intending to discuss this: So, right now, few images have `with_tag` as a helper function in their image implementations. How should we go about them? Can we delete them and use `with_tag` from `RunnableImage`?

src/images/google_cloud_sdk.rs
yottabytt: Yeah, definitely. I thought the same. But wasn't sure whether we should separate `Args` from `Image`.

src/core/image.rs
thomaseizinger: Yes we can do that !

tests/images.rs
yottabytt: Sorry for the confusion. I wasn't paying attention to details while I was fixing the tests. The `run` API already gets `image` param as `impl Into<RunnableImage<I>>`. So, just `docker.run(images::coblox_bitcoincore::BitcoinCore::default())` will work since we already have `From` trait implemented for `RunnableImage<I>` to convert from `I: Image` Right?

thomaseizinger: Ah right, of course!

Lets please use that notation then so newcomers learn about this when looking at the tests!

src/images/google_cloud_sdk.rs
thomaseizinger: Sorry, I got confused. We need this because the `Image` instance no longer carries the `Args` right?

Can we add an `impl From(Image, Image::Args) for RunnableImage` and have the constructors on GoogleCloudSdk returna tuple?

src/core/image.rs
yottabytt: I think we can just use `RunnableImage::from((Image, Image::Args))` for this case? and avoid introducing `with_args()`?

src/clients/http.rs
yottabytt: I did see `ImageExt` on your big PR. But wasn't sure why would we really need it. Let's discuss more on that.

src/core/container.rs
yottabytt: WDYT of current approach?

tests/images.rs
```diff
@@ -363,7 +356,8 @@ fn postgres_one_plus_one_with_custom_mapped_port() {
 #[test]
 fn postgres_custom_version() {
     let docker = clients::Cli::default();
-    let postgres_image = images::postgres::Postgres::default().with_version(13);
+    let postgres_image =
+        RunnableImage::<images::postgres::Postgres>::default().with_tag("13-alpine");
```
yottabytt, line 360: I just thought we can have an example test that shows such invocations are possible too. But not sure if this would be preferred by users? WDYT?

src/clients/http.rs
thomaseizinger: > Let's discuss more on that. I guess to avoid users to directly deal with `RunnableImage` ?

Exactly. In particular, the extension trait design achieves the following:

- It "hides" the conversion of `Image` -> `RunnableImage` from the user, making it seem like we are passing instances of `Image` to `run` in all circumstances.
- Contrary to adding default functions on `Image`, functionality defined via the extension trait cannot be overridden. (Overriding default functionality could break certain assumptions and hence broken functionality).
- It automatically applies to all `Image`s, reducing boilerplate for users and at the same time providing a rich interface.

The only downside is that users have to import another trait for this to work but I think that is acceptable :)

tests/images.rs
thomaseizinger: I am having 2nd thoughts on the `Default` impl :grimacing: 

It doesn't actually save you from typing any characters because you still need to add the concrete image type. At that point, you might as well call `images::postgres::Postgres::default()`.

If we add the extension trait, this just becomes:

```rust
let postgres_image = images::postgres::Postgres::default().with_tag("13-alpine");
```

I think I would rather have just one idiomatic way of doing things and remove the `Default` impl again :sweat_smile:
Sorry for sending you on this wild goose chase!

yottabytt: > It doesn't actually save you from typing any characters because you still need to add the concrete image type

Yeah. I felt the same. 😆 

> I think I would rather have just one idiomatic way of doing things and remove the Default impl again 😅
Sorry for sending you on this wild goose chase!

Ah no worries. It was nice to think about and try them and to have these discussions now 😄

thomaseizinger (changes requested):
Thank you very much for working on this! Very excited to land this :)

We can merge this soon! Added a couple of more comments.

src/clients/http.rs
```diff
@@ -88,7 +81,7 @@ impl Http {
 
         // volumes
         let vols: Vec<String> = image
-            .volumes()
+            .env_vars()
```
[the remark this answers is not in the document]
yottabytt: yes 😅  ! will fix.

src/images/google_cloud_sdk.rs
```diff
@@ -123,60 +81,54 @@ impl Image for CloudSdk {
 }
 
 impl CloudSdk {
-    pub fn new(emulator: Emulator) -> Self {
+    pub fn new(port: u16, emulator: Emulator, ready_condition: WaitFor) -> (Self, CloudSdkArgs) {
```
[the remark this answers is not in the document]
yottabytt: WDYT abt we rather make it mis-use resistant by not having `ready_condition` as a param. So, users can use other ports if needed?

thomaseizinger (approved):
Thank you so much!

bors r+

thomaseizinger:
@yottabytt It seems the new nightly compiler flags a couple of now unused fields now: https://github.com/testcontainers/testcontainers-rs/runs/3671810350

yottabytt:
@thomaseizinger I am very glad about the warnings (Rust 😍 ). I was badly wanting to get rid of them. Then it slipped away. Here you go https://github.com/testcontainers/testcontainers-rs/pull/308 !

PR #296, opened 2021-09-02
Fix latest clippy lints

thomaseizinger:


yottabytt:
> thread 'test_produce_and_consume_messages' panicked at 'called `Result::unwrap()` on an `Err` value: KafkaError (Message consumption error: UnknownTopicOrPartition (Broker: Unknown topic or partition))'

ahh I have faced this error. I guess I know what needs to be done (equal chance that it could be a flaky thing). Let me PR one for that.

yottabytt:
Try landing https://github.com/testcontainers/testcontainers-rs/pull/297 instead. I would strongly suggest you to try building that atleast once. I while ago I was playing with Redpanda as well as Kafka. In Redpanda, I was able to consistently observe the error coz of subscribing before producing. Don't exactly remember if it was a flaky thing in Kafka. Unable to notice now though. 

Since there are three things (i.e. `rust-rdkafka`, `librdkafka`, `kafka or redpanda`), it would be a challenging debugging session. 

If we enable logs like [here](https://github.com/testcontainers/testcontainers-rs/pull/294/files#diff-299299a092e8df32a02b485f35cbc9dac37b4e5fa03dad92f5608264d0814a5dR14), there will be a few misleading errors consistently from `librdkafka` in `dev` branch. But they won't be seen consistently in #294 . Don't ask me why :)

thomaseizinger:
Closing in favor of #297.

PR #308, opened 2021-09-22
Fix clippy warnings and simplify structs in images

yottabytt:
Attempt to resolve https://github.com/testcontainers/testcontainers-rs/pull/295#issuecomment-924628336 that also inspired me to further simplify a few things, which also includes addressing https://github.com/testcontainers/testcontainers-rs/pull/295#discussion_r700707178 
> Images that don't need arguments can just leave them as (). We can make our own trait to go from args -> string iterator so we can provide a default impl here.

### DIFF
--- a/src/images/dynamodb_local.rs
+++ b/src/images/dynamodb_local.rs
@@ -4,23 +4,11 @@ const NAME: &str = "amazon/dynamodb-local";
 const TAG: &str = "latest";
 const DEFAULT_WAIT: u64 = 2000;
 
-#[derive(Debug, Default, Clone)]
-pub struct DynamoDbArgs;
-
-impl IntoIterator for DynamoDbArgs {
-    type Item = String;
-    type IntoIter = ::std::vec::IntoIter<String>;
-
-    fn into_iter(self) -> <Self as IntoIterator>::IntoIter {
-        vec![].into_iter()
-    }
-}
-
 #[derive(Default, Debug)]
 pub struct DynamoDb;
 
 impl Image for DynamoDb {
-    type Args = DynamoDbArgs;
+    type Args = ();
 
     fn name(&self) -> String {
         NAME.to_owned()
--- a/src/images/dynamodb_local.rs
+++ b/src/images/dynamodb_local.rs
@@ -1,8 +1,8 @@
 use crate::{core::WaitFor, Image};
 
 const NAME: &str = "amazon/dynamodb-local";
+const TAG: &str = "latest";
 const DEFAULT_WAIT: u64 = 2000;
-const DEFAULT_TAG: &str = "latest";
 
 #[derive(Debug, Default, Clone)]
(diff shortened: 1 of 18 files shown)


src/images/kafka.rs
```diff
@@ -1,44 +1,41 @@
...
+            ]
+            .into_iter(),
+        )
     }
 }
 
 #[derive(Debug)]
 pub struct Kafka {
-    arguments: KafkaArgs,
     env_vars: HashMap<String, String>,
```
yottabytt, line 38: I am kinda neutral (slightly leaning towards the latter) on whether we should have fields like these: here (or) just have them within `impl Image` blocks under respective methods and just `[derive(Default)]` here.

thomaseizinger: I don't fully follow? Can you give an example of both?

src/images/dynamodb_local.rs
```diff
@@ -1,7 +1,7 @@
 use crate::{core::WaitFor, Image};
 
 const NAME: &str = "amazon/dynamodb-local";
-const TAG: &str = "latest";
```
thomaseizinger: If I remember correctly, this was purposefully set to `latest` because `latest` always refers to what is deployed on AWS so testing your app against that makes sense.

src/images/generic.rs
```diff
@@ -55,7 +55,7 @@ impl GenericImage {
 }
 
 impl Image for GenericImage {
-    type Args = Vec<String>;
```
thomaseizinger, line 60: If we do this, then we cannot pass any arguments for generic images. I think this has to stay!

thomaseizinger (changes requested):
Thanks! Very nice cleanup :)

Some comments!

src/images/dynamodb_local.rs
yottabytt: Ah I see. cool then. will drop that commit.

src/images/kafka.rs
yottabytt: Sure. As you can see in https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f899fe85e2d4bb8a79d5f76b07f0e373 - why to define like `Foo` when can also define like `Bar` in the era of `RunnableImage` ?

thomaseizinger: A fair enough yes. I am in favor of the latter as well.

yottabytt: My bad that I didn't realize that we also need to change the return types of the methods `env_vars` and `volumes` in `Image` trait to `Box<dyn Iterator<Item = (String, String)>`. I'll think about this and we shall discuss more on this when I start to work on `ImageExt`. For now, I am leaving this as it is.

thomaseizinger (approved):
Thank you!

bors r+

PR #311, opened 2021-10-27
Support registries other than Docker Hub 

quettabit:
Hey @thomaseizinger ! 

We think this would be super helpful for our use cases. Lemme know WDYT. Have added more info in the commit messages.

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -170,9 +170,14 @@ impl Client {
             command.arg("-P"); // publish all exposed ports
         }
 
+        let image_path = match image.registry_prefix() {
+            Some(prefix) => format!("{}/{}", prefix, image.descriptor()),
+            None => image.descriptor(),
+        };
+
         command
             .arg("-d") // Always run detached
-            .arg(image.descriptor())
+            .arg(image_path)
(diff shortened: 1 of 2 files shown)


thomaseizinger (commented):
Thanks!

As far as I know, the registry is part of the container identifier with Docker Hub simply being the default.

Shouldn't we implement this as a default fn on the trait then where each image can define, which registry it is located at?

Some registries might be mirrors but not every registry serves every image.

Ideally, all images would define their identifiers as fully-qualified. Hence, if an image is located at a different registry, you can already use that image by just specifying its registry as part of the identifier.

Do I understand you correctly that you want to override, what the default registry is? Is there a way of doing this in the CLI without changing the image identifier? Should probably be a setting on `Cli` then!

LegNeato:
Fwiw, the go library matches this patch rather than putting it in the CLI...you can pass a full image name, which can have the registry pretended and it will parse it and behave correctly.

DDtKey:
Closing as outdated, but this is gonna be supported soon.
Either by #549 or by introducing another solution aligned with implementation for other languages.

Thanks!

PR #312, opened 2021-10-28
Use emulators tag for google cloud-sdk

yottabytt:
To keep things lightweight, it is better to use the image that
has only the emulators and its dependencies. Also, included
support for Spanner emulator which was missed out earlier.


thomaseizinger (approved):
Sorry for the late reply, I was on vacation :)

bors r+

Issue #279, opened 2021-03-19
Not working on Windows 10


Weasy666:
When running the `dev`-branch on Windows, i get a few errors regarding `tokio::net::UnixListener`, like this
```rust
error[E0432]: unresolved imports `tokio::net::UnixListener`, `tokio::net::UnixStream`
  --> C:\Users\weasy\.cargo\registry\src\github.com-1ecc6299db9ec823\hyperlocal-0.8.0\src\server.rs:17:22
   |
17 |     use tokio::net::{UnixListener, UnixStream};
   |                      ^^^^^^^^^^^^  ^^^^^^^^^^ no `UnixStream` in `net`
   |                      |
   |                      no `UnixListener` in `net`
   |                      help: a similar name exists in the module: `TcpListener`

error[E0433]: failed to resolve: could not find `UnixStream` in `net`
```
Which makes perfect sense, because the Unix stuff does not work on Windows. As far as u could tell, `hyperlocal` is a dependency of `shiplift` and not necessary for the `Cli` implementation. Could you possibly put `shiplift` behind a feature flag? That would enable us to compile and use `testcontainers` with its `Cli` feature.

thomaseizinger:
> Could you possibly put shiplift behind a feature flag?

The proper fix for this would be upstream: https://github.com/softprops/shiplift/issues/241

As a work-around, we can conditionally set the necessary features on shiplift based on the target platform. I currently don't have time to do this but happy to merge a PR for this if it is important to you that `dev` works on Windows. Alternatively, I'd recommend to use the latest release.

dzmitry-lahoda:
On windows docker talks named pipes.

Issue #271, opened 2021-02-19
Logs of shiplift client don't `follow`


thomaseizinger:

yottabytt:
Hey @thomaseizinger , while thinking of TODOs (esp. async support over http client and some hacking around it) for #293 and #294 , I guess I landed upon this bug. IIRC, it ended up at [this error](https://github.com/testcontainers/testcontainers-rs/blob/dev/src/core/logs.rs#L79). In our team at [Etsy](https://en.wikipedia.org/wiki/Etsy), we rely on async a lot whenever possible. It would be great if we get both #293 and #294 with async support. It's also fine to add that support as an immediate follow-up to those PRs (I can also drive that if thats okay). Thoughts?

thomaseizinger:
Yes, chunking things into smaller PRs is totally fine!

thomaseizinger:
Unfortunately, I have little time at the moment to code myself but I'll try to be responsive to anything on here :)

yottabytt:
>I'll try to be responsive to anything on here 

Awesome. That would be really helpful. Thank you :)

thomaseizinger:
Fixed in #323.

PR #323, opened 2022-01-14
Replace all Shiplift usages with Bollard

andrey-yantsen:
Judging by the low activity in [shiplift](https://github.com/softprops/shiplift) (last commit on 1 Sep 2021, quite a few unmerged PRs after September and lack of response in the maintenance status clarification request softprops/shiplift#321), I think it's a good time to switch to a more actively maintained crate.

I picked Ballard because of the more active repo than `docker-api-rs`. And I liked the idea with the code generation from the API :)

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -2,7 +2,7 @@ use crate::{
     core::{env, env::GetEnvValue, logs::LogStream, ports::Ports, ContainerState, Docker, WaitFor},
     Container, Image, ImageArgs, RunnableImage,
 };
-use shiplift::rep::ContainerDetails;
+use bollard::models::ContainerInspectResponse;
 use std::{
     collections::HashMap,
     ffi::{OsStr, OsString},
@@ -298,12 +298,13 @@ impl Docker for Cli {
     fn ports(&self, id: &str) -> Ports {
         self.inspect(id)
             .network_settings
+            .unwrap_or_default()
             .ports
-            .map(Ports::new)
+            .map(Ports::from)
             .unwrap_or_default()
     }
 
-    fn inspect(&self, id: &str) -> ContainerDetails {
+    fn inspect(&self, id: &str) -> ContainerInspectResponse {
         let child = self
             .inner
             .command()
@@ -315,7 +316,7 @@ impl Docker for Cli {
 
         let stdout = child.stdout.unwrap();
 
-        let mut infos: Vec<ContainerDetails> = serde_json::from_reader(stdout).unwrap();
+        let mut infos: Vec<ContainerInspectResponse> = serde_json::from_reader(stdout).unwrap();
 
         let info = infos.remove(0);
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   This reduces the API surface of the crate which allows for fewer breaking changes in the future.
   All functionality from `Docker` (start, stop, rm, and ports) is available on a container directly.
 - `descriptor` is broken down into `name` and `tag` within `Image` trait.
+- [shiplift](https://github.com/softprops/shiplift) replaced with [bollard](https://github.com/fussybeaver/bollard).
```
thomaseizinger, line 37: Okay, so this is a bit embarrassing. I've been holding off to make a release for so long that we are replacing the initial implementation of this with something else and the initial one wasn't even release yet!

As such, I think we don't need to mention this because for people upgrading from `0.12`, they didn't even know that this was even based on `shiplift` ...

thomaseizinger (commented):
Thank you!

concept ACK but I think we can omit the changelog entry. See comment!

src/clients/cli.rs
```diff
@@ -2,7 +2,7 @@ use crate::{
     core::{env, env::GetEnvValue, logs::LogStream, ports::Ports, ContainerState, Docker, WaitFor},
     Container, Image, ImageArgs, RunnableImage,
 };
-use shiplift::rep::ContainerDetails;
+use bollard::models::ContainerInspectResponse;
```
thomaseizinger, line 5: If we would inline this struct, then we could feature-flag `bollard` on the `experimental`. What do you think?

CHANGELOG.md
andrey-yantsen: Ha! :) Yeah, I know a thing or two about delaying a release :) Removed.

src/clients/cli.rs
andrey-yantsen: Interesting idea. Not sure about inlining the struct, though — we'd have to track the changes in the API, and it's not fun :) We could use [bollard-stubs](https://crates.io/crates/bollard-stubs) (it's the crate where bollard stores the autogenerated models) inside CLI; it should be enough. Are you OK with it?

andrey-yantsen:
Anyway, I pushed the usage of `bollard-stubs` and made `bollard` dependency optional (enabling it only for the `experimental` feature) :) If you are not happy with it — I'm open to suggestions.

P.S. I just discovered that the switch to `bollard` fixes #279. There're a couple more shiplift-related issues: #270 and #271; I haven't checked those yet.

src/clients/cli.rs
thomaseizinger: Yep, that is much better! Certainly more lightweight!

thomaseizinger:
Thank you very much for this effort. Nice that this fixes #279!

bors r+

PR #325, opened 2022-01-15
Refactor dependencies

andrey-yantsen:
This PR aims to reduce the build time for the users who don't need the experimental features and don't need the bitcoincore image.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,10 +21,10 @@ rand = "0.8"
 serde = { version = "1", features = [ "derive" ] }
 serde_json = "1"
 sha2 = "0.10"
-tokio = { version = "1", features = [ "macros" ] }
+tokio = { version = "1", features = [ "macros" ], optional = true }
 
 [features]
-experimental = [ "async-trait", "bollard" ]
+experimental = [ "async-trait", "bollard", "tokio" ]
 
 [dev-dependencies]
 bitcoincore-rpc = "0.14"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/testcontainers/testcontainers-rs"
 description = "A library for integration-testing against docker containers from within Rust."
 
 [dependencies]
-async-trait = "0.1"
+async-trait = { version = "0.1", optional = true }
 bollard = { version = "0.11", optional = true }
 bollard-stubs = "1.41"
 futures = "0.3"
@@ -24,7 +24,7 @@ sha2 = "0.10"
 tokio = { version = "1", features = [ "macros" ] }
 
 [features]
-experimental = [ "bollard" ]
+experimental = [ "async-trait", "bollard" ]
 
 [dev-dependencies]
 bitcoincore-rpc = "0.14"


thomaseizinger (commented):
I am okay with hiding `async-trait` given that it is a proc-macro but would like to avoid feature-flagging images. The now feature-flagged dependencies are extremely common and fairly small. How much build-time improvement did you measure?

andrey-yantsen:
To be honest, I didn't make the comparison before creating the PR, but I did it now! :)

Features=default:
* Current branch, 122 deps, avg of 5 — 16.49s
* Dev, 144 deps, avg of 5 — 17.36s

Features=default,experimental
* Current branch, 125 deps, avg of 5 — 18.85s
* Dev, 144 deps, avg of 5 — 19.63s

I checked on my Macbook Air with M1, and the results obviously could be different with another CPU. ≈1 second improvement doesn't sound much (especially with incremental compilation). However, I still think it's worth "removing" 19-22 dependencies not required for some end-users.

To be fair, I have little experience with Rust, and I don't have a strong opinion about the dependencies management, so I'll be OK with any decision :)

thomaseizinger:
> Features=default:
> 
>     * Current branch, 122 deps, avg of 5 — 16.49s
> 
>     * Dev, 144 deps, avg of 5 — 17.36s
> 
> 
> Features=default,experimental
> 
>     * Current branch, 125 deps, avg of 5 — 18.85s
> 
>     * Dev, 144 deps, avg of 5 — 19.63s

Unless I am missing something, this comparison does not measure the benefit of feature-flagging `image_coblox_bitcoincore`? :)
I am all for optimizing compile-times but would like to do so without making the library too complex to use.

While looking at the diff though, I don't remember why we are depending on `tokio` in our `[dependencies]`? That sounds like something we should be able to move to `[dev-dependencies]`!

andrey-yantsen:
> Unless I am missing something, this comparison does not measure the benefit of feature-flagging image_coblox_bitcoincore? :)

I might've just misunderstood you :) I compared the compilation time of the `dev` branch (where the `coblox_bitcoincore` and its dependencies are compiled) and the current branch (without `coblox_bitcoincore`) — that shows us the build-time improvement for the users who don't need the `coblox_bitcoincore`; what did you want to see?

> I am all for optimizing compile-times but would like to do so without making the library too complex to use.

The problem here is that I don't know how many users require the bitcoincore image. If it's a small-ish per cent, then the majority will benefit from making the single image's dependencies (and the image itself) optional, but I could only guess here.

> While looking at the diff though, I don't remember why we are depending on tokio in our [dependencies]? That sounds like something we should be able to move to [dev-dependencies]!

Curiously enough, it's also in dev-dependencies :) Unfortunately, we can't just remove it completely because of the single usage: https://github.com/testcontainers/testcontainers-rs/blob/dev/src/core/container_async.rs#L191. Still, I made it optional too, behind the `experimental` feature flag.

thomaseizinger:
> > Unless I am missing something, this comparison does not measure the benefit of feature-flagging image_coblox_bitcoincore? :)
> 
> I might've just misunderstood you :) I compared the compilation time of the `dev` branch (where the `coblox_bitcoincore` and its dependencies are compiled) and the current branch (without `coblox_bitcoincore`) — that shows us the build-time improvement for the users who don't need the `coblox_bitcoincore`; what did you want to see?

Ah right, my bad! Makes total sense now :)

> > I am all for optimizing compile-times but would like to do so without making the library too complex to use.
> 
> The problem here is that I don't know how many users require the bitcoincore image. If it's a small-ish per cent, then the majority will benefit from making the single image's dependencies (and the image itself) optional, but I could only guess here.

I'd like to not make them optional for now. The gains are marginal and the dependencies it pulls in are so common that they are very likely already in the dev-dependency tree of an application.

> > While looking at the diff though, I don't remember why we are depending on tokio in our [dependencies]? That sounds like something we should be able to move to [dev-dependencies]!
> 
> Curiously enough, it's also in dev-dependencies :) Unfortunately, we can't just remove it completely because of the single usage: [`dev`/src/core/container_async.rs#L191](https://github.com/testcontainers/testcontainers-rs/blob/dev/src/core/container_async.rs?rgh-link-date=2022-01-17T23%3A12%3A52Z#L191). Still, I made it optional too, behind the `experimental` feature flag.

Ah nice, thank you!

andrey-yantsen:
Sure thing, no problem! I've dropped the commit about the image, leaving only the changes regarding `async-trait` and `tokio`.

andrey-yantsen:
```
authors = [ "CoBloX developers <team@coblox.tech>" ]
```

I haven't noticed this bit previously. Fair enough :D

thomaseizinger (approved):
Thanks!

bors r+

thomaseizinger:
> ```
> authors = [ "CoBloX developers <team@coblox.tech>" ]
> ```
> 
> I haven't noticed this bit previously. Fair enough :D

Yes although the usage / features of this lib have long extended beyond our internal needs :)

PR #327, opened 2022-01-18
Add WaitFor::Healthcheck

### LINKED ISSUE #321: Add WaitFor::Healthy
Currently implemented waiting conditions can't guarantee that the application in the container is ready and is working correctly.

Some of the containers have healthchecks (e.g. [plexinc/pms-docker](https://hub.docker.com/r/plexinc/pms-docker)) and for those it could be useful to rely on the healthcheck results instead of processing stdout/stderr.

andrey-yantsen:
In this PR, I added the most naive and simple implementation for the `WaitFor::Healthcheck`. The ideal option will be to remove `delay` from the Enum. Still, the current implementation gives some flexibility to the users and eliminates the need of hardcoding the delay inside the crate.

As always, I'm open to suggestions :)

Fixes #321.

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -2,7 +2,7 @@ use crate::{
     core::{env, env::GetEnvValue, logs::LogStream, ports::Ports, ContainerState, Docker, WaitFor},
     Container, Image, ImageArgs, RunnableImage,
 };
-use bollard_stubs::models::ContainerInspectResponse;
+use bollard_stubs::models::{ContainerInspectResponse, HealthStatusEnum};
 use std::{
     collections::HashMap,
     ffi::{OsStr, OsString},
@@ -400,6 +400,25 @@ impl Docker for Cli {
                 WaitFor::Duration { length } => {
                     std::thread::sleep(length);
                 }
+                WaitFor::Healthcheck => loop {
+                    use HealthStatusEnum::*;
+
+                    let health_status = self
+                        .inspect(id)
+                        .state
+                        .unwrap_or_else(|| panic!("Container state not available"))
+                        .health
+                        .unwrap_or_else(|| panic!("Health state not available"))
+                        .status;
+
+                    match health_status {
+                        Some(HEALTHY) => break,
+                        None | Some(EMPTY) | Some(NONE) => {
+                            panic!("Healthcheck not configured for container")
+                        }
+                        Some(STARTING) | Some(UNHEALTHY) => sleep(Duration::from_millis(100)),
+                    }
+                },
                 WaitFor::Nothing => {}
             }
         }
--- a/src/core/image.rs
+++ b/src/core/image.rs
@@ -289,6 +289,8 @@ pub enum WaitFor {
     StdErrMessage { message: String },
     /// Wait for a certain amount of time.
     Duration { length: Duration },
+    /// Wait for the container's status to become `healthy`.
+    Healthcheck,
 }
 
 impl WaitFor {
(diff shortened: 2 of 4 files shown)


src/clients/cli.rs
```diff
@@ -400,6 +402,22 @@ impl Docker for Cli {
...
+                    }) = container.state
+                    {
+                        if health.status == Some(HealthStatusEnum::HEALTHY) {
+                            break;
+                        }
+
+                        sleep(delay);
+                    } else {
+                        panic!("Healthcheck for the container is not configured");
+                    }
```
thomaseizinger, line 419: You should be able to write this as a match:

```suggestion
                    match container.state {
                    	Some(BollardContainerState {
                            health: Some(HealthStatusEnum::HEALTHY),
                            ..
                        }) => break,
                    	Some(BollardContainerState {
                            health: Some(HealthStatusEnum::UNHEALTHY), // TODO: CHeck if correct enum variant
                            ..
                        }) => sleep(delay),
                        Some(BollardContainerState {
                            health: None,
                            ..
                        }) => panic!("Healthcheck for the container is not configured"),
                        _ => // handle other cases
                    }
```

src/core/image.rs
```diff
@@ -289,6 +289,9 @@ pub enum WaitFor {
     StdErrMessage { message: String },
     /// Wait for a certain amount of time.
     Duration { length: Duration },
+    /// Wait for the container's status to become `healthy`.
+    /// Delay is the pause between the consequitive container status checks.
+    Healthcheck { delay: Duration },
```
thomaseizinger, line 294: I think we can easily hardcode a value here. Can start with 100ms!

thomaseizinger (commented):
Some ideas!

src/clients/cli.rs
andrey-yantsen: Well, while I was rewriting it as a match, I realised two things:
1. I missed another case of how docker can report that the health check is not configured — it can be reported either by `ContainerState.state == None` or by `ContainerState.state.health.status == HealthStatusEnum::NONE`.
2. Match can be ugly! :) But I'm not sure how to avoid a tiny bit of repetition without it.

Which one of the following would you prefer? I like the second option, but at the same time, having to write the same `panic!` twice disappoints me.

image
image

andrey-yantsen: Anyway, I force-pushed the changes from the second screenshot. Feel free to disagree :)

thomaseizinger:
@andrey-yantsen I've pushed an alternative idea as a commit on top of this. Feel free to squash it in if you like it :)

(Please also double check that it is functionally equivalent :D)

thomaseizinger:
Oh, I see we would have to raise MSRV for that. I'd be okay with that, we are a test-only library.

andrey-yantsen:
This looks much nicer, thanks! I have replaced `Some(STARTING | UNHEALTHY)` with the good old `Some(STARTING) | Some(UNHEALTHY)`, I hope you don't mind. TBH, I don't feel like a few bytes are worth bumping MSRV 7 releases (to 1.53). On the other hand, it was released six months ago... Anyway, feel free to revert that commit if you want the bump :)

thomaseizinger:
> This looks much nicer, thanks! I have replaced `Some(STARTING | UNHEALTHY)` with the good old `Some(STARTING) | Some(UNHEALTHY)`, I hope you don't mind. TBH, I don't feel like a few bytes are worth bumping MSRV 7 releases (to 1.53). On the other hand, it was released six months ago... Anyway, feel free to revert that commit if you want the bump :)

Yeah that totally works, thanks!

Do you mind squashing all this into one commit?

andrey-yantsen:
@thomaseizinger done :)

PR #328, opened 2022-01-18
Add MSRV to Cargo.toml

andrey-yantsen:
cargo officially allows key `rust-version` starting from Rust 1.56.0. In the older versions, the key will be ignored (though a warning will be displayed).

Having this value defined in Cargo.toml provides better visibility within the ecosystem. And it's just nicer to have this value specified in the config and not in a Github Workflow, even if reading it inside the workflow is a bit clunky :)


thomaseizinger (approved):
Cool, thanks!

bors r+

PR #334, opened 2022-02-17
Add MinIO image

DDtKey:
I think it should be pretty useful to have a `S3`-compatible default image.

Features:
- Specify server args via `MinIOServerArgs`

In addition, one test case was added.

### DIFF
--- a/src/images/minio.rs
+++ b/src/images/minio.rs
@@ -1,0 +1,79 @@
+use crate::{core::WaitFor, Image, ImageArgs};
+use std::collections::HashMap;
+
+const NAME: &str = "minio/minio";
+const TAG: &str = "RELEASE.2022-02-07T08-17-33Z";
+
+const DIR: &str = "/data";
+const CONSOLE_ADDRESS: &str = ":9001";
+
+#[derive(Debug)]
+pub struct MinIO {
+    env_vars: HashMap<String, String>,
+}
+
+impl Default for MinIO {
+    fn default() -> Self {
+        let mut env_vars = HashMap::new();
+        env_vars.insert(
+            "MINIO_CONSOLE_ADDRESS".to_owned(),
+            CONSOLE_ADDRESS.to_owned(),
+        );
+
+        Self { env_vars }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct MinIOServerArgs {
+    pub dir: String,
+    pub certs_dir: Option<String>,
+    pub json_log: bool,
+}
+
+impl Default for MinIOServerArgs {
+    fn default() -> Self {
+        Self {
+            dir: DIR.to_owned(),
+            certs_dir: None,
+            json_log: false,
+        }
+    }
+}
+
+impl ImageArgs for MinIOServerArgs {
+    fn into_iterator(self) -> Box<dyn Iterator<Item = String>> {
+        let mut args = vec!["server".to_owned(), self.dir.to_owned()];
+
+        if let Some(ref certs_dir) = self.certs_dir {
+            args.push("--certs-dir".to_owned());
+            args.push(certs_dir.to_owned())
+        }
+
+        if self.json_log {
+            args.push("--json".to_owned());
+        }
+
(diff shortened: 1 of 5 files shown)


src/images/minio.rs
```diff
@@ -0,0 +1,88 @@
...
+    fn name(&self) -> String {
+        NAME.to_owned()
+    }
+
+    fn tag(&self) -> String {
+        self.tag.to_owned()
+    }
+
+    fn ready_conditions(&self) -> Vec<WaitFor> {
+        vec![WaitFor::message_on_stdout("API:")]
```
thomaseizinger, line 82: That seems odd. This really uniquely identifies a fully started container?

src/images/minio.rs
```diff
@@ -0,0 +1,88 @@
...
+        Self::new_with_tag(TAG.to_owned())
+    }
+}
+
+impl MinIO {
+    pub fn new(tag: String, env_vars: HashMap<String, String>) -> Self {
+        Self { tag, env_vars }
+    }
+
+    pub fn new_with_tag(tag: String) -> Self {
```
thomaseizinger, line 27: The builder on `RunnableImage` already allows to override this!

src/images/minio.rs
```diff
@@ -0,0 +1,88 @@
+use crate::{core::WaitFor, Image, ImageArgs};
+use std::collections::HashMap;
+
+const NAME: &str = "minio/minio";
+const TAG: &str = "latest";
```
thomaseizinger, line 5: Is this some Amazon campatibility thing again? I'd love to use fixed tags here.

thomaseizinger (commented):
Thanks! Some comments.

src/images/minio.rs
DDtKey: image

DDtKey: Full output:
```sh
API: http://172.17.0.2:9000  http://127.0.0.1:9000 

Console: http://172.17.0.2:9001 http://127.0.0.1:9001 

Documentation: https://docs.min.io
```

src/images/minio.rs
DDtKey: I think they can coexist 🤔 
The tag is quite an important part and it's easier to specify it in the constructor than to create a wrapper.
```rs
// RunnableImage approach
let image = RunnableImage::from(images::minio::MinIO::default())
        .with_tag("latest");
        
// Direct approach        
let image = images::minio::MinIO::with_tag("latest");

```

If you disagree, let me know, I'll remove the constructors. No problems.

src/images/minio.rs
DDtKey: I think they can coexist 🤔 
The tag is quite an important part and it's easier to specify it in the constructor than to create a wrapper.
```rs
// RunnableImage approach
let image = RunnableImage::from(images::minio::MinIO::default())
        .with_tag("latest");
        
// Direct approach        
let image = images::minio::MinIO::with_tag("latest");

```

If you disagree, let me know, I'll remove the constructors. No problems.

DDtKey: Perhaps this distorts the general interface for working with images, but it seems worth documenting `RunnableImage` 🤔

src/images/minio.rs
DDtKey: I agree, fixed the tag

src/images/minio.rs
thomaseizinger: You don't need to create the wrapper, there is an extension trait so you can call `with_tag` directly on the image :)

DDtKey:
I've updated PR and fixed MSRV + style warnings. Thanks!

thomaseizinger:
> I've updated PR and fixed MSRV + style warnings. Thanks!

Sorry for the late reply!
Do you mind squashing all commits into one please?

After that, feel free to merge the PR. I'll delegate permissions now so that I don't have to come back to it. Trusting you that you will squash first ;)

bors d=DDtKey

PR #341, opened 2022-03-06
Fix WaitFor::Healthcheck

kezhuw:
* Reclaim zombie docker inspect command process
* Panic on unhealthy

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -305,18 +305,22 @@ impl Docker for Cli {
     }
 
     fn inspect(&self, id: &str) -> ContainerInspectResponse {
-        let child = self
+        let output = self
             .inner
             .command()
             .arg("inspect")
             .arg(id)
             .stdout(Stdio::piped())
-            .spawn()
+            .output()
             .expect("Failed to execute docker command");
+        assert!(
+            output.status.success(),
+            "Failed to inspect docker container"
+        );
 
-        let stdout = child.stdout.unwrap();
+        let stdout = output.stdout;
 
-        let mut infos: Vec<ContainerInspectResponse> = serde_json::from_reader(stdout).unwrap();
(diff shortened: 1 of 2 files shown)


kezhuw:
Without this fix, it could causes `WaitFor::Healthcheck` blocking indefinitely and fails with `runtime/cgo: pthread_create failed: Resource temporarily unavailable` finally due to zombie `docker inspect` processes.

I did not add test case for this fix as this will introduce extra image pulling. But you could see https://github.com/kezhuw/testcontainers-rs/commits/645a37fd33dcd8f145152e2f35e6dc9c1e8e9766.

kezhuw:
I have added an ignored test case in #343 using healthcheck customization.

kezhuw:
Ping @andrey-yantsen for review.

andrey-yantsen:
@kezhuw, hey :) I'm not sure if `panic` is a good idea because we can't know why the container is `unhealthy` (maybe we have a slow CPU, and the container takes a lot of time to start everything up). IMO a better solution would be to introduce a timeout, i.e. if the status is still not `healthy` after X seconds, it's time to panic.

P.S. And yeah, I still don't have the write access :)

kezhuw:
@andrey-yantsen [`HEALTHCHECK`](https://docs.docker.com/engine/reference/builder/#healthcheck) itself has time control. Together with #343, caller call tweak time options. I guess it is a convention in `testcontainers-rs` to panic when condition unmet.

From code, there is no panic in `STARTING` phase. I quoted docs from docker links.
> When a container has a healthcheck specified, it has a health status in addition to its normal status. This status is initially starting. Whenever a health check passes, it becomes healthy (whatever state it was previously in). After a certain number of consecutive failures, it becomes unhealthy.

andrey-yantsen:
Ah, yes, there's `--start-period`. Thanks!

thomaseizinger (approved):
Yep that looks good to me!

Thank you :)

bors r+

PR #343, opened 2022-03-07
Support healthcheck customization

### LINKED ISSUE #342: Support healthcheck customization in docker run
Some docker images do not have `HEALTHCHECK`, but they do have commands to verify its health status. Support of healthcheck customization can avoid ad-hoc image.

Given zookeeper(#255) as an example, `./bin/zkServer.sh status` could be used as parameter to `--health-cmd` options in `docker run`.

kezhuw:
Sees #341.
Closes #342.

### DIFF
--- a/src/clients/cli.rs
+++ b/src/clients/cli.rs
@@ -157,6 +157,30 @@ impl Client {
             command.arg("--entrypoint").arg(entrypoint);
         }
 
+        if let Some(healthcheck) = image.healthcheck() {
+            if let Some(cmd) = healthcheck.cmd {
+                command.arg("--health-cmd").arg(cmd.join(" "));
+            }
+            if let Some(interval) = healthcheck.interval {
+                command
+                    .arg("--health-interval")
+                    .arg(interval.as_secs().to_string() + "s");
+            }
+            if let Some(timeout) = healthcheck.timeout {
+                command
+                    .arg("--health-timeout")
+                    .arg(timeout.as_secs().to_string() + "s");
+            }
+            if let Some(start_period) = healthcheck.start_period {
+                command
+                    .arg("--health-start-period")
+                    .arg(start_period.as_secs().to_string() + "s");
+            }
+            if let Some(retries) = healthcheck.retries {
+                command.arg("--health-retries").arg(retries.to_string());
+            }
+        }
+
         let is_container_networked = image
             .network()
             .as_ref()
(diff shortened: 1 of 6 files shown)


kezhuw:
Ping @andrey-yantsen for review.

andrey-yantsen:
@kezhuw why me though? :) I don't have write access to the repo.

kezhuw:
O_O

I found through recent commits. Seems that this repo using bots to merge pr. Sorry for interruption.

Ping @thomaseizinger for review.

thomaseizinger:
Thanks for the PR!

I am not fully sold on adding this feature.

We support the `WaitFor::Healthcheck` command so that images which do have a `HEALTHCHECK` instruction can easily be waited on.
If an image does not support that and you need to add the healthcheck command manually, you might as well not use `WaitFor::Healthcheck` and instead just express directly using other `WaitFor` instructions what needs to happen for the image to be ready. Now, it is debatable whether `WaitFor` can express everything that is necessary. I am open to add other `WaitFor` strategies but I'd like to not add too many ways of doing the same thing.

For now, I think it is out of scope of testcontainers-rs to "extend" a container this way. I'd recommend to either:

- Express the "waiting" condition using existing `WaitFor` possibilities
- Add a HEALTHCHECK instruction to the upstream image you are using
- Fork the upstream image or create a new image that is created `FROM` the parent image and add your own HEALTHCHECK instruction

kezhuw:
I think `HEALTHCHECK` customization in testing code has advantages over other approaches:
* `WaitFor::Xyz`: One purpose of this customization to support command execution to detect readiness. I guess it will be similar to "--health-xyz" options in `docker run` if `WaitFor` support similar functionality in future.
* Build dedicated images. Images are hard coded artifacts. It is much difficult to build/publish images (possible in upstream repos) than customization in testing code. Besides, healthcheck customization enables fine control per project(different projects may have different testing environments) other than per image.

thomaseizinger:
If you truly require custom waiting strategies based on commands, you can use `Container#exec`: https://github.com/testcontainers/testcontainers-rs/blob/8d937a8ee53eb9d86a219578c65c10d4a7564db9/src/core/container.rs#L143-L155

kezhuw:
Then I am reinventing the wheel.
* I need a command to execute. It is `--health-cmd`.
* What if the command fails ? Should I retry ?
   * What is the retry interval ? It is `--health-interval`.
   * How many times should I retry ? It is `--health-retries`.
 * What if the command blocks without responding ? Should I consider it as failure after some timeout ? It is `--health-timeout`.
 * Next will be the journey of `--health-start-period`(moby/moby#26664)

PR #345, opened 2022-03-19
Add Image For LocalStack

ryskit:
[LocalStack](https://hub.docker.com/r/localstack/localstack)  Image was needed for development, so I added it. 

In addition, three test cases(SNS, SQS, DynamoDB) was added.

@thomaseizinger 
Please review when you have time!!


thomaseizinger (commented):
Thanks!

I've never heard of `localstack` but looking at the image definition it does not look particularly useful to me.
Not all images need to be part of this library :)
Especially when they are as simple as this, it doesn't seem worth the effort of maintaining it in here.

Thanks anyway though!

Issue #348, opened 2022-03-25
[feature-request] support specifying expose_ports in GenericImage


lukehsiao:
https://github.com/testcontainers/testcontainers-rs/pull/293 added the great new feature of manually exposing ports in an image. This is great for images that are defined in testcontainers (like the Google Cloud SDK in that PR).

However, this functionality would also be very useful to GenericImage, for users using images that are not first-class images in testcontainers.

To expose this feature more generally, I'd propose that GenericImage take a new builder option: `with_exposed_ports(self, ports: Vec<u16>)` with in turn sets self.exposed_ports.

Happy to work on a PR for this if it sounds like a good idea.

lukehsiao:
I've added a simple implementation to #349 for review.

Issue #187, opened 2020-07-08
Split testing public API into a separate crate


thomaseizinger:
This allows us to build containers dedicated for testing locally using a buildscript. See https://github.com/testcontainers/testcontainers-rs/pull/174 for more context.

PR #349, opened 2022-03-25
Add support to expose ports in GenericImage

lukehsiao:
By default, testcontainers-rs runs docker with `-P`. However, for a user
to get the port mappings, this relies on the image itself properly
exposing ports using the EXPOSE instruction [[1]]. When images do not use
this instruction, a user can manually expose these ports using the
`--expose` option in the docker CLI.

However, testcontainers-rs does not allow users to specify this argument
in GenericImage. This commit exposes a new method, with_exposed_port()
as part of a GenericImage builder that allows a user to properly expose
ports for any generic image.

[1]: https://docs.docker.com/engine/reference/builder/#expose
Closes: https://github.com/testcontainers/testcontainers-rs/issues/348


thomaseizinger (approved):
That makes sense to me, thank you!

bors r+

thomaseizinger:
> Would you happen to be aware of a simpler docker image that could take Vespa's place? Ideally it would be small, not require a delay to be ready, and also not expose its ports.

We had the idea at some point to maintain a list of docker images _within_ this repository that we use for testing. There is an issue here: https://github.com/testcontainers/testcontainers-rs/issues/187 which links to a further discussion.

If you are keen, you could make yourself exactly the docker image you need for the test with this approach! I don't mind about the CI runtime too much as we don't have a whole lot of PRs going in here anyway. But if you are keen to experiment with some of the ideas outlined in https://github.com/testcontainers/testcontainers-rs/issues/187 and the linked discussion, it would be greatly appreciated :)

thomaseizinger:
> > If you are keen, you could make yourself exactly the docker image you need for the test with this approach!
> 
> I've done just this. A dead-simple http server for this purpose :).
> 
> Source is here: [lukehsiao/hello-world](https://github.com/lukehsiao/hello-world)

Thanks!

It would be great if we could maintain these images _within_ the repository. I would like to avoid depending on external repositories for this. It makes it unclear, where the image is coming from, who is meant to maintain it and it bears the risk of our CI breaking if the image ever disappears.

If you'd like to use this image, it would be great if we can move it into this repository as part of the test. The linked discussion above has some ideas on how to do that but in a nutshell:

- Separate crate with build-script
- Introduce `dev-depenency` to said crate
- Build and tag docker images as part of the build-script

This allows us to use the images without having to push them anywhere and we can refer to them by name in our tests.

lukehsiao:
> If you'd like to use this image, it would be great if we can move it into this repository as part of the test. The linked discussion above has some ideas on how to do that but in a nutshell:
> 
> * Separate crate with build-script
> * Introduce `dev-depenency` to said crate
> * Build and tag docker images as part of the build-script
> 
> This allows us to use the images without having to push them anywhere and we can refer to them by name in our tests.

Apologies, as I'm still learning Rust. To do so, you're suggesting make a new crate *within* this repository? Are you suggesting we organize the repository into workspaces (i.e., the testcontainers library crate, and a separate crate for images, all within the same workspace)?

thomaseizinger:
> > If you'd like to use this image, it would be great if we can move it into this repository as part of the test. The linked discussion above has some ideas on how to do that but in a nutshell:
> > 
> > * Separate crate with build-script
> > * Introduce `dev-depenency` to said crate
> > * Build and tag docker images as part of the build-script
> > 
> > This allows us to use the images without having to push them anywhere and we can refer to them by name in our tests.
> 
> Apologies, as I'm still learning Rust. To do so, you're suggesting make a new crate _within_ this repository? Are you suggesting we organize the repository into workspaces (i.e., the testcontainers library crate, and a separate crate for images, all within the same workspace)?

Yes exactly!

The reasoning is:

- We want the docker images to be built with the tests and they need to be up-to-date, hence the use of a build-script
- We can't have the build-script as part of the original `testcontainers` crate because then it would get published to users as well but they don't need it

lukehsiao:
All the reasoning makes sense, and overall it sounds like an excellent idea to not be reliant on external images.

That will be a bit larger of an effort (I'll need to learn more about build scripts and whatnot), but I can give it a shot next time I have some more free time.

If you happen to have any pointers to docs or examples you think would be helpful and are willing to drop them here, it would likely help me get a jump start.

Any preference on the commit organization? E.g., should we do both tasks in a single commit, or perhaps I keep this first commit as is, and add a second commit to setup workspaces and remove the dependency on my hello-world image?

thomaseizinger:
> If you happen to have any pointers to docs or examples you think would be helpful and are willing to drop them here, it would likely help me get a jump start.

No direct example at hand but I'd expect the buildscript to be as simple as using https://doc.rust-lang.org/std/process/struct.Command.html to call the `docker build -f ./path/to/Dockerfile -t your-tag`. Plus cargo has a feature to rerun a script if a file changes (https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed) and we probably want to do that to make sure the docker image is up-to-date if we change the file.

> Any preference on the commit organization? E.g., should we do both tasks in a single commit, or perhaps I keep this first commit as is, and add a second commit to setup workspaces and remove the dependency on my hello-world image?

In an ideal world, we do the workspace and test-crate stuff first and then rebase this work on top of it by leveraging the existing infrastructure. But we can also just do it all in one commit :)

lukehsiao:
I've made the requested changes, please view the commit message themselves for more details.

In short, the first commit just sets up a workspace, with testcontainers being the only crate.

The second adds a new crate, `testimages` in the workspace, along with an example image and dockerfile.

The third commit implements the expose_ports, and uses the image from the `testimages` crate.

testcontainers/tests/images.rs
```diff
@@ -314,6 +314,29 @@ fn generic_image_with_custom_entrypoint() {
...
 
+#[test]
+fn generic_image_exposed_ports() {
+    let _ = pretty_env_logger::try_init();
+    let docker = clients::Cli::docker();
+
+    let target_port = 8080;
+
+    // This server does not EXPOSE ports in its image.
+    let generic_server = images::generic::GenericImage::new("expose_port", "latest")
```
thomaseizinger, line 325: Maybe the image should be called `no_expose_port`? :D

testcontainers/tests/images.rs
```diff
@@ -314,6 +314,29 @@ fn generic_image_with_custom_entrypoint() {
     );
 }
 
+#[test]
+fn generic_image_exposed_ports() {
```
thomaseizinger, line 318: What do you think of adding a test that fails because the port is _not_ exposed?

thomaseizinger (approved):
Thanks, this is amazing work! Very clean commits and implementation!

 I triggered CI, lets see.

testcontainers/tests/images.rs
lukehsiao: Great idea. I've made the change.

testcontainers/tests/images.rs
lukehsiao: Also a great idea, I've added a test.

lukehsiao:
I've addressed the style checks. 

I'm not sure what to do about this though: https://github.com/testcontainers/testcontainers-rs/runs/5882052991?check_suite_focus=true

lukehsiao:
I'd matched MSRV between testimages and testcontainers, which should fix the MSRV CI job.

I'm not sure why the postgres tests failed (they passed earlier, and should not unrelated to these changes).

lukehsiao:
Looks like MSRV is still failing. Do you have any pointers on how to fix this properly?

https://github.com/testcontainers/testcontainers-rs/runs/5957795062?check_suite_focus=true

thomaseizinger:
> On potential solution is updating rust version. Would it be reasonable to bump up to Rust 2021, rather than 2018, and/or a newer rust-version in testcontainers/Cargo.toml?

Yeah that is fine! We only track MSRV so we know what it is and can make an according semver bump. I consider MSRV part of the public API.

> Perhaps we should add a rust-toolchain file if we want to enforce some specific version, so that this is easier for contributors to test locally without waiting for a CI run? For example, it looks like CI is using rustc v1.46.0, while locally I'm on v1.60.0.

That doesn't quite solve the issue unfortunately. MSRV is important for users so the library only has to build (and work) with that. Any `cfg(test)` code for example is free to not adhere to MSRV :)

CI runs are typically fast. It is just that GitHub has this annoying feature where I need to manually approve every run because this is your first contribution 🙄

lukehsiao:
> Yeah that is fine! We only track MSRV so we know what it is and can make an according (i.e. breaking) semver bump. I consider MSRV part of the public API.

Sounds great. I'll add a patch for updating versions.

> CI runs are typically fast. It is just that GitHub has this annoying feature where I need to manually approve every run because this is your first contribution 🙄

Apologies for the overhead here, but thanks for you promptness in kicking off those jobs.

lukehsiao:
There we go! :)

thomaseizinger (approved):
Thank you for the effort and commitment to this!

bors r+

thomaseizinger:
I'll release in the upcoming days (hopefully). Need to update the changelog and have a skim over what was merged.

lukehsiao:
Hi @thomaseizinger, I'm sure you've just been busy, but we'd still love to move off a git dependency if you can have time to do a release :). 

Just wanted to bump that in case it has slipped through the cracks. Thanks!

PR #366, opened 2022-04-28
Add `watchdog` feature to allow cleanup on kill

### LINKED ISSUE #364: Add a singal handler to clean up test containers when tests are killed
Right now when a test run is cancled (CTRL-C, kill, etc) for whatever reason the containers are left dangling and require manual cleanup or will be hogging resources on the system.

In 0.12 we solved this by adding a signal handler to remove the containers something like this:

```
b(crate) struct SignalHandler {
    signal_handle: Handle,
    handle_task: Option<JoinHandle<()>>,
}

impl SignalHandler {
    pub(crate) fn new(container_id: String) -> Result<Self> {
        let mut signals = Signals::new(&[SIGTERM, SIGINT, SIGQUIT])?;
        let signal_handle = signals.handle();
        let handle_task = async_std::task::spawn(async move {
            let signal_docker = clients::Cli::default();
            while let Some(signal) = signals.next().await {
                signal_docker.stop(container_id.as_str());
                signal_docker.rm(container_id.as_str());
                let _ = signal_hook::low_level::emulate_default_handler(signal);
            }
        });
        Ok(Self {
            signal_handle,
            handle_task: Some(handle_task),
        })
    }
}
impl Drop for SignalHandler {
    fn drop(&mut self) {
        self.signal_handle.close();
        if let Some(s) = self.handle_task.take() {
            async_std::task::block_on(s.cancel());
        }
    }
}
```

The 0.13 API changes make this impossible since the `stop` and `rm` methods are now bound to a container so they can't be passed into a task for later cleanup.

This functionality seems to be something generally useful so it'd be lovely to have it as part of the framework itself, perhaps behind a feature flag that keeps a global registry of containers started by the framework and just uses one signal handler to close them all on shutdown?

This seems especially useful since rust does not guarantee Drop to be executed.

Licenser:
The watchdog will spawn a background thread that registers signal
handlers for SIGTERM, SIGINT, and SIGQUIT, if any such signal is
observed it will clean up currently running containers that are not
marked as `Keep`.

This fixes #364

### DIFF
--- a/testcontainers/src/core/container.rs
+++ b/testcontainers/src/core/container.rs
@@ -186,6 +186,8 @@ where
             Command::Keep => {}
             Command::Remove => self.rm(),
         }
+        #[cfg(feature = "watchdog")]
+        crate::watchdog::unregister(self.id());
     }
 }
 
--- a/testcontainers/src/lib.rs
+++ b/testcontainers/src/lib.rs
@@ -40,6 +40,9 @@ pub use crate::core::{Container, Image, ImageArgs, RunnableImage};
 #[cfg(feature = "experimental")]
 pub use crate::core::ContainerAsync;
 
+#[cfg(feature = "watchdog")]
+pub(crate) mod watchdog;
+
 /// All available Docker clients.
 pub mod clients;
 pub mod core;
--- a/testcontainers/src/watchdog.rs
+++ b/testcontainers/src/watchdog.rs
@@ -1,0 +1,52 @@
+use crate::{clients::Cli, core::Docker};
+use conquer_once::Lazy;
+use signal_hook::{
+    consts::{SIGINT, SIGQUIT, SIGTERM},
+    iterator::Signals,
+};
+use std::{collections::BTreeSet, sync::Mutex, thread};
+
+static WATCHDOG: Lazy<Mutex<Watchdog>> = Lazy::new(|| {
+    thread::spawn(move || {
+        let signal_docker = Cli::default();
+        let mut signals =
+            Signals::new(&[SIGTERM, SIGINT, SIGQUIT]).expect("failed to register signal handler");
+
+        for signal in &mut signals {
+            for container_id in WATCHDOG
+                .lock()
+                .map(|s| s.containers.clone())
+                .unwrap_or_default()
+            {
+                signal_docker.stop(&container_id);
+                signal_docker.rm(&container_id);
+            }
+
+            let _ = signal_hook::low_level::emulate_default_handler(signal);
+        }
+    });
+
+    Mutex::new(Watchdog::default())
+});
+
+#[derive(Default)]
+pub(crate) struct Watchdog {
+    containers: BTreeSet<String>,
+}
+
+/// Register a container for observation
+pub(crate) fn register(container_id: String) {
+    WATCHDOG
+        .lock()
+        .expect("failed to access watchdog")
+        .containers
+        .insert(container_id);
+}
+/// Unregisters a container for observation
+pub(crate) fn unregister(container_id: &str) {
+    WATCHDOG
+        .lock()
+        .expect("failed to access watchdog")
+        .containers
+        .remove(container_id);
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `watchdog` feature that spawns a background thread keeping track of docker containers that are started by the test suite and removes them in the case of a `CTRL+C` or `kill` of the test process.
+
 ## [0.13.0] - 2022-04-04
 
 ### Added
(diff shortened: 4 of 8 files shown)


testcontainers/src/lib.rs
```diff
@@ -40,6 +40,9 @@ pub use crate::core::{Container, Image, ImageArgs, RunnableImage};
 #[cfg(feature = "experimental")]
 pub use crate::core::ContainerAsync;
 
+#[cfg(feature = "watchdog")]
+pub mod watchdog;
```
thomaseizinger, line 44: Does it need to be public?

testcontainers/src/watchdog.rs
```diff
@@ -0,0 +1,55 @@
...
+    iterator::Signals,
+};
+use std::{collections::BTreeSet, sync::Mutex, thread};
+
+lazy_static! {
+    static ref WATCHDOG: Mutex<Watchdog> =
+        Mutex::new(Watchdog::new().expect("failed to register signal handler"));
+}
+
+use crate::{clients::Cli, core::Docker};
```
thomaseizinger, line 13: This use should go at the top with the others.

testcontainers/src/watchdog.rs
```diff
@@ -0,0 +1,55 @@
...
+use signal_hook::{
+    consts::{SIGINT, SIGQUIT, SIGTERM},
+    iterator::Signals,
+};
+use std::{collections::BTreeSet, sync::Mutex, thread};
+
+lazy_static! {
+    static ref WATCHDOG: Mutex<Watchdog> =
+        Mutex::new(Watchdog::new().expect("failed to register signal handler"));
+}
```
thomaseizinger, line 11: Personally, I am more a fan of the macro-free version: https://docs.rs/conquer-once/latest/conquer_once/type.Lazy.html

testcontainers/src/watchdog.rs
```diff
@@ -0,0 +1,55 @@
...
+                    .lock()
+                    .map(|s| s.containers.clone())
+                    .unwrap_or_default()
+                {
+                    signal_docker.stop(&container_id);
+                    signal_docker.rm(&container_id);
+                }
+                let _ = signal_hook::low_level::emulate_default_handler(signal);
+            }
+        });
```
thomaseizinger, line 34: Using `conquer_once::Lazy` would allow us to spawn this outside of the constructor.

testcontainers/src/watchdog.rs
```diff
@@ -0,0 +1,55 @@
...
+            .insert(container_id);
+    }
+    /// Unregisters a container for observation
+    pub(crate) fn unregister(container_id: &str) {
+        WATCHDOG
+            .lock()
+            .expect("failed to access watchdog")
+            .containers
+            .remove(container_id);
+    }
```
thomaseizinger, line 54: These don't need to be associated, do they? I think free functions would make it more obvious that we are mutating global state.

thomaseizinger (commented):
Thanks!

Did you have a look at how we could support the experimental async client as well?
I'd like these two clients to retain feature parity as much as possible. It is only marked as experimental because I wanted to get a release out and didn't have sufficient time to test it thoroughly.

testcontainers/src/lib.rs
Licenser: nope good catch :D

testcontainers/src/watchdog.rs
Licenser: Good call :) updated it and pulled it into the lazy function!

testcontainers/src/watchdog.rs
Licenser: Very good point!

Licenser:
Heya I didn't look at the async client originally, but added it in with incorporating the feedback :)

testcontainers/src/core/container.rs
```diff
@@ -186,6 +186,8 @@ where
             Command::Keep => {}
             Command::Remove => self.rm(),
         }
+        #[cfg(feature = "watchdog")]
+        crate::watchdog::unregister(self.id());
```
thomaseizinger, line 190: I think `ContainerAsync` will need the same thing?

testcontainers/src/watchdog.rs
```diff
@@ -0,0 +1,50 @@
...
+                .map(|s| s.containers.clone())
+                .unwrap_or_default()
+            {
+                signal_docker.stop(&container_id);
+                signal_docker.rm(&container_id);
+            }
+            let _ = signal_hook::low_level::emulate_default_handler(signal);
+        }
+    });
+    watchdog
```
thomaseizinger, line 27: I think this can be slightly re-ordered to be a bit more readable:

```suggestion
    thread::spawn(move || {
        let signal_docker = Cli::default();
	let mut signals = Signals::new(&[SIGTERM, SIGINT, SIGQUIT]).expect("failed to register signal handler");
    
        for signal in &mut signals {
            for container_id in WATCHDOG
                .lock()
                .map(|s| s.containers.clone())
                .unwrap_or_default()
            {
                signal_docker.stop(&container_id);
                signal_docker.rm(&container_id);
            }

            let _ = signal_hook::low_level::emulate_default_handler(signal);
        }
    });
    
    Mutex::new(Watchdog::default())
```

thomaseizinger (changes requested):
Thanks!

Two minor suggestions and then this can be merged :)

Licenser:
Updated from suggestioned and squashed as per CONTRIBUTING.md :)

CleverAkanoa:
May ask you when this neat feature will be added to a release ?

Great work :D

PR #368, opened 2022-05-05
Handle IPv4 and IPv6 host ports separately

### LINKED ISSUE #365: Container::get_host_port() and IP addresses
I ran into an issue recently when trying to connect to a container's mapped port. The root cause was that Docker may assign different host port numbers for a published port on the host's IPv4 and IPv6 addresses. If I used "localhost" with a port number returned from `get_host_port()` I would sometimes get connections to the wrong container, due to mixing and matching an IPv6 address with a port number from Docker's IPv4 port binding. I see that `impl From<bollard_stubs::models::PortMap> for Ports` takes the first port binding from the list it receives from Docker.

To resolve this issue, I think that `get_host_port()` should either make a guarantee that it will return a port number bound on 0.0.0.0, or a breaking change should be made so it either takes an IP address (or choice of IPv4/IPv6) as an argument, or returns one along with the port number.

divergentdave:
This introduces new methods for getting host ports mapped from containers, specific to IPv4 or IPv6 addresses. The existing methods are deprecated in favor of their replacements.

When Docker publishes all of a container's exposed ports, it does so on both IPv4 and IPv6. Port numbers are chosen separately for each, from the ephemeral port range, though it is common for Docker to assign the same host port number for both the IPv4 and IPv6 bindings of one exposed port. However, if Docker encounters a port that is in use, it may choose
another.

Previously, get_host_port() would return the first port binding it found in Docker's list for each container port. The new methods filter the list by IP protocol version. As Docker tended to return port bindings for 0.0.0.0 first, the now-deprecated get_host_port() forwards to get_host_port_ipv4().

Closes #365.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `watchdog` feature that spawns a background thread keeping track of docker containers that are started by the test suite and removes them in the case of a `CTRL+C` or `kill` of the test process.
-- Introduced `Container::get_host_port_ipv4` `Container::get_host_port_ipv6`, `ContainerState::host_port_ipv4` and `ContainerState::host_port_ipv6` to better handle automatically assigned ports.
+- Introduced `Container::get_host_port_ipv4`, `Container::get_host_port_ipv6`, `ContainerState::host_port_ipv4`, and `ContainerState::host_port_ipv6` to better handle automatically assigned ports.
   Docker may bind the same exposed container port to different host ports on `0.0.0.0` and `::`, depending on influences from the environment.
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `watchdog` feature that spawns a background thread keeping track of docker containers that are started by the test suite and removes them in the case of a `CTRL+C` or `kill` of the test process.
 - Introduced `Container::get_host_port_ipv4` `Container::get_host_port_ipv6`,
   `ContainerState::host_port_ipv4` and `ContainerState::host_port_ipv6` to better handle
   automatically assigned ports. Docker may bind the same exposed container port to different
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Introduced `Container::get_host_port_ipv4` `Container::get_host_port_ipv6`,
+  `ContainerState::host_port_ipv4` and `ContainerState::host_port_ipv6` to better handle
+  automatically assigned ports. Docker may bind the same exposed container port to different
+  host ports on `0.0.0.0` and `::`, depending on influences from the environment.
+
+### Changed
+
+- `Container::get_host_port` and `ContainerState::host_port` are now deprecated in favor of the new
+  IPv4- and IPv6-specific methods.
+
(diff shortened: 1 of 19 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Introduced `Container::get_host_port_ipv4` `Container::get_host_port_ipv6`,
+  `ContainerState::host_port_ipv4` and `ContainerState::host_port_ipv6` to better handle
```
thomaseizinger, line 13: `ContainerState` is not public so we don't need to mention it here :)

thomaseizinger: Ah damn, I think it is public.

Can we make it `pub(crate)` as part of this PR? I don't think this should be public API!

testcontainers/src/core/ports.rs
```diff
@@ -30,34 +31,75 @@ impl Ports {
...
-    /// Returns the host port for the given internal port.
+    /// Returns the host port for the given internal port, on the host's IPv4 interfaces.
+    #[deprecated(
+        since = "0.13.1",
+        note = "Use `map_to_host_port_ipv4()` or `map_to_host_port_ipv6()` instead."
+    )]
     pub fn map_to_host_port(&self, internal_port: u16) -> Option<u16> {
-        self.mapping.get(&internal_port).cloned()
+        self.map_to_host_port_ipv4(internal_port)
+    }
```
thomaseizinger, line 41: `Ports` is definitely not public so no need to deprecate it!

testcontainers/tests/dual_stack_host_ports.rs
```diff
@@ -0,0 +1,73 @@
...
+
+/// Test the functionality of exposing container ports over both IPv4 and IPv6.
+#[tokio::test]
+async fn test_ipv4_ipv6_host_ports() {
+    let _ = pretty_env_logger::try_init();
+    let docker = clients::Cli::default();
+
+    let wait_for = WaitFor::message_on_stdout("server is ready");
+    let image =
+        GenericImage::new("tumdum/simple_web_server", "latest").with_wait_for(wait_for.clone());
```
thomaseizinger, line 13: Where is this coming from? I would like to either depend on images that are known to be maintained or if you need a specific one for your usecase, you can add one in the `testimages` crate!

thomaseizinger (commented):
Thanks!

Did an initial pass :)

testcontainers/tests/dual_stack_host_ports.rs
divergentdave: I copied this out of the `generic_image_with_custom_entrypoint` test. I can take a crack at re-implementing it in `testimages`.

CHANGELOG.md
thomaseizinger: Ah I remember now, this was added as part of `exec_after_start`!

It is fine as is actually, I guess this has to be part of the public API then. Sorry about the wild-goose chase!

testcontainers/tests/dual_stack_host_ports.rs
thomaseizinger: Right, I do remember this now as well 😅

I would much appreciate if we could move this image into `testimages` 😊

testcontainers/src/core/image.rs
```diff
@@ -111,9 +111,9 @@ impl ContainerState {
         Self { ports }
     }
 
-    pub fn host_port(&self, internal_port: u16) -> u16 {
+    pub(crate) fn host_port_ipv4(&self, internal_port: u16) -> u16 {
```
thomaseizinger, line 114: This has to stay public actually, otherwise the whole point of exposing it makes no sense :)

Sorry about the confusion!

thomaseizinger (changes requested):

thomaseizinger (approved):
Sorry for the delay and thanks a lot for the contribution! :)

PR #371, opened 2022-05-24
Link Matrix badge to room

lukehsiao:
This way it's easy to click through to the room itself.


thomaseizinger (approved):
Thanks!

bors r+

PR #372, opened 2022-05-25
[Feature] Add tty option

asaphaaning:
Thanks for branching this wonderful project out into the Rust space!

Some images rely on a tty being available to properly function. This should cover it!

### DIFF
--- a/testcontainers/src/core/image.rs
+++ b/testcontainers/src/core/image.rs
@@ -93,6 +93,11 @@ where
     fn exec_after_start(&self, cs: ContainerState) -> Vec<ExecCommand> {
         Default::default()
     }
+
+    /// Whether or not a pseudo-tty should be allocated within the container.
+    fn tty(&self) -> bool {
+        true
+    }
 }
 
 #[derive(Default, Debug)]
@@ -173,6 +178,10 @@ impl<I: Image> RunnableImage<I> {
 
     pub fn args(&self) -> &I::Args {
         &self.image_args
+    }
+
+    pub fn tty(&self) -> bool {
+        self.image.tty()
     }
 
     pub fn network(&self) -> &Option<String> {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `watchdog` feature that spawns a background thread keeping track of docker containers that are started by the test suite and removes them in the case of a `CTRL+C` or `kill` of the test process.
 - Introduced `Container::get_host_port_ipv4`, `Container::get_host_port_ipv6`, `ContainerState::host_port_ipv4`, and `ContainerState::host_port_ipv6` to better handle automatically assigned ports.
   Docker may bind the same exposed container port to different host ports on `0.0.0.0` and `::`, depending on influences from the environment.
+- Added a `tty` option to the `Image` trait defaulting to true. This option allocates a pseudo-tty inside the docker container.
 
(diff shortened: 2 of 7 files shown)


mfelsche:
Can you provide an example docker image that requires docker to have a pseudo-tty configured? So this can be verified. I am not 100% sure this works with stdout piped https://github.com/testcontainers/testcontainers-rs/blob/dev/testcontainers/src/clients/cli.rs#L188 this is why i ask

ASAPHAANING:
I can't quite reproduce the image (some php + openrc stuff) I was working with but essentially it looked like the snippet below, symlinking from the tty over to the log locations and expecting `/dev/pts/0` to be writable. 

_Example_
```
FROM nginx:stable-alpine

COPY init.sh /

RUN ln -nsf /dev/pts/1 /var/log/nginx/access.log

ENTRYPOINT ["/init.sh"]
```

init.sh
```
chmod o+w /dev/pts/1
```

I tested my changes with the image and confirmed that running with `-it` allows the container to boot.

thomaseizinger:
Thanks! Maintaining a minimal API is something that is really important, hence a couple of questions:

- Would it hurt to always try and allocate a tty and not expose this at all?
- The option seems to be image specific so I think we can remove the override on `RunnableImage`?
- It would be nice to add a test-case with an actual docker image for such purposes. You can add them [here](https://github.com/testcontainers/testcontainers-rs/tree/dev/testimages) so they are available within the test-suite!

ASAPHAANING:
Hi @thomaseizinger!

Thanks for putting this thing together!

I actually do agree that having ptty available is a sensible default. If I flip the bit in the trait default implementation the implementing party will not be required to override it while leaving the the option there if they specifically want to _not_ allocate a ptty. What do you think?

I added a test case too as per request 👍

testcontainers/src/core/image.rs
```diff
@@ -157,6 +162,7 @@ impl ImageArgs for () {
 pub struct RunnableImage<I: Image> {
     image: I,
     image_args: I::Args,
+    tty: bool,
```
thomaseizinger, line 165: Is this necessary? Can't we access it through `image`?

thomaseizinger (changes requested):
Thank you! I think exposing this makes sense :)

One minor thing.

testcontainers/src/core/image.rs
ASAPHAANING: Ah, eagle eyes! 🦅  An artifact left from the other review change. It's gone :)

ASAPHAANING:
@thomaseizinger 

Anything else you'd like me to cover? :)

thomaseizinger (approved):
Yep this looks good. Feel free to `bors r+` at your convenience.

bors d=ASAPHAANING

thomaseizinger:
bors retry

ASAPHAANING:
> bors retry

I think the `test_ipv4_ipv6_host_ports` test is broken on `dev`. I can make the entire suite run if I comment out the last assertion, let me push an example

thomaseizinger:
> > bors retry
> 
> I think the `test_ipv4_ipv6_host_ports` test is broken on `dev`. I can make the entire suite run if I comment out the ipv6 assertions.

Damn. Thanks for investigating.

ASAPHAANING:
bors retry

DDtKey:
We've released a new version of `testcontainers`: [0.16.0](https://github.com/testcontainers/testcontainers-rs/releases/tag/0.16.0).

Could you update and re-open the PR if it's still relevant. 

Closing because of inactivity for now

PR #383, opened 2022-06-29
Allow extra parameters to `docker run`

### LINKED ISSUE #380: Run container in privileged mode
Hello

I'm using `testcontainers` for integration testing of services running in LXC containers, which means I need LXC running inside Docker, and for this reason, I need to run docker with the `--privileged` flag.

Do you think this could be implemented, or maybe accept a PR implementing it?

andrenth:
Fixes #380


thomaseizinger (changes requested):
Thank you!

Such flags should be configured via `RunnableImage`. You can add a flag `priviledged` there, then users can configure it if needed for a particular `run` of an image.

andrenth:
Hi

I've force-pushed an implementation using a `privileged` field in `RunnableImage`.

Thanks for looking.

thomaseizinger (approved):
Thanks, this looks good!

bors d=andrenth

PR #384, opened 2022-07-06
Make Docker Send and Sync

### LINKED ISSUE #354: Singleton instace of a container
Currentlty i am runig multiple e2e test that make use of some docker images, the problem is that each test is starting his own containers so i ask if there is a way in the curret api to share the containers

### LINKED ISSUE #381: Container and Cli are no longer "Send"
We have been working with v0.12.0 and the newer releases have substantial refactors. One that affects us is that we can no longer hold global statics for spawned Containers since the Container object no longer impls Send. Was this a conscious decision and does it really need to be that way. We would like to use your latest release but were not expecting to need to refactor our own test infra.

ufoscout:
This fixes #354 allowing singleton instances of a container.
Fixes  #381 also


testcontainers/tests/container.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[test]
+fn container_should_be_sync() {
+    let docker = Cli::default();
+    let node = docker.run(HelloWorld {});
+    need_sync(node);
+}
+
+fn need_send<T: Send>(_t: T) {}
+fn need_sync<T: Sync>(_t: T) {}
```
thomaseizinger, line 18: We can combine this into one and make the test cheaper by just passing it as a type parameter using turbo fish:

```rust
assert_send_and_sync::<Container<...>>();
```

thomaseizinger (changes requested):
Thanks!

Concept ACK, just one suggestion on how we can write the test.

testcontainers/tests/container.rs
ufoscout: Good point, done!

thomaseizinger:
I've pushed https://github.com/testcontainers/testcontainers-rs/pull/385. Need to wait for it to merge before we can merge this.

ufoscout:
> I've pushed #385. Need to wait for it to merge before we can merge this.

Now that #385 was merged, this one could be merged too.

PR #394, opened 2022-09-09
Allow setting shared memory size

### LINKED ISSUE #395: Allow setting shared memory size
Allow users to set the shared memory size (`--shm-size=N`) when creating a runnable container image.

An example for a container where this is useful would be [solace/solace-pubsub-standard](https://hub.docker.com/r/solace/solace-pubsub-standard), which requires at least 1GB of shared memory to be able to run (see [Solace Docs > Resource Requirements for Container Images](https://docs.solace.com/Software-Broker/System-Resource-Requirements.htm#res-req-container)). By default, the `docker run` command sets the shared memory size to 64MB (see [option `--shm-size`](https://docs.docker.com/engine/reference/run/#runtime-constraints-on-resources)).

Analog feature in testcontainers for Java: https://github.com/testcontainers/testcontainers-java/issues/952.

pbillaut:
Allow setting shared memory size (`--shm-size`) in bytes when creating an image.

Fixes #395.

### DIFF
--- a/testcontainers/src/clients/cli.rs
+++ b/testcontainers/src/clients/cli.rs
@@ -147,6 +147,10 @@ impl Client {
             command.arg("--privileged");
         }
 
+        if let Some(bytes) = image.shm_size() {
+            command.arg(format!("--shm-size={}", bytes));
+        }
+
         if let Some(network) = image.network() {
             command.arg(format!("--network={}", network));
         }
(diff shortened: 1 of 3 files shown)


thomaseizinger (changes requested):
Thank you!

This needs to be supported in the HTTP client too before we can merge it.

pbillaut:
> Thank you!
> 
> This needs to be supported in the HTTP client too before we can merge it.

Thanks for pointing that out. Implemented the missing functionality in 681d643.

When trying to run the tests I hit #397 though, so that PR would have to be merged first.

thomaseizinger (approved):
Thank you!

Before merging please:

- Squash all commits into one
- Rebase onto master once https://github.com/testcontainers/testcontainers-rs/pull/398 is merged

bors d=pbillaut

PR #403, opened 2022-09-15
CI with `--all-features` and `--no-default-features`

### LINKED ISSUE #399: Modify CI workflow to run with a matrix that checks `--all-features` and `--no-default-features`

pbillaut:
Fix #399.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,14 @@ jobs:
           - test --all-features
           - test --no-default-features
     steps:
+      - name: Enable Docker Remote API on Localhost
+        shell: bash
+        run: |
+          sudo sed -i \
+            's#ExecStart=/usr/bin/dockerd#ExecStart=/usr/bin/dockerd -H tcp://127.0.0.1:2375#g' \
+            /lib/systemd/system/docker.service
+          sudo systemctl daemon-reload
+          sudo systemctl restart docker
       - uses: actions/checkout@v3
       - uses: actions-rs/toolchain@v1
         id: toolchain
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,11 +1,20 @@
 name: Continuous Integration
+
 on:
   pull_request:
   push:
     branches: [master, dev, trying, staging]
+
 jobs:
   msrv:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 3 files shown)


thomaseizinger (commented):
Thanks!

This is great. I am not sure why the tests are failing though. Have they been broken all this time?

To merge this, we will also need to extend the jobs in `bors.toml`, otherwise bors will listen for jobs that are never going to run.

pbillaut:
> Thanks!
> 
> This is great. I am not sure why the tests are failing though. Have they been broken all this time?
> 
> To merge this, we will also need to extend the jobs in `bors.toml`, otherwise bors will listen for jobs that are never going to run.

Hey, the tests are failing since the `Http` client needs the Docker daemon to run with the Remote API Access enabled. This can for example be achieved by starting the Docker daemon like this:
```bash
dockerd --host fd:// --containerd=/run/containerd/containerd.sock --host tcp://127.0.0.1:2375
```

By default, and thus on the GitHub worker, Remote API Access is not enabled and that's why we're seeing _connection refused_ errors. I'm currently experimenting with running [Docker-in-Docker](https://hub.docker.com/_/docker) to provide the Remote API Access on the worker via the dockerized Docker daemon.

pbillaut:
Instead of running another Docker container that provides the Docker Remote API, it's simpler to just enable it via systemd, of course only exposed on localhost.

.github/workflows/ci.yml
```diff
@@ -20,10 +28,25 @@ jobs:
...
+        cargo-cmd:
+          - test --all-features
+          - test --no-default-features
     steps:
+      - name: Enable Docker Remote API on Localhost
+        shell: bash
+        run: |
+          sudo sed -i \
+            's#ExecStart=/usr/bin/dockerd#ExecStart=/usr/bin/dockerd -H tcp://127.0.0.1:2375#g' \
+            /lib/systemd/system/docker.service
```
thomaseizinger, line 47: I guess this is good enough here. We could also consider [installing an override file](https://www.digitalocean.com/community/tutorials/how-to-use-systemctl-to-manage-systemd-services-and-units#editing-unit-files).

What do you think?

pbillaut: I have no real preference here - adapted pipeline accordingly in 5921126.

thomaseizinger (approved):
Thank you for this and sorry for the delay in getting back to you!

bors r+

PR #406, opened 2022-09-30
Expose a ctor for Http client to accept custom bollard Docker client.

bobmcwhirter:
I know it's an edge-case, but OSX folks using Podman  4.x can't currently use the CLI, due to `inspect` differences.

Using the HTTP API works perfectly (and is async, yay).  But it also is a non-default usage of Bollard.

By providing a way to customize the Bollard Docker client being used, this patch allows me to use a random, non-standard, OSX/m1->QEMU->podman stack with success.

### DIFF
--- a/testcontainers/src/clients/http.rs
+++ b/testcontainers/src/clients/http.rs
@@ -41,7 +41,7 @@ impl fmt::Debug for Http {
 
 impl Default for Http {
     fn default() -> Self {
-        Self::new()
+        Self::new(Docker::connect_with_http_defaults().unwrap())
     }
 }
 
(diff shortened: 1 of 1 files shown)


thomaseizinger (commented):
I am hesitant to merge this because it leaks a dependency into our public API which makes it more difficult for testcontainers to eventually reach 1.0 if we want to provide some guarantees on for how long we won't be making breaking changes.

What customization of the bollard client do you need? I am open to adding configuration options, as long as they are client agnostic and not specific to bollard's implementation.

thomaseizinger:
> I know it's an edge-case, but OSX folks using Podman 4.x can't currently use the CLI, due to `inspect` differences.

What does "currently" mean? Is this expected to be solved eventually in bollard or podman?

thomaseizinger:
Thanks for the PR but we decided to remove support for podman: https://github.com/testcontainers/testcontainers-rs/issues/422.

PR #408, opened 2022-10-07
Add container function to get the gateway ip address

jltorresm:
The container's gateway IP address is necessary for setups where your `testcontainer` runs inside another Docker container. Unfortunately, the bridge IP address doesn't coincide with the gateway address.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.15.0] - 2023-09-28
 
 ### Added
+
+- Added `privileged` parameter to `RunnableImage`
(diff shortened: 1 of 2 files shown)


ns-sjorgedeaguiar:
Looking forward to have this merged. Alternatively the `inspect` method could be made public for more flexibility.

thomaseizinger:
> Looking forward to have this merged. Alternatively the `inspect` method could be made public for more flexibility.

This is unlikely going to happen as it would be a big commitment to the public API of `testcontainers` which I am reluctant to make.

thomaseizinger (changes requested):
Thanks and sorry for getting back to this so late.

Given that this PR extends our API with a public function, I'd like to understand in detail, how and when this function is useful. (See "Simple" section in our upcoming "Design Principles" documentation: https://github.com/testcontainers/testcontainers-rs/pull/420).

Can you please extend the PR description with some more documentation that explains, how and when bridge and gateway IP differ?

- Is the bridge IP generally not useful if we are within another docker container?
- Is there a gateway IP if we are not running inside another docker container?
- Can we somehow merge these two functions in a way to give back one IP address that always works?

From a formality PoV, this PR is missing:

- A changelog entry
- An adaption of the `ContainerAsync` API

PR #410, opened 2022-10-11
fix: `get_bridge_ip_address`

DDtKey:
Currently `get_bridge_ip_address` uses `NetworkSettings.IPAddress` which is [deprecated](https://docs.rs/bollard/latest/bollard/models/struct.NetworkSettings.html#structfield.ip_address)
So this PR uses `NetworkSettings.Networks` for that and perfectly combines with `network` (if it's passed to `Image)

~~Moreover in some cases it useful to use `inspect` result directly (it was mentioned [here](https://github.com/testcontainers/testcontainers-rs/pull/408#issuecomment-1272628962) as well)~~

### DIFF
--- a/testcontainers/src/core/container.rs
+++ b/testcontainers/src/core/container.rs
@@ -168,18 +168,32 @@ where
             })
     }
 
-    /// Returns the bridge ip address of docker container as specified in NetworkSettings.IPAddress
+    /// Returns the bridge ip address of docker container as specified in NetworkSettings.Networks.IPAddress
     pub fn get_bridge_ip_address(&self) -> IpAddr {
-        IpAddr::from_str(
-            &self
-                .docker_client
-                .inspect(&self.id)
-                .network_settings
-                .unwrap_or_default()
-                .ip_address
-                .unwrap_or_default(),
-        )
-        .unwrap_or_else(|_| panic!("container {} has missing or invalid bridge IP", self.id))
+        let settings = self
+            .docker_client
+            .inspect(&self.id)
+            .network_settings
+            .unwrap_or_else(|| panic!("container {} has no network settings", self.id));
+
+        let mut networks = settings
+            .networks
+            .unwrap_or_else(|| panic!("container {} has no any networks", self.id));
+
+        let bridge_name = self
+            .image
+            .network()
+            .clone()
+            .or(settings.bridge)
+            .unwrap_or_else(|| panic!("container {} has missing bridge name", self.id));
+
+        let ip = networks
+            .remove(&bridge_name)
+            .and_then(|network| network.ip_address)
+            .unwrap_or_else(|| panic!("container {} has missing bridge IP", self.id));
+
+        IpAddr::from_str(&ip)
+            .unwrap_or_else(|_| panic!("container {} has invalid bridge IP", self.id))
     }
 
     pub fn exec(&self, cmd: ExecCommand) {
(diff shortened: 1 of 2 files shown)


testcontainers/src/core/container.rs
```diff
@@ -168,20 +168,30 @@ where
...
+    /// Returns the bridge ip address of docker container as specified in NetworkSettings.Networks.IPAddress
     pub fn get_bridge_ip_address(&self) -> IpAddr {
         IpAddr::from_str(
             &self
                 .docker_client
                 .inspect(&self.id)
                 .network_settings
-                .unwrap_or_default()
-                .ip_address
+                .and_then(|settings| {
```
thomaseizinger, line 178: Instead of continuing to nest here, please panic early. Makes the control flow much easier to understand!

testcontainers/src/core/container.rs
```diff
@@ -168,20 +168,30 @@ where
...
+                .and_then(|net| net.ip_address)
                 .unwrap_or_default(),
         )
         .unwrap_or_else(|_| panic!("container {} has missing or invalid bridge IP", self.id))
     }
 
+    /// Returns the inspect result of docker container
+    pub fn inspect(&self) -> ContainerInspectResponse {
+        self.docker_client.inspect(&self.id)
+    }
```
thomaseizinger, line 193: Please remove this part. We are not going to add `ContainerInspectResponse` to our public API.

testcontainers/src/core/container_async.rs
```diff
@@ -117,23 +117,28 @@ where
...
-                        .unwrap_or_default(),
-                )
-                .unwrap_or_else(|_| {
-                    panic!("container {} has missing or invalid bridge IP", self.id)
+        let result = self.docker_client.inspect(&self.id).await;
+
+        IpAddr::from_str(
+            &result
+                .network_settings
+                .and_then(|settings| {
```
thomaseizinger, line 127: Same as for the blocking version.

thomaseizinger (changes requested):
Thanks and sorry for the delay!

Please see the inline comments.

testcontainers/src/core/container.rs
DDtKey: panic? here was `unwrap_or_defaul` so I saved this behavior and `and_then` won't apply anything for `Err` value (so even if we would use panic, it also has fail-fast semantic). It's idiomatic rust code. The only reason I can see - panic with different messages on different levels

It isn't easier to understand to me, but I definitely can change, so I will use rules in this repo

testcontainers/src/core/container.rs
DDtKey: Ok, no-problem. I re-exported it cause it was asked in another PR.

testcontainers/src/core/container.rs
DDtKey: @thomaseizinger I refactored it a bit with more clear flow & error messages, pls take a look

DDtKey:
@thomaseizinger any updates on this? I think I addressed all comments above

DDtKey:
@thomaseizinger gentle ping 🙂

thomaseizinger (approved):
LGTM

Sorry for the long delay.

DDtKey:
Should I fix linter messages? looks not related to this PR 🤔

thomaseizinger:
> Should I fix linter messages? looks not related to this PR thinking

They are unrelated yes. I haven't merged PRs in a while so it didn't come up before and likely, new clippy lints have been released a while ago.

DDtKey:
Rebased & squashed

PR #417, opened 2022-10-19
Provide docker remote api with docker-in-docker

pbillaut:
This is an alternative approach to #403, running the test suite against a Docker Remote API in CI using Docker-in-Docker.

Unfortunately this approach leads to some tests failing.


thomaseizinger:
Thanks for investigating an alternative!

I already sent #403 into the merge queue so I think we no longer need this. But out of curiosity, what made you explore an alternative solution?

PR #419, opened 2022-10-21
Feature: Add support for mysql

### LINKED ISSUE #436: Add support for MySQL
MySQL is a very popular flavour of SQL which is used extensively across the industry.

I put together the code to support MySQL images here: #419 

I've followed the contributing guidelines but please let me know if there's anything you'd like done differently etc

tmay-sarsaparilla:
This change adds support for `mysql` images, both `8.0` and `5.7` versions:

- Add: `MySql` struct with default and `Image` implementation
- Add: integration tests for default image, custom port mapping, and custom tag

Tests make use of the `mysql` crate: https://crates.io/crates/mysql
Images from the official `mysql` docker hub: https://hub.docker.com/_/mysql

Closes #436


thomaseizinger:

PR #429, opened 2022-11-03
feat: expose spanner REST API port

plaflamme:
Also bump the `google/cloud-sdk` image tag for good hygiene.


thomaseizinger:

PR #431, opened 2022-11-04
Log error when starting a container fails

stepantubanov:
When container fails to start all you get is `failed to start container` message, and it's not always easy to guess why it failed. This PR includes command output in panic message.

Example:

`dev`:
> ```thread '<unnamed>' panicked at 'failed to start container'```

PR:
> ```thread '<unnamed>' panicked at 'failed to start container, output = Output { status:  ExitStatus(unix_wait_status(32000)), stdout: "e5866b9407496bc9a88b63528370de91ec43358a6a40e1262fb3cae9dd3a8ec1\n", stderr: "docker: Error response from daemon: driver failed programming external connectivity on endpoint dazzling_elbakyan (85503e92065801ce10c86a95ba9af00c53e0e9a3ec4eb0ad78e55e306fbf8434): Bind for 0.0.0.0:13306 failed: port is already allocated.\n" }'```

### DIFF
--- a/testcontainers/src/clients/cli.rs
+++ b/testcontainers/src/clients/cli.rs
@@ -44,8 +44,13 @@ impl Cli {
         log::debug!("Executing command: {:?}", command);
 
         let output = command.output().expect("Failed to execute docker command");
+        if !output.status.success() {
+            let stdout = std::str::from_utf8(&output.stdout).unwrap_or("{not utf8}");
+            let stderr = std::str::from_utf8(&output.stderr).unwrap_or("{not utf8}");
+            log::error!("Failed to start container.\nContainer stdout: {stdout}\nContainer stderr: {stderr}");
+            panic!("Failed to start container, check log for details")
+        }
 
-        assert!(output.status.success(), "failed to start container");
         let container_id = String::from_utf8(output.stdout)
             .expect("output is not valid utf8")
             .trim()


stepantubanov:
@thomaseizinger 
I was actually trying to see the message right away, without having to download & read the logs (in CI for example). What do you think about something like this (PR updated)?

```rust
if !output.status.success() {
    panic!(
        "failed to start container, client stderr: {}",
        std::str::from_utf8(&output.stderr).expect("stderr contains invalid utf-8")
    );
}
```

thomaseizinger:
Logs and panic are both printed to stderr in most cases, no?

I don't feel strongly though.

Why did you exclude stdout?

stepantubanov:
> Logs and panic are both printed to stderr in most cases, no?

No, in our case for example, we have logs written to a file (and we use cucumber, which does it's own printing, so printing logs to stdout/err isn't an option).

> Why did you exclude stdout?

You mean from panic message? Just thought stderr is more suitable for panic message, but I don't mind including stdout if you think it makes sense. Or we could log container's stdout.

thomaseizinger:
> > Logs and panic are both printed to stderr in most cases, no?
> 
> No, in our case for example, we have logs written to a file (and we use cucumber, which does it's own printing, so printing logs to stdout/err isn't an option).

That is arguably a shortcoming of your test configuration. In a typical Rust test setup, logs are printed to stderr using `env_logger` or similar and hidden in case the test is successful.

I am a bit hesitant to include entire outputs in the panic message.

Also, what just occurred to me: How should this behave for outputs that are extremely large? Perhaps we should do something like docker compose where we redirect all output of the container to our output in a streaming fashion?

I understand the need for making debugging this easy and it should be easy, but we need a solution that considers edge cases too and will work well for the wider Rust community.

stepantubanov:
> That is arguably a shortcoming of your test configuration. In a typical Rust test setup, logs are printed to stderr using env_logger or similar and hidden in case the test is successful.

Sure, fair point. In our setup we do have to look at the logs when other errors occur too.

Although I do think that the log should have an `error` level. And if possible in one multi-line message (to appear as a single trace event when tracing is used). Something like this:
```rust
if !output.status.success() {
	log::error!("failed to start container.\n  container stdout: {}\n  container stderr: {}\n");
	panic!("failed to start container, check logs for details")
}
```

thomaseizinger:
> > That is arguably a shortcoming of your test configuration. In a typical Rust test setup, logs are printed to stderr using env_logger or similar and hidden in case the test is successful.
> 
> Sure, fair point. In our setup we do have to look at the logs when other errors occur too.
> 
> Although I do think that the log should have an `error` level. And if possible in one multi-line message (to appear as a single trace event when tracing is used). Something like this:
> 
> ```rust
> if !output.status.success() {
> 	log::error!("failed to start container.\n  container stdout: {}\n  container stderr: {}\n");
> 	panic!("failed to start container, check logs for details")
> }
> ```

That would work for me!

stepantubanov:
@thomaseizinger Could you please take a look?

thomaseizinger (approved):

DDtKey (approved):

thomaseizinger:
Can you update to latest master? I think that is why CI is not running.

DDtKey (approved):

PR #437, opened 2022-11-25
feat!: make `Container::exec` synchronous and return output

### LINKED ISSUE #344: Support exec in container
It would be great to have an equivalent to `execInContainer` as described here: https://www.testcontainers.org/features/commands/.

Not only with the ability to execute a command in a container, but also to obtain the stdout/stderr of that execution.

DDtKey:
Closes #344

Remove of `-d` argument for `Container::exec` to ensure command has finished.
It's extremely useful when you need to ensure that command was executed as expected, but container logs can't tell this.

It also now returns output of command.

Any comments are welcome, that's quite straightforward way to implement it.

### DIFF
--- a/testcontainers/src/clients/cli.rs
+++ b/testcontainers/src/clients/cli.rs
@@ -6,7 +6,7 @@ use bollard_stubs::models::{ContainerInspectResponse, HealthStatusEnum};
 use std::{
     collections::HashMap,
     ffi::{OsStr, OsString},
-    process::{Command, Stdio},
+    process::{Child, Command, Stdio},
     sync::{Arc, RwLock},
     thread::sleep,
     time::{Duration, Instant},
@@ -390,20 +390,23 @@ impl Docker for Cli {
             .expect("Failed to start docker container");
     }
 
-    fn exec(&self, id: &str, cmd: String) {
-        self.inner
+    fn exec(&self, id: &str, cmd: String) -> std::process::Output {
+        let exec_output = self
+            .inner
             .command()
             .arg("exec")
-            .arg("-d")
             .arg(id)
             .arg("sh")
             .arg("-c")
-            .arg(cmd)
+            .arg(&cmd)
             .stdout(Stdio::piped())
             .spawn()
-            .expect("Failed to execute docker command")
-            .wait()
-            .expect("Failed to exec in a docker container");
+            .and_then(Child::wait_with_output)
+            .expect("Failed to execute docker command");
+
+        log::debug!("command {} was executed!", cmd);
+
+        exec_output
     }
 
     fn block_until_ready(&self, id: &str, ready_conditions: Vec<WaitFor>) {
--- a/testcontainers/src/core/container.rs
+++ b/testcontainers/src/core/container.rs
@@ -3,6 +3,7 @@ use crate::{
     Image, RunnableImage,
 };
 use bollard_stubs::models::ContainerInspectResponse;
+
 use std::{fmt, marker::PhantomData, net::IpAddr, str::FromStr};
 
 /// Represents a running docker container.
@@ -182,7 +183,7 @@ where
         .unwrap_or_else(|_| panic!("container {} has missing or invalid bridge IP", self.id))
     }
 
-    pub fn exec(&self, cmd: ExecCommand) {
+    pub fn exec(&self, cmd: ExecCommand) -> ExecOutput {
         let ExecCommand {
             cmd,
             ready_conditions,
@@ -190,10 +191,15 @@ where
 
         log::debug!("Executing command {:?}", cmd);
 
-        self.docker_client.exec(self.id(), cmd);
+        let output = self.docker_client.exec(self.id(), cmd);
 
         self.docker_client
             .block_until_ready(self.id(), ready_conditions);
+
+        ExecOutput {
+            stdout: output.stdout,
+            stderr: output.stderr,
+        }
     }
 
     pub fn stop(&self) {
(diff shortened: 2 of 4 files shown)


DDtKey:
@thomaseizinger I apologize for being annoying 😅 but I'd really appreciate for reviewing/merging

thomaseizinger (changes requested):
Thanks for the ping and sorry for the long delay, I am only passively maintaining this library.

I am not sure I fully understand your usecase. Can you give me an example of what you want to do?

The commands are currently executed synchronously, meaning you can be sure that they executed once `Container::exec` returned. Why would we need another list of `WaitFor`s (apart from the fact that this isn't designed for arbitrary commands anyway)?

thomaseizinger:
Are you trying to just solve https://github.com/testcontainers/testcontainers-rs/issues/344 or do you have your own usecase you want to solve here?

DDtKey:
@thomaseizinger yes, it solves #344

> The commands are currently executed synchronously, meaning you can be sure that they executed once Container::exec returned.

That's actually not true.  Currently (in the main branch) it uses `.arg("-d")`, so it doesn't wait. So I removed that flag and added additional checks of output if you would need it.

For example I want wait until there is a message `success` in stdout OR `failed` in stderr (too simplified example though), but it's helpful in order to make sure conditions are passed without manual checks of output (moreover, `exec` doesn't return output)

DDtKey:
Agree 👍

testcontainers/src/core/logs.rs
```diff
@@ -36,21 +36,19 @@ impl<'d> LogStreamAsync<'d> {
...
 }
 
-impl LogStream {
-    pub fn new(stream: impl Read + 'static) -> Self {
-        Self {
-            inner: Box::new(stream),
-        }
+impl<R: Read> LogStream<R> {
+    pub fn new(stream: R) -> Self {
+        Self { inner: stream }
```
DDtKey, line 51: It's a bit unrelated because I reverted changes of `WaitFor`, but decided to keep it because it removes extra `Box`ing without breaking any backward compatibility

testcontainers/src/core/container.rs
```diff
@@ -213,6 +225,13 @@ where
     }
 }
 
+/// Represents an output of `exec` command.
+/// `stdout` & `stderr` is represented as bytes because it might contain non-utf chars and responsibility should be on the caller end.
+pub struct ExecOutput {
+    pub stdout: Vec<u8>,
+    pub stderr: Vec<u8>,
+}
```
DDtKey, line 233: Created separate type, because tuple with the same types could be confusing and erroneous. 

Also decided not to convert it to string, because it's up to caller if output can contain non-utf8

DDtKey:
I can squash commits if needed after review

testcontainers/src/core/container.rs
```diff
@@ -182,18 +189,23 @@ where
         .unwrap_or_else(|_| panic!("container {} has missing or invalid bridge IP", self.id))
     }
 
-    pub fn exec(&self, cmd: ExecCommand) {
+    pub fn exec(&self, cmd: ExecCommand) -> ExecOutput {
```
thomaseizinger, line 192: It is unfortunately still a breaking change but that is okay. Can you add a changelog entry please?

testcontainers/src/core/logs.rs
thomaseizinger: I am pretty sure this is a breaking change still. I'd like to keep the boxing to keep the API simpler and the type encapsulated.

thomaseizinger (commented):
Thanks for amending the PR. Some comments!

testcontainers/src/core/logs.rs
DDtKey: 🤔  looks like it's `pub(crate)` and not exposed. 

I can revert ofc, but seems to be not breaking one

testcontainers/src/core/container.rs
DDtKey: Yeah. you're right. Sure

testcontainers/src/core/container.rs
```diff
@@ -182,18 +183,23 @@ where
...
```
thomaseizinger, line 186: Can you replicate this API for the `ContainerAsync` type please?

testcontainers/src/clients/cli.rs
```diff
@@ -418,7 +421,7 @@ impl Docker for Cli {
                     self.stderr_logs(id).wait_for_message(&message).unwrap()
                 }
                 WaitFor::Duration { length } => {
-                    std::thread::sleep(length);
+                    sleep(length);
```
thomaseizinger, line 424: Let's keep the diff minimal please.

thomaseizinger (commented):
Thanks!

testcontainers/src/clients/cli.rs
DDtKey: AFAIR it was highlighted by clippy locally - because 'extra' export, but sure, not related to this PR

testcontainers/src/core/container.rs
DDtKey: `DockerAsync` doesn't have `exec`. Should I add this?

thomaseizinger: > `ContainerAsync` & `DockerAsync` doesn't have `exec`. Should I add this? 

Oh okay. We really need some tests to keep these APIs on par.

All good then, can you open an issue? Thanks!

DDtKey: Sure, created #475

PR #444, opened 2022-12-16
Add dump command to allow container logs dumping

alexpusch:
A common use case in integration tests development and debugging is inspecting created container logs. 
Currently the way to do so is to run the tests using the "keep" command and inspect the logs afterwards. This isn't a good fit for CI testing environment, and might not be ideal for local development as well.

This PR adds a new "dump" command that removes the containers just like "remove" but will dump container logs into a "./testcontainers" directory.

The logs filename is `{container_name/image_name}_{stdout/stderr}_{current date in ISO format}.log`

usage example: `TESTCONTAINERS=dump cargo test`

### DIFF
--- a/testcontainers/src/core/container.rs
+++ b/testcontainers/src/core/container.rs
@@ -290,7 +290,13 @@ where
         .image
         .container_name()
         .to_owned()
-        .unwrap_or(container.image.inner().name());
+        .unwrap_or_else(|| {
+            format!(
+                "{}_{}",
+                container.image.inner().name(),
+                &container.id()[..12]
+            )
+        });
 
     get_log_dump_file_path(log_dump_dir, &container_name, stdtype)
 }
--- a/testcontainers/src/core/container.rs
+++ b/testcontainers/src/core/container.rs
@@ -247,7 +247,9 @@ where
             Command::Remove => self.rm(),
             Command::Dump => {
                 self.stop();
-                dump_logs(self).expect("Failed to dump logs to file");
+                if let Err(error) = dump_logs(self) {
+                    log::warn!("failed to dump logs to disk - {}", error);
+                }
                 self.rm()
             }
         }
--- a/testcontainers/src/core/container.rs
+++ b/testcontainers/src/core/container.rs
@@ -3,7 +3,18 @@ use crate::{
     Image, RunnableImage,
 };
 use bollard_stubs::models::ContainerInspectResponse;
-use std::{fmt, marker::PhantomData, net::IpAddr, str::FromStr};
+use std::{
+    env, fmt,
+    fs::{self, File},
+    io,
+    marker::PhantomData,
+    net::IpAddr,
+    path::PathBuf,
+    str::FromStr,
+};
+use time::OffsetDateTime;
+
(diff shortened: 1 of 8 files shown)


testcontainers/src/core/container.rs
```diff
@@ -227,12 +246,64 @@ where
...
+        .unwrap_or("".into());
+
+    let log_file_name = format!("{container_name}_{stdtype}_{iso}");
+
+    PathBuf::from(log_file_name).with_extension("log")
+}
+
+fn get_log_dump_dir_path() -> PathBuf {
+    env::current_dir()
+        .unwrap_or(PathBuf::from("/tmp"))
```
thomaseizinger, line 303: This is gonna fail on Windows. I'd say if we can't get the current directory, print a warning and lets not dump them.

testcontainers/src/core/logs.rs
```diff
@@ -65,6 +65,10 @@ impl LogStream {
 
         Err(end_of_stream(lines))
     }
+
+    pub fn into_inner(self) -> Box<dyn Read> {
```
thomaseizinger, line 69: Is this now public API? Let's make it `pub(crate)` please.

testcontainers/src/core/container.rs
```diff
@@ -227,12 +246,64 @@ where
...
+                dump_logs(self).expect("Failed to dump logs to file");
+                self.rm()
+            }
         }
         #[cfg(feature = "watchdog")]
         crate::watchdog::unregister(self.id());
     }
 }
 
+fn dump_logs<I>(container: &Container<'_, I>) -> io::Result<()>
```
thomaseizinger, line 260: This needs to be provided for `ContainerAsync` too.

thomaseizinger (commented):
Thanks! A few comments.

alexpusch:
@thomaseizinger glad you like the feature :) 
I added support for this in ContainerAsync as well.

In addition I changed to logs directory structure so that all logs from same test run will be saved in its own directory: `./testcontainers/{iso date}/{container name|image name}_{container_id}_{stdout|stderr}.log`. I think this would provide a better user experience.
This is done by saving the date of the first log dump in a lazy static variable.

testcontainers/src/core/container.rs
alexpusch: On second thought, it might be better to allow passing the required path from outside - maybe an additional env var, or some syntax on the command env var i.e `TESTCONTAINERS=dump=/tmp/target` ? what do you think?

DDtKey:
The ability to dump logs is definitely helpful

However I'd consider more aligned interface with implementations in other languages, like [java](https://java.testcontainers.org/features/container_logs/) and [golang](https://golang.testcontainers.org/features/follow_logs/)

Also keep in mind, that we gonna deprecate CLI client soon (#563). So we actually can introduce this for http client only

Subjectively, I'd avoid configuration though env variable and dumping logs to filesystem. Instead, it could be achieved by not-removing container and accessing logs with docker cli.

DDtKey:
Let us know if you're still interested in implementing / actualizing this PR.
We actively revamp the project

Issue #386, opened 2022-07-15
Re-work public API to be `Container` centric


thomaseizinger:
Since its inception, `testcontainers-rs` had an API that focused on first creating a `Client` before one could `run` a `Container`. This was deemed acceptable as a `Client` is typically only created once at the top of the test.

Recent feature requests such as #354, show some of the limitations of this API. One of the primary design goals of `testcontainers-rs` is to have a really small API such that creating and running a new container is as easy as possible.

To accommodate feature requests such as #354 but still adhere to this design goal, I am thinking to re-design the API to something like this:

```rust
use testcontainers::RunViaCli; // Extension trait for running an image using the `Cli` client.

let container = MyImage::default().run();
```

```rust
use testcontainers::RunViaHttp; // Extension trait for running an image using the `Http` client.

let container = MyImage::default().run().await;
```

The client would be started on-demand and kept in a global variable with a `Weak` reference so that it shuts down once all containers are gone. The client itself would likely be made completely private to the crate to further reduce the public API (which allows us to make more changes without breaking compatibility).

We'd have to flesh out some details on how to configure the individual clients if they can no longer be constructed directly but that is not too important for this topic.

// CC several people who had issues with the current API in the past: @soulseekeer24 @lcmgh @marlon-sousa @jhershberg @pronvis @ufoscout 

Any feedback / input on this idea is much appreciated.

marlon-sousa:
Hello,

I strongly encourage this implementation.

While this crate was aimed at isolated tests, it is too good and people's needs (beyond this should be this or that way) are too extensive. All the needs can take advantage of this crate should the proposed architecture be implemented.

Real life some times goes further than what "important people", personal believes  or famous theories have to say, mature environments should consider all, and I mean really all things a matter of trade offs ratter than right or wrong.

This said, the proposed change will allow a vast range of testing needs to succeed, using an ergonomic, fast and well encapsulated strategy. Everyone wins, nobody is left behind and, independently of what we think testing should be like, everyone can just use the crate for whatever strategies they need to use. This is what a really good crate should be, and your crate just happens to be great, otherwise people would just switch for shell scripts or for invoking process around using std::process::Command instead of coming here and asking for help.

I for example am running a set of integration tests that run sequentially and all use the same database, thus I need to start a container when tests start to work and stop it when the suite ends.

[...]

PR #445, opened 2022-12-22
Remove redundant lifetime parameters

eharcevs:
Is the `'d` lifetime parameter in `Container` and `ContainerAsync` redundant? The comment says:
```
Tracks the lifetime of the client to make sure the container is dropped before the client.
```
This seems unnecessary since the container already holds a pointer (`Box`) to the client anyway, which ensures that the container will get dropped before the client. Dropping the redundant lifetime parameter results in a simpler API. wdyt?

### DIFF
--- a/testcontainers/src/clients/cli.rs
+++ b/testcontainers/src/clients/cli.rs
@@ -24,7 +24,7 @@ pub struct Cli {
 }
 
 impl Cli {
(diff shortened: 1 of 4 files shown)


thomaseizinger:
Thank you for the PR!

I'd like to move the API towards https://github.com/testcontainers/testcontainers-rs/issues/386 instead which will also remove this lifetime in favor of runtime reference counting.

PR #446, opened 2022-12-31
Support specifying images using sha256 digests

### LINKED ISSUE #411: Support specifying images uaing sha256 digests
You can run docker images by specifying the sha256 digest of the designated image instead of using a specific version.

Using sha256 is somewhat more secure as it pins the exact image, thus not allowing possible overriding of a tag by a malicious actor.

Usage can be something along:
```rust
impl Image for SomeImage {
    type Args = ();

    fn name(&self) -> String {
        "some name"
    }

    fn sha256(&self) -> String {
       "a826cb8a6a0023..."
    }
}
```

boazberman:
Closes #411

### DIFF
--- a/testcontainers/src/clients/cli.rs
+++ b/testcontainers/src/clients/cli.rs
@@ -608,6 +608,42 @@ mod tests {
     }
 
     #[test]
+    fn cli_run_command_with_custom_sha256_should_attach_that() {
+        let image = GenericImage::new("hello", "0.0");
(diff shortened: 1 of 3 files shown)


DDtKey:
Closing the PR as we gonna deprecate CLI client soon and changes look a bit deprecated.

But let's keep the issue and discuss there the support after cleaning up the repo.

Thank you!

PR #462, opened 2023-02-08
Feature: Add support for localstack

joanmiespada:
This change adds support for localstack image, latest version.

localstack: https://hub.docker.com/r/localstack/localstack

### DIFF
--- a/testcontainers/src/images/local_stack.rs
+++ b/testcontainers/src/images/local_stack.rs
@@ -2,7 +2,7 @@ use crate::{core::WaitFor, Image};
 use std::collections::HashMap;
 
 const NAME: &str = "localstack/localstack";
-const TAG: &str = "latest";
+const TAG: &str = "1.4";
 const PORT: u16 = 4566;
 
 #[derive(Debug)]
@@ -12,25 +12,13 @@ pub struct LocalStack {
     tag: String,
 }
 
-impl LocalStack{
-    pub fn set_services(&mut self, services: &str) {
-
-        self.env_vars.insert("SERVICES".to_owned(), services.to_owned());
-
-    }
-}
-
 impl Default for LocalStack {
     fn default() -> Self {
         let mut env_vars = HashMap::new();
-        //env_vars.insert("SERVICES".to_owned(), "s3".to_owned());
         env_vars.insert("DEBUG".to_owned(), "1".to_owned());
         env_vars.insert("PORT_WEB_UI".to_owned(), "8080".to_owned());
         env_vars.insert("LAMBDA_EXECUTOR".to_owned(), "docker".to_owned());
         env_vars.insert("DOCKER_HOST".to_owned(), "unix:///var/run/docker.sock".to_owned());
-        env_vars.insert("AWS_DEFAULT_REGION".to_owned(), "eu-central-1".to_owned());
-        env_vars.insert("AWS_ACCESS_KEY_ID".to_owned(), "test".to_owned());
-        env_vars.insert("AWS_SECRET_ACCESS_KEY".to_owned(), "test".to_owned());
         env_vars.insert("DATA_DIR".to_owned(), "/tmp/localstack/data".to_owned());
 
         let mut volumes = HashMap::new();
--- a/testcontainers/src/images/local_stack.rs
+++ b/testcontainers/src/images/local_stack.rs
@@ -55,7 +55,7 @@ impl Image for LocalStack {
     }
 
     fn ready_conditions(&self) -> Vec<WaitFor> {
-        vec![WaitFor::message_on_stdout( format!("Starting edge router (https port {})...", PORT))]
+        vec![WaitFor::message_on_stdout( format!("Ready."))]
     }
 
     fn env_vars(&self) -> Box<dyn Iterator<Item = (&String, &String)> + '_> {
--- a/testcontainers/src/images/local_stack.rs
+++ b/testcontainers/src/images/local_stack.rs
@@ -1,0 +1,72 @@
+use crate::{core::WaitFor, Image};
+use std::collections::HashMap;
+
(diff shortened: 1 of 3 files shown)


testcontainers/src/images/local_stack.rs
```diff
@@ -0,0 +1,72 @@
...
+    fn name(&self) -> String {
+        NAME.to_owned()
+    }
+
+    fn tag(&self) -> String {
+        self.tag.to_owned()
+    }
+
+    fn ready_conditions(&self) -> Vec<WaitFor> {
+        vec![WaitFor::message_on_stdout( format!("Starting edge router (https port {})...", PORT))]
```
mdelapenya, line 58: Hi from testcontainers-go 👋 

I'm jumping here to share that in other languages we define a different waitFor log strategy:
- Go: https://github.com/testcontainers/testcontainers-go/pull/800/files#diff-bd71ec18d411b5b0b51723761d6d0a95e60ad7d5578828df544e14d6518a222dR53
- Java: https://github.com/testcontainers/testcontainers-java/blob/343c5a533f4c6148d6e6eea3a3374116365bc91d/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java#L98

We can probably ping @eddumelendez and @kiview for creating alignment across different implementations, but I'm not against setting this one as the default, so it's not a blocker for me if the maintainers decide to merge it :)

joanmiespada:
Hi @mdelapenya, thanks for your comment. I've updated it, in the same way as go and java. Please, check.

testcontainers/src/images/local_stack.rs
```diff
@@ -0,0 +1,72 @@
...
+    volumes: HashMap<String, String>,
+    tag: String,
+}
+
+impl LocalStack{
+    pub fn set_services(&mut self, services: &str) {
+
+        self.env_vars.insert("SERVICES".to_owned(), services.to_owned());
+
+    }
```
HofmeisterAn, line 20: Having an API to configure the services seems unnecessary for the latest versions of LocalStack. The services are "available" immediately anyway.

For Testcontainers implementations that do not need to support backwards compatibility (older versions) like .NET and Go we decided not to add an API to configure the services. Developers can simply use the API that configures environment variables (if necessary).

testcontainers/src/images/local_stack.rs
```diff
@@ -0,0 +1,72 @@
...
+    fn default() -> Self {
+        let mut env_vars = HashMap::new();
+        //env_vars.insert("SERVICES".to_owned(), "s3".to_owned());
+        env_vars.insert("DEBUG".to_owned(), "1".to_owned());
+        env_vars.insert("PORT_WEB_UI".to_owned(), "8080".to_owned());
+        env_vars.insert("LAMBDA_EXECUTOR".to_owned(), "docker".to_owned());
+        env_vars.insert("DOCKER_HOST".to_owned(), "unix:///var/run/docker.sock".to_owned());
+        env_vars.insert("AWS_DEFAULT_REGION".to_owned(), "eu-central-1".to_owned());
+        env_vars.insert("AWS_ACCESS_KEY_ID".to_owned(), "test".to_owned());
+        env_vars.insert("AWS_SECRET_ACCESS_KEY".to_owned(), "test".to_owned());
```
HofmeisterAn, line 33: I do not think these environment variables are necessary.

testcontainers/src/images/local_stack.rs
```diff
@@ -0,0 +1,72 @@
...
+    fn name(&self) -> String {
+        NAME.to_owned()
+    }
+
+    fn tag(&self) -> String {
+        self.tag.to_owned()
+    }
+
+    fn ready_conditions(&self) -> Vec<WaitFor> {
+        vec![WaitFor::message_on_stdout( format!("Ready."))]
```
HofmeisterAn, line 58: Recent versions should use the health check endpoint `/_localstack/health` to detect readiness of the container.

testcontainers/src/images/local_stack.rs
```diff
@@ -0,0 +1,72 @@
+use crate::{core::WaitFor, Image};
+use std::collections::HashMap;
+
+const NAME: &str = "localstack/localstack";
+const TAG: &str = "latest";
```
HofmeisterAn, line 5:
```suggestion
const TAG: &str = "1.4";
```

Pin the version, maybe even to `1.4.0`.

HofmeisterAn (commented):
We had some discussions today about the LocalStack implementation in .NET and Go. I would like to share our thoughts (see comments below).

testcontainers/src/images/local_stack.rs
joanmiespada: agree

testcontainers/src/images/local_stack.rs
joanmiespada: ok

testcontainers/src/images/local_stack.rs
joanmiespada: let's remove them.

testcontainers/src/images/local_stack.rs
joanmiespada: ok, but it'd requiere more effort parsing the /health response, right? or do you suggest to check it it reponses with HTTP 200/OK?

HofmeisterAn: Yes, we check for HTTP 200/OK.

HofmeisterAn: I'm afraid my knowledge of the Rust implementation is limited. Are you saying there is no wait strategy available to wait for an HTTP status code? Perhaps we could use `WaitFor::Healthcheck` or add the strategy to Rust as a new feature. That would be really useful.

thomaseizinger:

PR #465, opened 2023-02-19
Support changing command line args in RunnableImage

mrpi:
I found no way for setting args on a GenericImage ... so I added one ;-)

### DIFF
--- a/testcontainers/src/clients/cli.rs
+++ b/testcontainers/src/clients/cli.rs
@@ -561,6 +561,20 @@ mod tests {
     }
 
     #[test]
+    fn cli_run_command_should_include_args() {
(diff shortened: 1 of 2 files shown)


andrey-yantsen:
There's a way, but I don't think it's documented.
```rust
let image = GenericImage::new("hello", "0.0");
let image = RunnableImage::from((image, vec!["--a".to_string(), "arg_value".to_string()]));
```

mrpi:
> There's a way, but I don't think it's documented.
> 
> ```rust
> let image = GenericImage::new("hello", "0.0");
> let image = RunnableImage::from((image, vec!["--a".to_string(), "arg_value".to_string()]));
> ```

You are right, that works! Thanks

Issue #471, opened 2023-03-08
Remove all images from the library


thomaseizinger:
Based on https://github.com/testcontainers/testcontainers-rs/blob/dev/DESIGN_PRINCIPLES.md#ease-of-use, we are removing all existing images from the library.

For one, users tend to not own the images they are contributing, see https://github.com/testcontainers/testcontainers-rs/issues/255, thus placing a big maintenance burden on us.

Two, we want the library to be easy to use such that it is almost not worth bundling them up. Additionally, as we reach a stable API, users can create their own crates for images if they want to.

plaflamme:
Completely understandable from a maintenance perspective.

From an "ease-of-use" perspective, there's an argument to be made that having these ready-made images greatly simplifies the amount of work that a developer needs to do to "get going". A well-designed test image isn't always simple to create, so having a central place to share these would be great and would arguably make it easier to use the library.

Would it be reasonable to consider something like `testcontainers-rs-contrib`? It could allow the community to share images while allowing a different level of support than for `testcontainers-rs` itself from the maintainers. Having a repo-per-image seems like a high maintenance bar for anyone; bundling them up in a single repository would help in that regard.

gurinderu:
Probably it is better to split it into a few crates and describe a dependency list for each crate.
I suggest a such structure
root
|-- Cargo.toml
|-- crates
     |---- core
             |-- Cargo.toml
     |---- kafka
             |-- Cargo.toml
     ....

plaflamme:
Personally, I'd be more inclined to use cargo features instead. At least, I'd start with that before breaking things up into many crates; seems simpler to manage packaging / releasing that way?

thomaseizinger:
As said above, I am more than happy to link to one or more of these crates from the README. Somebody just needs to step up and create a repo :)

We are very permissibly licensed already but you are free to copy all the code verbatim if you want!

gurinderu:
Guys, I made another pr with another crate https://github.com/testcontainers/testcontainers-rs/pull/476

DDtKey:
@gurinderu I believe that @thomaseizinger meant that it shouldn't be as part of this repo & team. Because of responsibility, but it's fine to create community-library with all images

gurinderu:
@DDtKey this library is also community-driven ) I don't see any visible profits to make it in the separate repo, moreover, it will bring more pain

thomaseizinger:
[...]

PR #476, opened 2023-03-10
Split core library with images

gurinderu:
Another way to solve https://github.com/testcontainers/testcontainers-rs/issues/471


thomaseizinger:
The point of #471 is to remove the images from the repository such that we no longer need to maintain them. If you want to split it out into a crate, please create your own repository.

Issue #480, opened 2023-03-31
Connection Refused after exposing port


open-schnick:
Hi,
Thanks for porting testcontainers to rust.
Sadly i am having quite a hard time getting them to work :')
I broke down one of my problems into a minimal example:

```rust
#[tokio::test(flavor = "multi_thread")]
async fn minimal_test() {
    let docker = Cli::docker();
    let pubsub_port = docker.run(PubsubEmulator).get_host_port_ipv4(8085);

    reqwest::get(format!("http://localhost:{pubsub_port}"))
        .await
       .unwrap();
}
```

My image looks like this:

```rust
#[derive(Debug, Default)]
pub struct PubsubEmulator;

impl Image for PubsubEmulator {
    type Args = PubsubEmulatorArgs;

    fn name(&self) -> String {
        "gcr.io/google.com/cloudsdktool/cloud-sdk".to_owned()
    }

    fn tag(&self) -> String {
        "423.0.0-emulators".to_owned()
    }

    fn ready_conditions(&self) -> Vec<testcontainers::core::WaitFor> {
        // FIXME: does not work
        // vec![WaitFor::StdOutMessage {
        //     message: "[pubsub] INFO: Server started, listening on 8085".to_owned(),
        // }]
        vec![WaitFor::seconds(10)]
    }

    fn entrypoint(&self) -> Option<String> {
        Some("gcloud".to_owned())
    }

    fn expose_ports(&self) -> Vec<u16> {
        vec![8085]
    }
}

#[derive(Debug, Clone)]
pub struct PubsubEmulatorArgs {
    args: Vec<String>,
}

impl ImageArgs for PubsubEmulatorArgs {
    fn into_iterator(self) -> Box<dyn Iterator<Item = String>> {
        Box::new(self.args.into_iter())
    }
}

impl Default for PubsubEmulatorArgs {
    fn default() -> Self {
        let args = vec![
            "beta",
            "emulators",
            "pubsub",
            "start",
            "--project=local-project",
            "--host-port=0.0.0.0:8085",
        ]
        .iter()
        .map(|arg| arg.to_owned().to_owned())
        .collect::<Vec<String>>();

        Self { args }
    }
}
```

From reading the code and the docs i think this should work. But i cant connect to the docker container although it looks like it is mapped in the cli.
The same image implemented in a docker-compose file works:
```yml
services:
  pubsub-emulator:
[...]

Issue #227, opened 2020-10-21
Experiment with an API design that statically couples available ports to the started image


thomaseizinger:
Currently, users can request a mapped host port from a container through `Container#get_host_port`.

This API allows users to pass whichever port the want and the function will simply return an `Option` if the port was not mapped.

It would be interesting to explore an API where implementations of `Image` could somehow statically specify, which particular ports will be exposed. They should "know" that because we know which docker image we will be starting.

The advantage I am trying to gain from that is increased discoverability:

If I am starting a `postgres` container through `testcontainers`, I don't want to need to know that postgres exposes port `5432`. Instead, it would be nice if I could discover that through intelli-sense on the container that I just started.

PR #485, opened 2023-04-16
Improve docs around custom drop implementation of Containers

open-schnick:
Hi,
i improved the docs so other people don't fall into the same pitfalls as i did (#480)


thomaseizinger (commented):
Thank you for the suggestion!

To be frank, I think this is documented very clearly already. I literally says "The container will be stopped once it goes out of scope".

I don't think we need to explain how Rust works in our docs. Ideally, we fix this by making it a compile-error to not bind to a variable. See https://github.com/testcontainers/testcontainers-rs/issues/227.

PR #503, opened 2023-06-24
[DRAFT] add extension traits to run container without explicitly creating a c…

endertunc:
This is an in progress work regarding to #386 . I just wanted to share the current prototype I have to validate my understanding. 

- [ ] Agree on the design and identify the breaking changes and document
- [ ] Go thorough what we expose as public API and make sure that we don't expose something we don't need to.
- [ ] Update tests

### DIFF
--- a/testcontainers/Cargo.toml
+++ b/testcontainers/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2021"
 keywords = [ "docker", "testcontainers" ]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/testcontainers/testcontainers-rs"
-rust-version = "1.60.0"
+rust-version = "1.70.0"
 description = "A library for integration-testing against docker containers from within Rust."
 
 [dependencies]
--- a/testcontainers/src/clients/cli.rs
+++ b/testcontainers/src/clients/cli.rs
@@ -7,7 +7,7 @@ use std::{
     collections::HashMap,
     ffi::{OsStr, OsString},
     process::{Child, Command, Stdio},
-    sync::{Arc, RwLock},
+    sync::{OnceLock, RwLock},
     thread::sleep,
     time::{Duration, Instant},
 };
@@ -15,22 +15,21 @@ use std::{
 const ONE_SECOND: Duration = Duration::from_secs(1);
 const ZERO: Duration = Duration::from_secs(0);
 
-/// Implementation of the Docker client API using the docker cli.
-///
-/// This (fairly naive) implementation of the Docker client API simply creates `Command`s to the `docker` CLI. It thereby assumes that the `docker` CLI is installed and that it is in the PATH of the current execution environment.
-#[derive(Debug)]
-pub struct Cli {
-    inner: Arc<Client>,
+static CLI_DOCKER: OnceLock<Client> = OnceLock::new();
+fn docker_client() -> &'static Client {
+    CLI_DOCKER.get_or_init(|| Client::new::<env::Os>())
 }
 
-impl Cli {
-    pub fn run<I: Image>(&self, image: impl Into<RunnableImage<I>>) -> Container<'_, I> {
-        let image = image.into();
+pub trait RunViaCli<I: Image> {
+    fn run(self) -> Container<I>;
+}
 
-        if let Some(network) = image.network() {
-            if self.inner.create_network_if_not_exists(network) {
-                let mut guard = self
-                    .inner
+impl<I: Image> RunViaCli<I> for RunnableImage<I> {
+    fn run(self) -> Container<I> {
+        let docker = docker_client();
+        if let Some(network) = self.network() {
+            if docker.create_network_if_not_exists(network) {
+                let mut guard = docker
                     .created_networks
                     .write()
                     .expect("failed to lock RwLock");
@@ -39,7 +38,7 @@ impl Cli {
             }
         }
 
-        let mut command = Client::build_run_command(&image, self.inner.command());
+        let mut command = Client::build_run_command(&self, docker.command());
 
         log::debug!("Executing command: {:?}", command);
 
@@ -52,19 +51,16 @@ impl Cli {
             .to_string();
 
         #[cfg(feature = "watchdog")]
-        if self.inner.command == env::Command::Remove {
+        if cli.command == env::Command::Remove {
             crate::watchdog::register(container_id.clone());
         }
 
-        self.inner.register_container_started(container_id.clone());
+        docker.register_container_started(container_id.clone());
 
-        self.block_until_ready(&container_id, image.ready_conditions());
+        docker.block_until_ready(&container_id, self.ready_conditions());
 
-        let client = Cli {
-            inner: self.inner.clone(),
-        };
-
-        let container = Container::new(container_id, client, image, self.inner.command);
+        let container: Container<I> =
+            Container::new(container_id, docker.clone(), self, docker.command);
 
         for cmd in container
             .image()
@@ -241,35 +237,25 @@ impl Client {
     }
 }
 
-impl Default for Cli {
-    fn default() -> Self {
-        Self::new::<env::Os>()
-    }
-}
-
-impl Cli {
+impl Client {
     pub fn new<E>() -> Self
     where
         E: GetEnvValue,
     {
         Self {
-            inner: Arc::new(Client {
-                container_startup_timestamps: Default::default(),
-                created_networks: Default::default(),
-                binary: "docker".into(),
-                command: env::command::<E>().unwrap_or_default(),
-            }),
+            container_startup_timestamps: Default::default(),
+            created_networks: Default::default(),
+            binary: "docker".into(),
+            command: env::command::<E>().unwrap_or_default(),
         }
     }
 }
 
-impl Docker for Cli {
+impl Docker for &Client {
     fn stdout_logs(&self, id: &str) -> LogStream {
-        self.inner
-            .wait_at_least_one_second_after_container_was_started(id);
+        self.wait_at_least_one_second_after_container_was_started(id);
 
         let child = self
-            .inner
             .command()
             .arg("logs")
             .arg("-f")
@@ -283,11 +269,9 @@ impl Docker for Cli {
     }
 
     fn stderr_logs(&self, id: &str) -> LogStream {
-        self.inner
-            .wait_at_least_one_second_after_container_was_started(id);
+        self.wait_at_least_one_second_after_container_was_started(id);
 
         let child = self
-            .inner
             .command()
             .arg("logs")
             .arg("-f")
@@ -311,7 +295,6 @@ impl Docker for Cli {
 
     fn inspect(&self, id: &str) -> ContainerInspectResponse {
         let output = self
-            .inner
             .command()
             .arg("inspect")
             .arg(id)
@@ -335,7 +318,6 @@ impl Docker for Cli {
 
     fn rm(&self, id: &str) {
         let output = self
-            .inner
             .command()
             .arg("rm")
             .arg("-f")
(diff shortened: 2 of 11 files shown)


testcontainers/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2021"
 keywords = [ "docker", "testcontainers" ]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/testcontainers/testcontainers-rs"
-rust-version = "1.60.0"
+rust-version = "1.70.0"
```
endertunc, line 10: I am using sync::OnceLock which was added in 1.70 to store static value  but if we want to still be in 1.60 we can achieve the same with using something like https://crates.io/crates/once_cell

testcontainers/src/clients/cli.rs
```diff
@@ -7,75 +7,20 @@ use std::{
...
-/// Implementation of the Docker client API using the docker cli.
-///
-/// This (fairly naive) implementation of the Docker client API simply creates `Command`s to the `docker` CLI. It thereby assumes that the `docker` CLI is installed and that it is in the PATH of the current execution environment.
-#[derive(Debug)]
-pub struct Cli {
-    inner: Arc<Client>,
+static CLI_DOCKER: OnceLock<Client> = OnceLock::new();
+fn docker_client() -> &'static Client {
+    CLI_DOCKER.get_or_init(|| Client::new::<env::Os>())
 }
```
endertunc, line 21: Here we have global client which is used implicitly by `Docker` implementation.

testcontainers/src/clients/cli.rs
```diff
@@ -464,6 +392,59 @@ impl Drop for Client {
     }
 }
 
+pub trait RunViaCli<I: Image> {
+    fn run(self) -> Container<I>;
+}
```
endertunc, line 397: ^ extension trait to use CLI  as docker backend.

testcontainers/src/clients/cli.rs
```diff
@@ -464,6 +392,59 @@ impl Drop for Client {
...
+
+impl<I: Image> RunViaCli<I> for RunnableImage<I> {
+    fn run(self) -> Container<I> {
```
endertunc, line 400: `run` function moved from `Http` to `RunnableImage` it can be directly called on `RunnableImage`

testcontainers/src/clients/http.rs
```diff
@@ -33,46 +30,112 @@ struct Client {
...
+    fn drop(&mut self) {
+        match self.command {
+            env::Command::Remove => {
+                let guard = self.created_networks.read().expect("failed to lock RwLock");
+                for network in guard.iter() {
+                    block_on(async { self.bollard.remove_network(network).await.unwrap() });
+                }
+            }
+            env::Command::Keep => {}
+        }
```
endertunc, line 103: This is something we need to talk I believe. I recently learned that static variables don't actually call `Drop` so we might have to move this logic to `WatchDog` or find another way to create global static variable. 

Arc could be useful but I am not sure how to create and Arc that behaves like static variable.

endertunc (commented):
Hi @thomaseizinger I created this in progress pull request to validated my understanding. Could you please take a look and let me know if what I am doing makes sense here?

thomaseizinger:
> Hi @thomaseizinger I created this in progress pull request to validated my understanding. Could you please take a look and let me know if what I am doing makes sense here?

Will review next week!

testcontainers/Cargo.toml
thomaseizinger: In general I'd expect users to run tests with a recent Rust compiler so this is fine.

testcontainers/src/clients/cli.rs
thomaseizinger: Can we move this function up in the file such that the diff is minimal? That would help reviewing a lot :)

testcontainers/src/clients/http.rs
thomaseizinger: Yeah a `static` variable will only be dropped at the end of a program.

Rust executes each test in its own thread so we could look into using a thread-local variable for this. That would clean up a docker client at the end of each thread, i.e. at the end of each test and each test would gets its own docker client.

testcontainers/src/clients/http.rs
```diff
@@ -181,102 +242,22 @@ impl Http {
...
 
-        let client = Http {
-            inner: self.inner.clone(),
-        };
-
-        ContainerAsync::new(container_id, client, image, self.inner.command).await
-    }
-}
-
-impl Http {
```
thomaseizinger, line 202: Same as above, it would be nice to not move this code block but keep the diff as minimal as possible :)

testcontainers/src/clients/http.rs
```diff
@@ -416,7 +391,7 @@ mod tests {
         let image = RunnableImage::from(image).with_container_name("hello_container");
         let container = docker.run(image).await;
 
-        let container_details = inspect(&docker.inner.bollard, container.id()).await;
+        let container_details = inspect(&docker.bollard, container.id()).await;
```
thomaseizinger, line 394: Can you rewrite these tests to use the new API so we get an idea what it feels like / whether it works?

thomaseizinger (commented):
Great start overall! I'd like to see some tests updated so we get an idea what the API feels like.

testcontainers/src/clients/http.rs
endertunc: I just pushed a commit with updated test. Let me know how it looks.

testcontainers/src/clients/http.rs
endertunc: It should be better now. 👍

testcontainers/src/clients/cli.rs
endertunc: It should be better now. 👍

testcontainers/src/clients/http.rs
endertunc: Now the tests are updated there are 3 failing tests and all of them about related to "networks" which were suppose to be cleaned up but it's not working due to the issue I mentioned.

I see 3 options here;
- The method you mentioned - (I am not very familiar with rust in that regard so if you let me know how this could be achieved I can take a look)

- We can integrated with [ryuk](https://github.com/testcontainers/moby-ryuk) which would handle all kind of clean-up scenarios also align with other testcontainers projects. (I am not %100 sure but this would mean that some test cases can't be run in parallel) This approach is similar to watchdog feature we have in this repo but ryuk is client agnostic.
 
- For this specific use case, we can keep track of number of containers which uses the given network and remove at when the last container is dropped.

```
// assume this is global variable 
let networks = Mutex<Map<String,i32>> // network_name -> number_of_containers_using_the_network

// inside start function
insert or increase the number of reference to given network key in global map

// inside impl drop for container
decrease the number of reference to given network. If number of reference is 0 after this operation remove the network as well.
```

I personally prefer option two so we can align with other implementations and also I like the ryuk approach in terms of design but I would like to hear what you think.

testcontainers/src/clients/cli.rs
```diff
@@ -7,30 +7,39 @@ use std::{
...
-        if let Some(network) = image.network() {
-            if self.inner.create_network_if_not_exists(network) {
-                let mut guard = self
-                    .inner
+impl<I> RunViaCli<I> for I
+where
+    I: Image,
+    I::Args: Default,
+{
+    fn start(self) -> Container<I> {
```
endertunc, line 32: This is an helper implementation that allow you to start container directly from Image without needing to wrap image with RunnableImage::from

testcontainers/src/clients/http.rs
```diff
@@ -33,46 +32,49 @@ struct Client {
...
-impl Default for Http {
-    fn default() -> Self {
-        Self::new()
+#[async_trait]
+impl<I> RunViaHttp<I> for I
+where
+    I: Image,
+    I::Args: Default,
+{
+    async fn start(self) -> ContainerAsync<I> {
```
endertunc, line 46: This is an helper implementation that allow you to start container directly from Image without needing to wrap image with RunnableImage::from

testcontainers/src/clients/cli.rs
```diff
@@ -241,35 +247,25 @@ impl Client {
...
+            container_startup_timestamps: Default::default(),
+            created_networks: Default::default(),
+            binary: "docker".into(),
+            command: env::command::<E>().unwrap_or_default(),
         }
     }
 }
 
-impl Docker for Cli {
+impl Docker for &Client {
```
thomaseizinger, line 264: Do we still need the `Docker` trait now? By virtue of the `RunViaCli` trait, the implementation is no longer pluggable so I believe we could just remove it now.

testcontainers/src/clients/http.rs
```diff
@@ -276,7 +269,7 @@ impl Drop for Client {
 }
 
 #[async_trait]
-impl DockerAsync for Http {
+impl DockerAsync for &Client {
```
thomaseizinger, line 272: Same here, I think we can remove this trait and just say `impl Client` here.

testcontainers/tests/cli_client.rs
```diff
@@ -22,20 +23,16 @@ impl Image for HelloWorld {
...
 
 #[test]
 fn generic_image_with_custom_entrypoint() {
-    let docker = clients::Cli::default();
     let msg = WaitFor::message_on_stdout("server is ready");
 
     let generic = GenericImage::new("simple_web_server", "latest").with_wait_for(msg.clone());
 
-    let node = docker.run(generic);
+    let node = generic.start();
```
thomaseizinger, line 35: What I like about this new API is that we can chain it, can we rewrite the tests to do that?

testcontainers/src/clients/http.rs
thomaseizinger: Looks good!

testcontainers/src/clients/http.rs
```diff
@@ -411,49 +402,55 @@ mod tests {
 
     #[tokio::test(flavor = "multi_thread")]
     async fn http_run_command_should_include_name() {
-        let docker = Http::new();
+        let docker = docker_client();
         let image = GenericImage::new("hello-world", "latest");
-        let image = RunnableImage::from(image).with_container_name("hello_container");
-        let container = docker.run(image).await;
+        let container = RunnableImage::from(image)
+            .with_container_name("hello_container")
```
thomaseizinger, line 408: Can we figure out a way to make this more ergonomic? Ideally, I'd like users to never need to construct `RunnableImage` but only `Image`s. `with_container_name` is defined in a trait, right? Does it already work to just write:

```rust
GenericImage::new().with_container_name().start().await;
```

thomaseizinger (commented):
Sorry for the late review, I don't get to spend much time with this repository.

Great progress though, I left some comments :)

testcontainers/src/clients/http.rs
thomaseizinger: Essentially, what we need to achieve is that a client is only alive for as long as any of its containers are alive. Thus I'd suggest the following:

- Store a `Mutex<Option<Weak<Client>>>` in the static variable
- When you try to get a reference to the client:
	- Grab a lock to the mutex
	- If the value is `None`:
		- Make a new client
		- Wrap it in an `Arc`
		- Make a copy
		- Downgrade the copy to a `Weak`
		- Store the `Weak` inside the `Option`
		- Release the lock
		- Return the `Arc` you still have around
	- If the value is `Some`:
		- Attempt to upgrade to an `Arc`
			- If successful, return
			- If failed, do the same thing as for the above `None` case

I believe this should achieve the following:

- As long is there is at least 1 `Container` around, we can grab a reference to it and make more containers.
- If all containers are dropped, the last strong reference to the `Client` is also dropped which means `Drop` is being called on `Client`

I see integration with `ryuk` as a separate feature.

PR #510, opened 2023-07-13
Enhance: support mapping port to specified host ip:port

4t145:
Support to build run command with args like `-p 127.0.0.1:8000:8000`, and make struct `Port` can be converted from `(SocketAddr, u16)`. 
e.g.
```rust
let localhost_8000 = SocketAddrV4::new(Ipv4Addr::LOCALHOST, 8000);
some_runnable_image.with_mapped_port((localhost_8000 , 8000));
```

### DIFF
--- a/testcontainers/src/clients/cli.rs
+++ b/testcontainers/src/clients/cli.rs
@@ -180,7 +180,7 @@ impl Client {
             for port in ports {
                 command
                     .arg("-p")
-                    .arg(format!("{}:{}", port.local, port.internal));
+                    .arg(port.to_string());
             }
         } else if !is_container_networked {
             for port in image.expose_ports() {
--- a/testcontainers/src/core/image.rs
+++ b/testcontainers/src/core/image.rs
@@ -1,4 +1,10 @@
-use std::{collections::BTreeMap, env::var, fmt::Debug, time::Duration};
+use std::{
+    collections::BTreeMap,
+    env::var,
+    fmt::Debug,
+    net::{IpAddr, SocketAddr},
+    time::Duration,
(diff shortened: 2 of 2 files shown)


thomaseizinger:
Can you expand why this is needed?

4t145:
> Can you expand why this is needed?

If I don't specify the host, it would forward docker container's port to `::0` host so it would be exposed to public. In some cases in test this may be security concerns. Most of the time, test containers use weak passwords.

thomaseizinger:
> Most of the time, test containers use weak passwords.

But they are also ephemeral and only contain test data? I don't really understand the problem we are solving here. Could we just always bind the port to `127.0.0.1`. How do other implementations solve this?

4t145:
> But they are also ephemeral and only contain test data? 

Indeed. 
My thoughts at that time were like: “Why don't support it since docker allow us to do so?”.
Actually it was my leader asked me to bind it to localhost because it may be scaned by security tools and trigger alarm.
It's not a problem if we don't support this feature, but in turn, there is no loss in supporting this feature.

thomaseizinger:
> > But they are also ephemeral and only contain test data? 
> 
> Indeed. 
> My thoughts at that time were like: “Why don't support it since docker allow us to do so?”.

Because `testcontainers` is not a generic docker client :)

> Actually it was my leader asked me to bind it to localhost because it may be scaned by security tools and trigger alarm.
> It's not a problem if we don't support this feature, but in turn, there is no loss in supporting this feature.

There is. Every feature has a cost in terms of maintenance effort and users have to learn the API.

PR #518, opened 2023-08-23
feat: Add `user` option when running container.

liurenjie1024:


ns-sjorgedeaguiar:
@DDtKey I see this old PR is closed; is there a way to set the user when executing commands or starting the image? I'm working with Postgres and seeing the error `"root" execution of the PostgreSQL server is not permitted.`. Am I missing something?

PR #523, opened 2023-09-25
docs: mention `testcontainers-modules`

DDtKey:
It would be nice to mention the new location of ready-to-use images, because of https://github.com/testcontainers/testcontainers-rs/issues/471

It will helpful for users once `0.15.0` is released.

### DIFF
--- a/testcontainers/src/lib.rs
+++ b/testcontainers/src/lib.rs
@@ -27,7 +27,8 @@
 //!
 //! # Ecosystem
 //!
-//! The testcontainers ecosystem is split into multiple crates, however, the `testcontainers` crate itself is a meta-crate, re-exporting the others. Usually, depending on `testcontainers` should be sufficient for most users needs.
+//! `testcontainers` is the core crate that provides an API for working with containers in a test environment.
+//!  However, it does not provide ready-to-use modules, you can implement your [`Image`]s using the library directly or use community supported [`testcontainers-modules`].
 //!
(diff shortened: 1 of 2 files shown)


testcontainers/src/lib.rs
```diff
@@ -28,13 +28,15 @@
 //! # Ecosystem
 //!
 //! The testcontainers ecosystem is split into multiple crates, however, the `testcontainers` crate itself is a meta-crate, re-exporting the others. Usually, depending on `testcontainers` should be sufficient for most users needs.
```
thomaseizinger, line 30: This is actually outdated, we only have a single crate now :)

thomaseizinger (commented):
Thanks!

Can you fix that outdated doc line while we are at it?

testcontainers/src/lib.rs
DDtKey: Sure, you're right.

Updated, please take a look if it looks good to you

thomaseizinger (approved):

thomaseizinger:
@DDtKey The README also still points at the integration tests. Mind pointing towards the modules instead?

DDtKey:
It makes sense to me, I've updated the readme to include 2 sections:
1. How to use `testcontainers` directly
2. Reference to the ready-to-use modules 

Please check if it sounds good to you

thomaseizinger (approved):
Awesome, thanks!

PR #529, opened 2023-10-21
Added health-cmd definition

dlen:
Hello! Thanks for the project, love it!

I have been working in [adding mysql image to the community modules](https://github.com/testcontainers/testcontainers-rs-modules-community/pull/50) and the basic WaitFor checks on stderr output are not enough and using a duration is a bit flacky.

I noticed that there is no way to specify a `health-cmd` as part of the container arguments so I added it. Many docker images would benefit from custom healthchecks.

I had to include curl as dependency for one of the testing images that maybe you prefer to do in a different way. In any case let me know what you think.

### DIFF
--- a/testcontainers/tests/cli_client.rs
+++ b/testcontainers/tests/cli_client.rs
@@ -97,3 +97,22 @@ fn generic_image_port_not_exposed() {
     // Without exposing the port with `with_exposed_port()`, we cannot get a mapping to it.
     node.get_host_port_ipv4(target_port);
 }
+
+#[test]
+fn generic_image_with_custom_healthcheck_cmd() {
+    let docker = clients::Cli::default();
+
+    let generic = GenericImage::new("simple_web_server", "latest")
+        .with_wait_for(WaitFor::Healthcheck)
+        .with_health_cmd("curl --fail http://127.0.0.1 || exit 1");
+
+    let node = docker.run(generic);
+    let port = node.get_host_port_ipv4(80);
+    assert_eq!(
+        "foo",
+        reqwest::blocking::get(format!("http://127.0.0.1:{port}"))
+            .unwrap()
+            .text()
+            .unwrap()
+    );
+}
--- a/testcontainers/src/clients/cli.rs
+++ b/testcontainers/src/clients/cli.rs
@@ -176,6 +176,10 @@ impl Client {
             command.arg("--entrypoint").arg(entrypoint);
         }
 
+        if let Some(health_cmd) = image.health_cmd() {
+            command.arg("--health-cmd").arg(health_cmd);
+        }
+
         let is_container_networked = image
             .network()
             .as_ref()
--- a/testcontainers/src/core/image.rs
+++ b/testcontainers/src/core/image.rs
@@ -84,6 +84,11 @@ where
         Default::default()
     }
 
+    /// Returns the healtcheck cmd this instance was created with.
+    fn health_cmd(&self) -> Option<String> {
+        None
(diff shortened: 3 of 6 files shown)


testcontainers/tests/cli_client.rs
```diff
@@ -97,3 +97,22 @@ fn generic_image_port_not_exposed() {
...
     node.get_host_port_ipv4(target_port);
 }
+
+#[test]
+fn generic_image_with_custom_healthcheck_cmd() {
+    let docker = clients::Cli::default();
+
+    let generic = GenericImage::new("simple_web_server", "latest")
+        .with_wait_for(WaitFor::Healthcheck)
+        .with_health_cmd("curl --fail http://127.0.0.1 || exit 1");
```
thomaseizinger, line 107: I wonder if we shouldn't instead add an `Option<String>` to `WaitFor::Healthcheck` where `None` means use the one defined in the image and `Some` overrides it with something else.

thomaseizinger (commented):
Thanks!

One thought.

testcontainers/tests/cli_client.rs
dlen: Yes,  Good idea! Should I add it to this PR?

dlen: Now that I think about it a little bit more `.with_health_cmd` is more in line with the way you have in place to manage the cli arguments. Adding an option to the `WaitFor::Healthcheck` would be a bit ad-hoc and maybe confusing for users. Don't you think?

thomaseizinger: So far, we've followed the philosophy of making modules self-contained. If you are creating a mysql module, it should come with the healthcheck embedded, shouldn't it? How should the user know which healthcheck to pass?

dlen: Yeah, providing a self contained container is the idea. There are base images like the mysql one that do not define any `HEALTHCHECK` instruction. Providing a cli option to inject a healthcheck command would enable the mysql testcontainer module, or any other module, to define their default healthcheck command by default, making the module self-contained. Here a sample https://github.com/testcontainers/testcontainers-rs-modules-community/pull/50/commits/ba447a1f19a00c0e12ecf5623c355d794d83c7e9

In any case if a user wants to define a `health-cmd` in their containers at runtime this feature would be useful for them IMO.

thomaseizinger: > Yeah, providing a self contained container is the idea. There are base images like the mysql one that do not define any `HEALTHCHECK` instruction. Providing a cli option to inject a healthcheck command would enable the mysql testcontainer module, or any other module, to define their default healthcheck command by default, making the module self-contained. Here a sample [testcontainers/testcontainers-rs-modules-community@ba447a1](https://github.com/testcontainers/testcontainers-rs-modules-community/commit/ba447a1f19a00c0e12ecf5623c355d794d83c7e9)

Yeah I am definitely open to supporting this. Where do you see the problem in adding this to `WaitFor::Healthcheck`?

Asked differently, what should `testcontainers` do if you specify a `health_cmd` but don't say `WaitFor::Healthcheck`? I'd consider whatever `WaitFor` is specified to take precedence and thus we'd have to ignore the `health_cmd` option which would likely confuse users.

DDtKey:
We're going to deprecate CLI client very soon, and since this PR doesn't introduce the changes of http one - I'm going to close it. Also it looks inactive. 

Thank you very much for all efforts and sorry about that 🙏

PR #531, opened 2023-10-24
Actually map exposed ports for HTTP client

Chrisss93:
When switching from the Cli client to the Http client I noticed that the expose_ports of my Image were no longer dynamically mapping to free ports on the docker daemon's host.  It seems like this is a gap to fill as a user might want a specific container port (that the image manifest may or may not have an EXPOSE layer for) mapped to a free port on the daemon host.

### DIFF
--- a/testcontainers/src/clients/http.rs
+++ b/testcontainers/src/clients/http.rs
@@ -1,5 +1,5 @@
 use crate::{
-    core::{env, logs::LogStreamAsync, ports::Ports, DockerAsync},
+    core::{env, logs::LogStreamAsync, ports::Ports, DockerAsync, Port},
     ContainerAsync, Image, ImageArgs, RunnableImage,
 };
 use async_trait::async_trait;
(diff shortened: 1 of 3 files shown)


Chrisss93:
Hi @thomaseizinger , if you have a free moment, could you let me know if this is a change that might be useful for the experimental client? Many thanks!

testcontainers/src/clients/http.rs
```diff
@@ -372,6 +375,18 @@ mod tests {
...
+    #[tokio::test(flavor = "multi_thread")]
+    async fn api_run_command_should_map_exposed_port() {
+        let docker = Http::new();
+        let image = GenericImage::new("simple_web_server", "latest").with_exposed_port(5000);
+        let container = docker.run(image).await;
+
+        // inspect volume and env
+        let container_details = inspect(&docker.inner.bollard, container.id()).await;
+        let ports = container_details.network_settings.unwrap().ports.unwrap();
+        assert!(ports.contains_key("5000/tcp"), "{:?}", ports);
```
thomaseizinger, line 387: Would it perhaps make more sense to use the public API of the container to test this?

thomaseizinger (commented):
This looks like an oversight yes, thanks for addressing it. One question.

testcontainers/src/clients/http.rs
Chrisss93: Yes that makes far more sense. I've amended the test :)

thomaseizinger:
Can you deal with the CI failurse?

thomaseizinger:
Still some failures unfortunately.

PR #534, opened 2023-10-31
Remove Container lifetime

ufoscout:
The lifetime on the `Container` struct is useless because the container owns the Cli, so Rust guarantees that the client cannot be dropped while the container is in scope.
This makes also the Container Send and Sync which can be useful in some complex test scenarios

### DIFF
--- a/testcontainers/src/clients/cli.rs
+++ b/testcontainers/src/clients/cli.rs
@@ -24,7 +24,7 @@ pub struct Cli {
 }
 
 impl Cli {
-    pub fn run<I: Image>(&self, image: impl Into<RunnableImage<I>>) -> Container<'_, I> {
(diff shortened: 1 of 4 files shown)


thomaseizinger:
I agree that this lifetime should go away. The discussion around that is in https://github.com/testcontainers/testcontainers-rs/issues/386.

I don't want to just remove it without the requirements around creation and removal of networks being properly addressed.

DDtKey:
This PR hides the fact of dependency between Client and Container.
Any networks created for container are being removed only on drop of client. 

I also agree that we need to rework this, but not this way.

I'm going to close the PR, because another reason is that we going to deprecate Cli client and probably re-work HTTP one.

Don't hesitate to continue discussion in the related issue. We will figure out optimal way to make it reliable.
Thank you!

PR #536, opened 2023-11-16
Fix race condition when creating docker network

stepantubanov:
When multiple threads are starting containers with the same network, `docker network create` call often fails due to network already existing. I've added an extra check to fix that.

EDIT: Seems like I'm still getting errors, so I'm going to test it a bit more.


stepantubanov:
Seems like there's a race condition in the docker daemon itself because I'm getting 2 networks with the same name, so this fix doesn't really help.

Issue #347, opened 2022-03-24
Looking for maintainers


thomaseizinger:
Help in maintaining this project would be greatly appreciated.

Please comment if you are interested or ping me on matrix: https://matrix.to/#/@thomaseizinger:matrix.org

marlon-sousa:
Hello,

Could you elaborate on what is involved omn this?

As far as I understand the infrastructure so to say for the project is pretty much done, with either new images or small improvement on existing images functionality being the most likely features to be integrated.

For example, I need rabbitMQ for my project and as far as I could understand image is on master but not yet published.

A simple question though is this: how can I help to publish this one to crates.io (e.e what is involved on that process)?

May be by specifing small baby steps you can find a group of people to help.

I see rust as a brilliant language not only for systems development but also for microservices and other domains. Projects such as this one and cucumber, as well as other testing frameworks and also mock servers and such are key to help the eco system to grow and make it easier for new adopters to come. I am getting very afraid that this risks not succeeding at all because so many of these important projects did not see any release in a long period.

Not to blame you, everyone has their lifes, question here is wht can we do to help to publish at least, as a first step, stuff already on master as a first step.

thomaseizinger:
I am actually just working on publishing a release :)

We are using git flow in this repository, meaning releases are cut by branching off a `release/x.y.z` branch from `dev` and pointing it to `master`. Once merged into `master`, a release will automatically be uploaded to crates.io!

See the release PR here: https://github.com/testcontainers/testcontainers-rs/pull/350

> As far as I understand the infrastructure so to say for the project is pretty much done, with either new images or small improvement on existing images functionality being the most likely features to be integrated.

More or less. The latest release includes a massive re-write of many parts that should make it much easier to define new images outside of this repo, i.e. the boilerplate required goes down significantly. That work is not fully complete yet, there is still more that can be done :)

In the last release, I've also tried to trim down the public API to make breaking changes less likely (so we can eventually comfortably hit 1.0.0). I haven't looked into things in a while but I expect that we can also do more on that front!

mervyn-mccreight:
I'd love to volunteer for this - is there any "official" communication channel for this project where I'd have to join?

thomaseizinger:
There is no official communication channel other than Github issues. I guess I could create a Matrix room :)

thomaseizinger:
I've created a matrix room: https://matrix.to/#/#testcontainers-rs:matrix.org

kiview:
[...]

PR #539, opened 2023-11-24
Add hostname support

maexn:
Adding the support for the docker `--hostname` option

### DIFF
--- a/testcontainers/src/clients/cli.rs
+++ b/testcontainers/src/clients/cli.rs
@@ -160,6 +160,10 @@ impl Client {
             command.arg(format!("--network={network}"));
         }
 
+        if let Some(hostname) = image.hostname() {
+            command.arg(format!("--hostname={hostname}"));
+        }
+
         if let Some(name) = image.container_name() {
             command.arg(format!("--name={name}"));
         }
@@ -574,6 +578,18 @@ mod tests {
         assert_eq!(
             format!("{command:?}"),
             r#""docker" "run" "--network=awesome-net" "-P" "-d" "hello:0.0""#
+        );
+    }
+
+    #[test]
+    fn cli_run_command_should_include_hostname() {
+        let image = GenericImage::new("hello", "0.0");
+        let image = RunnableImage::from(image).with_hostname("what-a-host");
+        let command = Client::build_run_command(&image, Command::new("docker"));
+
+        assert_eq!(
+            format!("{command:?}"),
+            r#""docker" "run" "--hostname=hosti" "-P" "-d" "hello:0.0""#
         );
     }
 
--- a/testcontainers/src/clients/http.rs
+++ b/testcontainers/src/clients/http.rs
@@ -80,6 +80,14 @@ impl Http {
             }
         }
 
+        // hostname
+        if let Some(hostname) = image.hostname() {
+            config.host_config = config.host_config.map(|mut host_config| {
+                host_config.hostname = Some(hostname.to_string());
+                host_config
+            });
+        }
+
         // name of the container
         if let Some(name) = image.container_name() {
             create_options = Some(CreateContainerOptions {
--- a/testcontainers/src/core/image.rs
+++ b/testcontainers/src/core/image.rs
@@ -152,6 +152,7 @@ pub struct RunnableImage<I: Image> {
     image_tag: Option<String>,
     container_name: Option<String>,
     network: Option<String>,
+    hostname: Option<String>,
     env_vars: BTreeMap<String, String>,
     volumes: BTreeMap<String, String>,
     ports: Option<Vec<Port>>,
@@ -170,6 +171,10 @@ impl<I: Image> RunnableImage<I> {
 
     pub fn network(&self) -> &Option<String> {
         &self.network
+    }
+
+    pub fn hostname(&self) -> &Option<String> {
+        &self.hostname
     }
 
     pub fn container_name(&self) -> &Option<String> {
@@ -246,6 +251,13 @@ impl<I: Image> RunnableImage<I> {
         }
     }
 
+    pub fn with_hostname(self, hostname: impl Into<String>) -> Self {
+        Self {
+            hostname: Some(hostname.into()),
+            ..self
+        }
+    }
+
     pub fn with_env_var(self, (key, value): (impl Into<String>, impl Into<String>)) -> Self {
         let mut env_vars = self.env_vars;
         env_vars.insert(key.into(), value.into());
@@ -298,6 +310,7 @@ impl<I: Image> From<(I, I::Args)> for RunnableImage<I> {
             image_tag: None,
             container_name: None,
             network: None,
+            hostname: None,
             env_vars: BTreeMap::default(),
             volumes: BTreeMap::default(),
             ports: None,


thomaseizinger (commented):
Thanks! Can you explain why you need this in your test setup?

maexn:
Hi @thomaseizinger 

One of the containers in my test setup requires a preset hostname, localhost cannot be used due to the communication between the containers. 
I think Docker's hostname flag would be a nice and lean solution to this problem.

thomaseizinger:
What if you set the name of the container? Does that work for your usecase?

silwol:
Working on the same team as @maexn I think we should elaborate our use case a little further.

One of the containers we start with `testcontainers-rs` contains an instance of [keycloak](https://keycloak.org). Apparently keycloak inserts its own hostname into the id tokens that it hands out to clients.

We want to be able to either run the composition on containers fully inside our automated tests, or to start the same composition interactively for debugging purposes, accessing them from the host through the web frontend of our software.

Because keycloak encodes the hostname into the id tokens, the hostname must be the same in both scenarios. This can be achieved by the workaround of using e.g. `keycloak.localhost`, which resolves to `127.0.0.1` outside the container, and to the other container's ip when used inside. This is what the change enables us to do, but can not be implemented with current stock `testcontainers-rs`.

If you know a different method that allows us to reach the same goal, we'd be happy to use that instead, but given the fact that this is a standard docker feature, I think supporting it makes sense.

silwol:
> How are you using `testcontainers` for this? This sounds like a usecase for `docker-compose`!

We're actually migrating away from `docker-compose` for this use case.

The idea is to have several scenarios available that can be used for tests. The service can be started in multiple instances which coordinate through RabbitMQ and Redis, and include an OIDC provider. Example scenarios:
- Single instance, with example users **Alice**, **Bob**, **Charlie**, **Dave** and **Erin**
- Multiple instances, randomly generated users and simulated mass usage
- Multiple instances, and during the test we restart RabbitMQ or Redis, and verify that the services recover properly
- etc

The use case for interactively debugging the setup comes in whenever a test fails, and you want to be able to analyze why. Then you can execute that single test and keep the `testcontainers` composition running if a specific environment variable is set, and investigate the situation.

Drawbacks we had with `docker-compose`, which is why we started looking around and found `testcontainers` which we really appreciate for that use case:
- `docker-compose` is much more static, starting multiple scenarios in parallel requires a lot of thought to be put into due to fixed ports
- Scenarios of restarting services to test service robustness are much more difficult to implement
- Integration in test code is a lot more verbose, `testcontainers` allows us to provide abstractions a lot easier

thomaseizinger:
Thank you for sharing that!

I am still not sure if I fully understand your usecase. The idea of testcontainers (and the design of `testcontainers-rs` specifically) is to make integration tests of individual components possible. For example if you have a data-access-layer, you want to test your SQL queries against a real database.

This is the usecase I want to serve well with `testcontainers` and until [another maintainer steps up](https://github.com/testcontainers/testcontainers-rs/issues/347), I am somewhat reluctant to repurpose it into something else (like a replacement for docker swarm or docker compose).

> `docker-compose` is much more static, starting multiple scenarios in parallel requires a lot of thought to be put into due to fixed ports

If everything is running in docker, you don't need to expose any host ports so you don't have conflicts.

> Scenarios of restarting services to test service robustness are much more difficult to implement

I am surprised you say that. I've had very good experiences with using `docker-compose` for exactly that:

- Start services with `docker compose up -d`
- Check connectivity: `docker compose -it exec $some_command`
- Stop one service: `docker compose stop some_service`
- Check connectivity again: `docker compose -it exec $some_command`

silwol:
Thanks for taking the time to write such a detailed answer.

> I am still not sure if I fully understand your usecase. The idea of testcontainers (and the design of `testcontainers-rs` specifically) is to make integration tests of individual components possible. For example if you have a data-access-layer, you want to test your SQL queries against a real database.

Ok, if that's how it is intended to be used, we're probably aiming for something that's beyond this scope. The use case we have in mind is to test the full service in interaction with other components. The service and the other components are started using `testcontainers-rs` inside the each test, then the action is performed (requests are sent and their answers verified) and the expected behavior of the service in this scenario is verified.

> This is the usecase I want to serve well with `testcontainers` and until [another maintainer steps up](https://github.com/testcontainers/testcontainers-rs/issues/347), I am somewhat reluctant to repurpose it into something else (like a replacement for docker swarm or docker compose).

Totally understandable, I recognize this situation just too well.

> > `docker-compose` is much more static, starting multiple scenarios in parallel requires a lot of thought to be put into due to fixed ports
> 
> If everything is running in docker, you don't need to expose any host ports so you don't have conflicts.

Yes, as explained above, the test itself sends requests to the "containered" service, and that runs on the host.

> > Scenarios of restarting services to test service robustness are much more difficult to implement
> 
> I am surprised you say that. I've had very good experiences with using `docker-compose` for exactly that:
> 
>     * Start services with `docker compose up -d`
> 
>     * Check connectivity: `docker compose -it exec $some_command`
> 
>     * Stop one service: `docker compose stop some_service`
> 
>     * Check connectivity again: `docker compose -it exec $some_command`

As explained above, the checks are performed in rust code, while this sequence would fit nicely in shell scripts it would be rather cumbersome to implement in rust, at least requiring some level of abstraction in addition. We need to encapsulate that functionality in any case, and `testcontainers-rs` seemed like a nice fit for that.

> `testcontainers` is never going to be a full purpose docker client. The codebase itself is really small and if I understand you correctly in what you want to achieve, I'd suggest building your own integration test harness instead of extending this one with configuration options. `testcontainers` has a very permissive license so you can take whatever code you need from it :)

Thanks a lot! We'll think about how to proceed here. The integration tests dev branch we currently have is actually running quite nicely with just this - as we perceive - rather small improvement to `testcontainers-rs`. Our code will be published as soon as the dev branch is merged. Sadly our GitLab instance is not open to the public, so dev branches cannot be seen until merged to `main` and [published to OpenCoDE](https://gitlab.opencode.de/opentalk/controller).

Feel free to proceed with this PR as desired, we will consider our options for achieving what we need. `testcontainers-rs` proved quite reliable and promising for the use case otherwise.

thomaseizinger:
> The integration tests dev branch we currently have is actually running quite nicely with just this - as we perceive - rather small improvement to `testcontainers-rs`

I agree that it would be a small addition but we have to draw the line somewhere. docker has lots of config options because it servers many usecases. `testcontainers-rs` does not serve all docker usecases. We already support booting up multiple containers and attaching it to the same network. Docker will default the hostname to the container name so from the perspective of this library, you can actually set the hostname.

> > > Scenarios of restarting services to test service robustness are much more difficult to implement
> > 
> > I am surprised you say that. I've had very good experiences with using `docker-compose` for exactly that:
> > 
> > ```
> > * Start services with `docker compose up -d`
> > 
> > * Check connectivity: `docker compose -it exec $some_command`
> > 
> > * Stop one service: `docker compose stop some_service`
> > 
> > * Check connectivity again: `docker compose -it exec $some_command`
> > ```
>
> As explained above, the checks are performed in rust code, while this sequence would fit nicely in shell scripts it would be rather cumbersome to implement in rust, at least requiring some level of abstraction in addition. We need to encapsulate that functionality in any case, and testcontainers-rs seemed like a nice fit for that.

Even though I've not seen your code, I have a hunch of what it might look like because I've written similar tests using `testcontainers-rs` before and let me tell you, it didn't turn out well.

Good tests survive big refactors because they are testing an interface that remains stable. In my experience, automated tests at a level that involve booting up other services are best done using an interface that is also used by users (either through the UI or via an API). That is what I meant with `docker compose -it exec $some_command`. Even if that is not possible for some reason, it might be a lot easier to write yourself a small integration-test client that provides programmatic access to internal APIs, basically, a layer above what you are currently doing.

thomaseizinger:
In any case, thank you for the contribution and explaining the usecase. Unfortunately, we can't accept feature request.

thomaseizinger:
> basically, a layer above what you are currently doing.

What I've found is that in Rust, the typical test-pyramid does not apply if you use the type system properly. In fact, you probably want a fair few end-to-end tests, no integration tests and lots of unit tests (unit meaning a particular, isolated unit of code which can be almost arbitrarily big).

The "integration" test is essentially done by the type checker because it ensures you are using all variables, exhaustively matching on all enum variants, etc. Thus the data-flow is guaranteed to handle all cases. Corner-cases can be tested using unit-tests. Thus, if your e2e tests pass (which implies a passing build), you are covering a lot of ground already. Don't forget that tests also have a (maintenance) cost. Every test you don't have to write is literally saving you money.

`testcontainers-rs` fills a niche that allows you to unit-test components that depend on external services to make sure those edge-cases are covered too as it is often hard to get coverage on that via e2e tests.

silwol:
> > basically, a layer above what you are currently doing.
> 
> What I've found is that in Rust, the typical test-pyramid does not apply if you use the type system properly. In fact, you probably want a fair few end-to-end tests, no integration tests and lots of unit tests (unit meaning a particular, isolated unit of code which can be almost arbitrarily big).

Yes, that's probably true. We currently have a lot of what we call "unit tests", which cover both traditional isolated units of code, and what you call end-to-end tests here, but we don't distinguish these. That's why we are looking into splitting those up into two separate sections, this is where `testcontainers-rs` came in.

The reason why I call them integration tests is because we have a web frontend as well, so the naming "end-to-end tests" is already occupied in our project by tests covering the frontend as well. And yes, at every single employer I had in the past, there was the discussion about which kinds of tests there are and should be, and how they should be named.

Thanks for concluding this ticket, as remarked above we'll investigate our options and decide based on the outcome which direction we will choose. Way better than just remaining silent. :+1:

PR #547, opened 2024-02-07
feat(cli): support for docker cli flag `--add-host`

amountainram:
When spawning multiple docker containers in the same test, communication amongst them can be achieved over the docker default gateway.

In DinD environments, such as when running jobs on a CI pipeline, the docker [`--add-host`](https://docs.docker.com/engine/reference/commandline/container_run/#option) can be employed to basically enter new entries in the `/etc/hosts` files. This allows to avoid binding all container ports to the host environment.

The flag API counterpart is implemented on the `RunnableImage` struct instead of on the generic image since I reckon it makes more sense within this crate phylosophy, I hope !-).

A test runs 2 containers:

- an http_server
- a `curl` client

and the `curl` client completes when receiving `foo` from the server on the alias hostname. In case the hostname wasn't resolve it'd panic.

### DIFF
--- a/testcontainers/src/core/image.rs
+++ b/testcontainers/src/core/image.rs
@@ -1,4 +1,10 @@
-use std::{collections::BTreeMap, env::var, fmt::Debug, time::Duration};
+use std::{
+    collections::BTreeMap,
+    env::var,
+    fmt::{Debug, Display},
+    net::IpAddr,
+    time::Duration,
+};
 
 use super::ports::Ports;
 
@@ -144,6 +150,21 @@ impl ImageArgs for () {
     }
 }
 
+#[derive(Debug)]
+pub enum Host {
+    Addr(IpAddr),
+    HostGateway,
+}
+
+impl Display for Host {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Host::Addr(addr) => write!(f, "{addr}"),
+            Host::HostGateway => write!(f, "host-gateway"),
+        }
+    }
+}
+
 #[must_use]
 #[derive(Debug)]
 pub struct RunnableImage<I: Image> {
@@ -153,6 +174,7 @@ pub struct RunnableImage<I: Image> {
     container_name: Option<String>,
     network: Option<String>,
     env_vars: BTreeMap<String, String>,
+    hosts: BTreeMap<String, Host>,
     volumes: BTreeMap<String, String>,
     ports: Option<Vec<Port>>,
     privileged: bool,
@@ -178,6 +200,10 @@ impl<I: Image> RunnableImage<I> {
 
     pub fn env_vars(&self) -> Box<dyn Iterator<Item = (&String, &String)> + '_> {
         Box::new(self.image.env_vars().chain(self.env_vars.iter()))
+    }
+
+    pub fn hosts(&self) -> Box<dyn Iterator<Item = (&String, &Host)> + '_> {
+        Box::new(self.hosts.iter())
     }
 
     pub fn volumes(&self) -> Box<dyn Iterator<Item = (&String, &String)> + '_> {
@@ -252,6 +278,12 @@ impl<I: Image> RunnableImage<I> {
         Self { env_vars, ..self }
     }
 
+    pub fn with_host(self, key: impl Into<String>, value: impl Into<Host>) -> Self {
+        let mut hosts = self.hosts;
+        hosts.insert(key.into(), value.into());
+        Self { hosts, ..self }
+    }
+
     pub fn with_volume(self, (orig, dest): (impl Into<String>, impl Into<String>)) -> Self {
         let mut volumes = self.volumes;
         volumes.insert(orig.into(), dest.into());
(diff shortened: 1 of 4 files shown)


testcontainers/src/core/image.rs
```diff
@@ -252,6 +257,12 @@ impl<I: Image> RunnableImage<I> {
         Self { env_vars, ..self }
     }
 
+    pub fn with_add_host(self, (key, value): (impl Into<String>, impl Into<String>)) -> Self {
```
thomaseizinger, line 260: This would be more idiomatic as two parameters.

thomaseizinger: And we should probably call it `with_host`.

testcontainers/src/core/image.rs
```diff
@@ -180,6 +181,10 @@ impl<I: Image> RunnableImage<I> {
         Box::new(self.image.env_vars().chain(self.env_vars.iter()))
     }
 
+    pub fn add_hosts(&self) -> Box<dyn Iterator<Item = (&String, &String)> + '_> {
```
thomaseizinger, line 184:
```suggestion
    pub fn hosts(&self) -> Box<dyn Iterator<Item = (&String, &String)> + '_> {
```

testcontainers/tests/cli_client.rs
```diff
@@ -81,6 +86,28 @@ fn generic_image_exposed_ports() {
...
+
+    let msg = WaitFor::message_on_stdout("foo");
+    let server_2 = GenericImage::new("curlimages/curl", "latest")
+        .with_wait_for(msg.clone())
+        .with_entrypoint("curl");
+
+    // Override hosts for server_2 adding
+    // custom-host as an alias for localhost
+    let server_2 = RunnableImage::from((server_2, vec![format!("http://custom-host:{port}")]))
+        .with_add_host(("custom-host", "host-gateway"));
```
thomaseizinger, line 106: What does `host-gateway` mean? Is that a docker thing?

thomaseizinger (commented):
Makes sense to me!

A few comments.

testcontainers/tests/cli_client.rs
amountainram: yup. it's a [special value](https://docs.docker.com/engine/reference/commandline/dockerd/#configure-host-gateway-ip) replacing a valid IP that can be used to refer to the host machine gateway IP

thomaseizinger: Hmm, can we test this somehow differently? It appears that usually, the value here should be an `IpAddr`: https://docs.docker.com/engine/reference/commandline/image_build/#add-host

Perhaps we can make it an enum?

```rust
enum Host {
	Addr(IpAddr),
	HostGateway
}
```

Stringly-typed APIs aren't very user-friendly.

thomaseizinger: > where would you place the enum? (i guess in `src/core/image.rs`)

Put it next to the API that uses it :)

> do you prefer i proceed keeping one single commit amending previous history or multiple commits to keep track of reviews/discussion?

Don't care really but preference for no force-pushing so I can see the diff between the reviews.

> about the test. any of these variant should cut it (and they do on my machine):

I think once we have the enum, that is good enough and you don't need to change the test. I just didn't want to have a stringly-typed API :)

amountainram: all good then 👍🏻!

amountainram:
@thomaseizinger I added the enum as requested and updated the code

testcontainers/tests/cli_client.rs
```diff
@@ -81,6 +84,28 @@ fn generic_image_exposed_ports() {
         .is_success());
 }
 
+#[test]
+fn generic_image_running_with_extra_hosts_added() {
+    let docker = clients::Cli::default();
+    let msg = WaitFor::message_on_stdout("server is ready");
+
+    let server_1 = GenericImage::new("simple_web_server", "latest").with_wait_for(msg.clone());
```
thomaseizinger, line 92: This is duplicated several times in these tests now, can we fix that by extracting a function that gives us a `GenericImage` for the `simple_web_server`?

thomaseizinger (commented):
One more request, otherwise LGTM

thomaseizinger (approved):

amountainram:
@thomaseizinger i fixed the lint issue but i'm kinda unsure on what to do for the tests.
it appears from the error `invalid argument "custom-host=host-gateway" for "--add-host" flag: bad format for add-host: "custom-host=host-gateway"` that the supported syntax is `{hostname}:{remote-host}` instead of `{hostname}={remote-host}`.

I've tried to run the CI locally with [act](https://github.com/nektos/act) but I could not make it work.

Can we run again with the new syntax for `--add-host` flag?

thomaseizinger:
> I've tried to run the CI locally with [act](https://github.com/nektos/act) but I could not make it work.

You should just be able to run `cargo test` locally, assuming you have docker!

amountainram:
> > I've tried to run the CI locally with [act](https://github.com/nektos/act) but I could not make it work.
> 
> You should just be able to run `cargo test` locally, assuming you have docker!

Sure thing.. that I did and tests passed both with the syntax `host=host` and `host:host` but here in github CI the Docker REST API is used instead of the unix socket to talk with the Docker daemon. It appears that such API understands only `host:host` for the `--add-host` flag.

Anyway I see tests pass now so all good 😄!

Issue #335, opened 2022-02-23
Allow custom docker registry


mervyn-mccreight:
I think it would be nice to allow some configuration for a custom docker registry to pull the images from.
I'm thinking about a general point of configuration.

This is useful if e.g. you are working in a company that relies on an internal docker repository (e.g. as a caching solution).

thomaseizinger:
Agreed.

I think the best way to extend the current codebase here is to have an optional field in the docker client and produce FQ container names in case they aren't already fully-qualified.

There may also be a config in the CLI that we can use.

mervyn-mccreight:
Cool, I think I'll give it a try 😃

DDtKey:
Actually I might want to use the same client against different registries, moreover the name of images may vary between registries. For example, AWS ECR for docker library would be: `https://gallery.ecr.aws/docker/library/redis`, but not all images from docker-hub are available there and not all of them have the same prefix (e.g bitnami images https://gallery.ecr.aws/bitnami/kafka)

Currently it's enough to set `name` for container to necessary one (e.g`public.ecr.aws/docker/library/redis` or `public.ecr.aws/bitnami/kafka`) 

But we don't expose the way to override the name through `RunnableImage` currently. So I used to re-define images locally with custom `name` for such purposes.

I think the correct way is to allow overriding the name as well as `tag` (already possible) as part of `RunnableImage`. WDYT?

thomaseizinger:
> But we don't expose the way to override the name through `RunnableImage` currently.

Yeah and there is a good reason for that! The `Image` trait is what ties together the image with its version and configuration. That creates a type-safe contract for a working container. Adding the ability to replace the image with a completely different one breaks that contract.

DDtKey:
Yeah, we do allow tag to be overridden and there is a comment that it may broke the image.

What do you think about the second solution? We could allow to change only registry and/or owner in `RunnableImage`. However it's basically the same: changing the registry also may lead to the failure, for example - the image with another hash is kept for the tag

Anyway by redefining the name/registry, I clearly take responsibility on myself.

Btw, looks like it's allowed in other languages: https://java.testcontainers.org/features/image_name_substitution/

DDtKey:
Moreover, nobody prevents us to expose API change the name on `Image` level (like `MyImageX:new().with_name("another")` - testcontainers can't prevent this, people decide what to implement)

[...]

PR #549, opened 2024-02-18
feat: add `name` parameter to `RunnableImage`

DDtKey:
This change allows to override name of the image. Thus, user will be able to change registry, owner and the name itself.
See https://github.com/testcontainers/testcontainers-rs/issues/335#issuecomment-1951399972 for more details

Closes https://github.com/testcontainers/testcontainers-rs/issues/335

### DIFF
--- a/testcontainers/src/core/image.rs
+++ b/testcontainers/src/core/image.rs
@@ -233,7 +233,7 @@ impl<I: Image> RunnableImage<I> {
         let original_tag = self.image.tag();
 
         let name = self.image_name.as_ref().unwrap_or(&original_name);
-        let tag = self.image_name.as_ref().unwrap_or(&original_tag);
+        let tag = self.image_tag.as_ref().unwrap_or(&original_tag);
 
         format!("{name}:{tag}")
     }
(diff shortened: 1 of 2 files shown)


testcontainers/src/core/image.rs
```diff
@@ -149,6 +149,7 @@ impl ImageArgs for () {
 pub struct RunnableImage<I: Image> {
     image: I,
     image_args: I::Args,
+    image_name: Option<String>,
```
DDtKey, line 152: Or, as an alternative, we could add optional `regitstry`, and prepend it for descriptor.

But since we don't have such config on `Image` level - it seems unreasonable without changing `Image` API

DDtKey: Added PR with alternative solution https://github.com/testcontainers/testcontainers-rs/pull/550,

WDYT @thomaseizinger? Which one do you prefer (if any)? My vision is described within #335 issue

thomaseizinger: I've responded in the issue.

PR #556, opened 2024-03-15
Owen/ext container run options

yubing744:
Provide extension points to facilitate users to add other container run parameters, such as --network-alias and --sig-proxy, etc.


DDtKey:
Thanks for the contribution!

We're going to deprecate CLI client and put efforts to re-work http one (#563)
So I'm closing this PR for now and we can consider alternatives later after the project cleanup

PR #558, opened 2024-03-16
feat: allow to override args in a `RunnableImage`

### LINKED ISSUE #554: Support overriding "command" for Image
Current Image supports overriding `entrypoint`, and `env_vars`. I also need to override `command`. 

Use-case is - I want to run multiple instances of the same server (on different ports), each will have `--name={1,2,3}` additional flag. 

With current `type Args;` implementation I can only hardcode a single "default" `command`, but cannot dynamically append new args directly in the test.

jorgehermo9:
Closes #554. This PR allows to an alternative way of creating a `RunnableImage` with other than default args. This can already be done with a tuple, but it is kind of "hidden" https://github.com/testcontainers/testcontainers-rs/blob/71bbdec013c1b19e5507217d52c612c71d9473f5/testcontainers/src/core/image.rs#L325

What could be done before:
```rust
let image = (GenericImage,vec!["one-arg","another-arg"]);
let runnable_image = RunnableImage::from(image)
```

Now:

```rust
let runnable_image = RunnableImage::from(GenericImage).with_args(vec!["one-arg","another-arg"]);
```

Furthermore, it is more consistent now, since all of the `RunnableImage` attributes but the `image_args` could be modified using a `with_*` method.


Warchant (approved):
👍

testcontainers/src/core/image.rs
```diff
@@ -249,6 +249,13 @@ impl<I: Image> RunnableImage<I> {
 }
 
 impl<I: Image> RunnableImage<I> {
+    pub fn with_args(self, args: I::Args) -> Self {
```
DDtKey, line 252: Let's add a documentation?

jorgehermo9: Sure! Gonna commit asap😁

jorgehermo9: done, what do you think about the examples? @DDtKey

DDtKey: Looks good to me, however I think the more common use-case would be a builder-like usage, i.e:

```rs
let args = vec!["arg1".to_string(), "arg2".to_string()];
let runnable_image = RunnableImage::from(image.clone())
   .with_args(args.clone());
```

This is generally a more concise syntax

jorgehermo9: updated with that syntax! I like it more as you proposed, sorry for the delay😁

jorgehermo9: Fixed dprint check issues in the CI, passing all checks now :smiley:

DDtKey:
Thank you!

jorgehermo9:
Thank you for your review and the quick responses! It's lovely to see the project is active again. Maybe I'll try to get more involved in it😊

Issue #561, opened 2024-03-22
Mapped ports are not updated after container restart


stormshield-gt:
Since https://github.com/testcontainers/testcontainers-rs/pull/184 we can restart a container. But if we have mapped ports, the value returned by `get_host_port_ipv4(8080)` is not updated.

The current workaround I found is the following, but it's not ideal:

```rust
 let host_port = vault.get_host_port_ipv4(8080) + 1;
```

DDtKey:
Gonna be solved with #392

Issue #392, opened 2022-08-19
Ports shouldn't be cached


schrieveslaach:
Since commit [`88e18b44`](https://github.com/testcontainers/testcontainers-rs/commit/88e18b44d9a6b783e06a66a89819f2dee696e291#diff-51fffd26ca8ee51cf68736432e6066a4d4609544e7675e07eda499c38a10f1ddR66) ports will be cached. However, ports can actually change if a container is assigned to a new network.

You can reproduce this issue by checking out [this commit](https://github.com/aixigo/PREvant/commit/fbdb4615974a1808ea0e5741a57b048a5f588e73) of the [aixigo/PREvant](https://github.com/aixigo/PREvant) respository and you can run `cargo test nginx --manifest-path api-tests/Cargo.toml -- --test-threads=1`.

During the test the traefic container will be connected to the network of an nginx container. This can cause that exposed port mapping will change.

To workaround this issue I have to map the port explicitly:

```rust
let traefik = docker.run(RunnableImage::from(Traefik::default()).with_mapped_port((9789, 80)));
```

schrieveslaach:
Workaround is included in https://github.com/aixigo/PREvant/pull/104. Search for comments mentioning the workaround. You can remove it to reproduce.

thomaseizinger:
Thanks for opening an issue!

I don't recall whether this on purpose or just an oversight. @yottabytt can you add something to this?
I think it should be fine to change this back to fetch the ports on demand. For `ContainerAsync`, we actually still do this.

schrieveslaach:
Thanks for fixing. Works with https://github.com/aixigo/PREvant/commit/22cc416b49939b74f5212bbe580f2da7ecb313db

PR #562, opened 2024-03-22
fix!: update ports on container restart

stormshield-gt:
fix https://github.com/testcontainers/testcontainers-rs/issues/561

### DIFF
--- a/testcontainers/src/core/container.rs
+++ b/testcontainers/src/core/container.rs
@@ -222,8 +222,9 @@ where
         self.docker_client.stop(&self.id)
     }
 
-    pub fn start(&self) {
+    pub fn start(&mut self) {
         self.docker_client.start(&self.id);
+        self.ports = self.docker_client.ports(&self.id);
     }
 
     pub fn rm(&self) {


testcontainers/src/core/container.rs
```diff
@@ -222,8 +222,9 @@ where
         self.docker_client.stop(&self.id)
     }
 
-    pub fn start(&self) {
+    pub fn start(&mut self) {
```
DDtKey, line 225: It's definitely breaking change fix, because public interface has been changed.

In order to avoid this I'd consider interior mutability for `ports` instead

DDtKey: However, I think it's not really safe to introduce interior mutability here. It's in general would be not really a safe API.

Thus mutable reference makes sense.

stormshield-gt: As this crate hasn't reached 1.0 yet maybe we should just do the breaking change and wait for the next release (0.16.0) to publish it.

DDtKey (approved):

DDtKey:
Actually it looks like we just don't need to cache the ports, see https://github.com/testcontainers/testcontainers-rs/issues/392#issuecomment-1220519230

I'm going to refactor this as part of new version (which addresses #563 and #386).

Thank you very much for the contribution 🙏 
But I'm going to close the PR, sorry for that

PR #570, opened 2024-04-06
build(ci): update msrv check

DDtKey:
Use cargo-hack instead of matrix to minimize number of GH jobs and make it more comprehensive (each feature is tested)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,25 +8,27 @@ on:
 
 jobs:
   msrv:
+    name: Build with MSRV
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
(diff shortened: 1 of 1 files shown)


mervyn-mccreight:
The downside of this is less parallelism isn't it? Just thinking about how this affects turnaround time of workflow runs.

DDtKey:
Yes, it may reduce parallelism.
But not necessarily, GH also has limitations of number of parallel jobs. And need to keep in mind additional time to setup the runner and install all deps.

However another benefit - it's more controllable. We can test all features independently. And it's still one isolated job. 

If we wanted to add a feature with previously existed approach - we would need to add one more GH job which is excessive imho.

Issue #563, opened 2024-03-27
Project revamp


eddumelendez:
* Deprecate `cli` support
* Support [testcontainers/ryuk](https://hub.docker.com/r/testcontainers/ryuk) as resource reaper
* Read `tc.host` from `~/.testcontainers.properties` to connect with http client. See [issue](https://github.com/fussybeaver/bollard/issues/395) related to the docker client
* Add User-Agent with value `tc-rust/<version>`, version should be the [version from Cargo.toml](https://github.com/testcontainers/testcontainers-rs/blob/dev/testcontainers/Cargo.toml#L3)
```[tasklist]
### Tasks
- [x] Deprecate `cli` client support (provide sync version with http client under the hood)
- [x] Implement docker host configuration (including tc.host)
- [ ] https://github.com/testcontainers/testcontainers-rs/issues/576
- [ ] https://github.com/testcontainers/testcontainers-rs/issues/577
```

DDtKey:
I've extracted separate issues for remaining tasks.
Main part is done within #575, so closing this issue and let's track new issues separately.

PR #575, opened 2024-04-18
feat!: container-centric API with refactored underlying layer

### LINKED ISSUE #89: Make it work with a remote docker host
The docker daemon can be hosted on a remote machine, hence in addition to the `port` of a container, we should also expose the `host` on which the docker container was started.

### LINKED ISSUE #198: Usage examples in rustdoc
Would it be possible to provide a minimal usage example directly in the rustdocs? Hints on how to get started are provided, but having a minimal complete example would go a long way (new user here).

### LINKED ISSUE #326: Use "drop-notifier" pattern for to asynchronously shut down containers with `Http` client
Not sure if this is documented anywhere but I've since learned about this very nifty pattern for doing `async` work in a `Drop` implementation:

1. Have a `Void` async, oneshot channel
2. Store the `Sender` in the struct you intend to eventually `Drop`
3. Listen on the `receiver` in a task
4. Once you get `oneshot::Canceled`, you know that the struct has been dropped
5. Perform the async work

To utilize this, we need to store the `Sender` in `ContainerAsync` and the `Receiver` in `Http`. The client (`Http`), needs to create one task per container that it starts and shut it down as soon as it receives `oneshot::Canceled`.

This should allow us to drop the `executor` feature on `futures`.

### LINKED ISSUE #386: Re-work public API to be `Container` centric

### LINKED ISSUE #392: Ports shouldn't be cached

### LINKED ISSUE #475: Sync `ContainerAsync` and `Container` api
`ContainerAsync` doesn't have `exec` & `block_until_ready` methods which is present in `Container`. 
It should be aligned

### LINKED ISSUE #507: [Question] Why would testcontainers fail to create docker network under non-root user
Hello, thanks for the crate.

We're trying to migrate our pipeline to use only non-privileged users, and we're getting:

```

---- test_set stdout ----
thread 'test_set' panicked at 'failed to create docker network', /home/user/.cargo/git/checkouts/testcontainers-rs-afc9a8ab8422234e/c4c98b1/testcontainers/src/clients/cli.rs:210:9
stack backtrace:
   0:     0x55595161192a - std::backtrace_rs::backtrace::libunwind::trace::ha271a8a7e1f3d4ef
                               at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x55595161192a - std::backtrace_rs::backtrace::trace_unsynchronized::h85739da0352c791a
                               at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x55595161192a - std::sys_common::backtrace::_print_fmt::hbc6ebcfb2910b329                
```

I've tried to debug myself, but I'm not really an expert in rust. Tried to create arbitrary docker network using a common cli on CI runner itself, and it worked without a problem, using non-root user.

Do you maybe know why would it fail?

Thanks.

### LINKED ISSUE #508: [Discussion] Reason to maintain two different implementation?
This library currently supports two way to communicate with docker;
- CLI using `docker` command
- HTTP using `bollard` (unofficial) Rust SDK for Docker 

I do think that this approach bring some unnecessary complexity without a real benefit (I would like to be corrected on this one if I missed something). Every feature has to be implemented twice with different approach, for example, sync vs async.

To my understanding, testcontainers-rs  is the only one which depends on `docker` command directly. I have been looking some of the popular implementation like java, golang and node, they all seem to take the second approach which is using SDK for Docker. 

Given the fact that wherever you have `docker` command available, I _believe_ you also have your `docker daemon` running on UNIX socket therefore there is no need to directly talk with `docker` command. (that's what `docker` command talks to as well if I am not mistaken)

I would assume that CLI implementation is around for _historical reasons_ but I don't see any benefits of keeping it around in the long term.

Of course this is my humble opinion and I would like to hear what others think. :) 

cc @thomaseizinger

### LINKED ISSUE #538: Feature: Retries due to periodic failure of underlying `docker` commands (ex. `rm`)?
As always thanks for the awesome library, it's been incredibly useful for testing.

I've been doing some stress-testing on my test suite (i.e. running the tests continuously until one failed) lately and found that sometimes the `Cli` actually fails to perform some lower level `docker` CLI commands.

The first failure I encountered was a failure with creating a container, but unfortunately I didn't have `--nocapture` on, so I couldn't get the output. After repeating the process I found that I got a failure:

```
.......... TERMINATING [>120.000s] project::mod components::mod::inner::test_name_ci_serial
thread '<unnamed>' panicked at /path/to/.cargo/registry/src/index.crates.io-6f17d22bba15001f/testcontainers-0.15.0/src/clients/cli.rs:354:9:
Failed to remove docker container
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'components::mod::inner::test_name_ci_serial' panicked at /path/to/.cargo/registry/src/index.crates.io-6f17d22bba15001f/testcontainers-0.15.0/src/clients/cli.rs:354:9:
Failed to remove docker container
test components::mod::inner::test_name_ci_serial ... FAILED

failures:

failures:
    components::mod::inner::test_name_ci_serial

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 10 filtered out; finished in 120.47s

     TIMEOUT [ 120.474s] project::mod components::mod::inner::test_name_ci_serial
   Canceling due to test failure
------------
     Summary [ 188.463s] 11 tests run: 10 passed, 1 timed out, 0 skipped
     TIMEOUT [ 120.474s] project::mod components::mod::inner::test_name_ci_serial
error: test run failed
error: Recipe `test-int` failed on line 124 with exit code 100
```

I've anonymized the details of the project and test suite, but it should be clear that the failure was inside (but not the fault of) `testcontainers`.

Looking at the output of my `docker` systemd service, I see a failure to write stderr (emphasis via spacing added below):

```
Nov 24 12:14:30 host dockerd[13873]: time="2023-11-24T12:14:30.456846562+09:00" level=info msg="ignoring event" container=444c245cd67691e5ad93decc764347c145299ffc33636300ffeff89348cdbca3 module=libcontainerd namespace=moby topic=/tasks/delete type="*events.TaskDelete"
...(many identical info messages for different containers)...

Nov 24 12:16:31 host dockerd[13873]: time="2023-11-24T12:16:31.062710838+09:00" level=error msg="Error running exec 53c30d4cedccd888d169f833fca4db8fab88d5520aadc2df6edc4affde44174b in container: exec attach failed: error attaching stderr stream: write unix /run/docker.sock->@: write: broken pipe"

Nov 24 12:16:31 host dockerd[13873]: time="2023-11-24T12:16:31.094747073+09:00" level=info msg="ignoring event" container=622b59590b33913f5ecbc7e93d01bad6a2fe4b610cda63f9c12a42efa79d9a18 module=libcontainerd namespace=moby topic=/tasks/delete type="*events.TaskDelete"
...(many identical info messages for different containers)...
```

After this error came up I restarted the test suite and it worked just fine -- the lower level failure seems transient.

Does it make sense to add error detection and/or a dumb retry policy at this level to the underlying client? I'm not sure if there's a better way to handle this, and unfortunately I didn't increase the log level on `docker` so it wasn't more specific on *why* it failed ([like it has been for others](https://github.com/moby/moby/issues/22221#issuecomment-967383169)).

### LINKED ISSUE #559: How to get container host ip?
Testcontainers Java has this `String ipAddress = container.getHost();` method to get container's host IP address. How to do the same with `testcontainers-rs`?

### LINKED ISSUE #561: Mapped ports are not updated after container restart

### LINKED ISSUE #564: Hyperlinks in the documentation are broken.
Hyperlinks in [the official documentation](https://docs.rs/testcontainers/latest/testcontainers/) are broken.

Here's *an* example (the top page): 

↓

DDtKey:
Quite large refactoring as part of project revamp #563, and also the long-awaited refactoring #386

Now it's really simple API, e.g:
```rs
let container = GenericImage::new("redis", "latest")
        .with_exposed_port(6379)
        .with_wait_for(WaitFor::message_on_stdout("Ready to accept connections"))
        .start();
```

I find this new API much easier to use, solves a lot of problems, and seems flexible enough to be extended.
This also works regardless of the tokio runtime flavor (multi-thread vs current-thread). And the sync API is still available, right under `blocking` feature (just like `reqwest` does).

From a maintainer's perspective this also simplifies the code, we don't have to worry about two different clients and their differences.

### Docker host resolution
The host is resolved in the following order:

1. Docker host from the `tc.host` property in the `~/.testcontainers.properties` file.
2. `DOCKER_HOST` environment variable.
3. Docker host from the "docker.host" property in the `~/.testcontainers.properties` file.
4. Else, the default Docker socket will be returned.

### Notes
- MSRV was bumped to `1.70` in order to use `std::sync::OnceLock`. This should NOT be a problem, tests usually executed on more recent versions (also see [this ref](https://github.com/testcontainers/testcontainers-rs/pull/503/files#r1242651354)).
- `Cli` client is removed, instead we provide `sync` (under `blocking` feature) and `async` impls based on HTTP client (bollard)
- tested with [modules](https://github.com/testcontainers/testcontainers-rs-modules-community)

## Migration guide 

- Sync API migration (`Cli` client)
  - Add `blocking` feature
  - Drop all usages of `clients::Cli`
  - Add `use testcontainers::runners::SyncRunner;`
  - Replace `client.run(image)` with `image.start()`
- Async API migration (`Http` client)
  - Remove `experimental` feature
  - Drop all usages of `clients::Http`
  - Add `use testcontainers::runners::AsyncRunner;`
  - Replace `client.run(image)` with `image.start()`

## References

Closes #386
Closes #326
Closes #475
Closes #508
Closes #392
Closes #561
Closes #559
Closes #564
Closes #538
Closes #507
Closes #89
Closes #198


README.md
```diff
@@ -15,7 +15,38 @@ The crate provides an API for working with containers in a test environment.
...
+#### Example:
+
+- Blocking API (under `blocking` feature)
+
+```rust
+use testcontainers::{core::WaitFor, runners::SyncRunner, GenericImage};
+
+#[test]
+fn test_redis() {
+    let container = GenericImage::new("redis", "latest")
```
mervyn-mccreight, line 29: I would not use `latest` here, since I personally would not recommend using `latest` in any test either due to (at least best effort) reproducibility.

testcontainers/src/core.rs
```diff
@@ -1,20 +1,15 @@
...
 
-mod container;
-#[cfg(feature = "experimental")]
-mod container_async;
-pub mod env;
 mod image;
 
+pub(crate) mod client;
+pub(crate) mod containers;
+pub(crate) mod env;
```
mervyn-mccreight, line 11: I personally like to avoid using abbreviations for more clarity in code, but that is a matter of taste.

testcontainers/src/core/utils.rs
```diff
@@ -0,0 +1,28 @@
+/// It allows using both runtime flavors without indefinite hanging.
```
mervyn-mccreight, line 1: I'd like to avoid naming something `utils` and in favor put things in files and/or modules that describe the concern of this particular function or macro.

`utils` often leads to being a trash dump.

testcontainers/src/runners/sync_runner.rs
```diff
@@ -0,0 +1,234 @@
...
+    fn start(self) -> Container<I>;
+}
+
+impl<T, I> SyncRunner<I> for T
+where
+    T: Into<RunnableImage<I>> + Send,
+    I: Image,
+{
+    fn start(self) -> Container<I> {
+        let rt = tokio::runtime::Builder::new_current_thread()
```
mervyn-mccreight, line 15: I personally like to avoid using abbreviations (except for widely known ones like e.g. `HTTP`) for more clarity inside the code.

testcontainers/src/watchdog.rs
```diff
@@ -8,22 +8,29 @@ use std::{collections::BTreeSet, sync::Mutex, thread};
 
 static WATCHDOG: Lazy<Mutex<Watchdog>> = Lazy::new(|| {
     thread::spawn(move || {
-        let signal_docker = Cli::default();
-        let mut signals =
-            Signals::new([SIGTERM, SIGINT, SIGQUIT]).expect("failed to register signal handler");
+        let rt = tokio::runtime::Builder::new_current_thread()
```
mervyn-mccreight, line 11: I personally like to avoid using abbreviations (except for widely known ones like e.g. `HTTP`) for more clarity inside the code.

testcontainers/src/core/utils.rs
DDtKey: Totally agree

testcontainers/src/runners/sync_runner.rs
DDtKey: I don't usually mind commonly used acronyms, but let me change this for clarity

kaspar-p:
The CHANGELOG that references this PR:
```
[❗] Container-centric API with refactored underlying layer (https://github.com/testcontainers/testcontainers-rs/pull/575)
```
does not contain enough information. The main way users were starting containers (Cli) was entirely removed, and the upgrade path had to be discovered in the comment of this PR.

DDtKey:
Hi @kaspar-p 👋 

All the changes has been planned and performed more than year ago. 
The description of the PR has been duplicated in the release notes. But not in changelog - that's true.

I apologize if it's not clear enough, but we had, at least:
- migration guides in release descriptions 
  - https://github.com/testcontainers/testcontainers-rs/releases/tag/0.16.0
  - https://github.com/testcontainers/testcontainers-rs/releases/tag/0.17.0
  - https://github.com/testcontainers/testcontainers-rs/releases/tag/0.18.0
- all examples and tests has been updated
- all pre-cooked images has been updated in https://github.com/testcontainers/testcontainers-rs-modules-community

kaspar-p:
That's fine, just giving an experience of how I was driven to these changes. I did `cargo upgrade` and `-i` on a few dependencies, which bumped me all the way from 0.14. The build broke, so looked into the changelog to see what changed, and didn't find much there.

I think for folks new to testcontainers, your existing paths are good coverage, just giving an anecdote on how already-using users might see.

Maybe the migration guide could be ported to the changelog? That would have been perfect for me. But if you feel most users know that that's where to look, I'm OK being the odd one out :).

DDtKey:
This is all fair, ofc.
At least it's not weird that the migration guides are "hidden" after a year, and the release notes are no longer visible on the front page.

I think we can add links to the migration guides in the changelog. That way it's clear that there is a separate guide
Migration guides can be added to the [docs site](https://rust.testcontainers.org/) as a whole.

However, I have limited bandwidth and can't prioritize this ASAP, so any PRs are welcome 🙏

Issue #262, opened 2021-02-08
Named volumes do not get removed


bonomat:
The PR: https://github.com/testcontainers/testcontainers-rs/pull/135 added volume mapping support to testcontainer-rs. 

With our current approach of removing containers (see [here](https://github.com/testcontainers/testcontainers-rs/blob/fe51e06fb2c44b1d3e3fd627cedea3fd582487ef/src/clients/cli.rs#L271-L280)) these _named_ volumes do not get removed, only _anonymous_ volumes [will be removed](https://docs.docker.com/engine/reference/commandline/rm/#options).

This means, your system will accumulated more and more containers and need to be removed manually. 

### Manually remove volumes 

We can manually remove the named volumes after we've removed the container (e.g. [when the container gets out of scope](https://github.com/testcontainers/testcontainers-rs/blob/fe51e06fb2c44b1d3e3fd627cedea3fd582487ef/src/core/container.rs#L148-L158)). 
For this, we would need to implement `volume rm [volume]`. See [docs here](https://docs.docker.com/engine/reference/commandline/volume_rm/).

The issue here is that if the volume is still shared with another container, this command will fail. 
A possible solution here would be to only print a warning if this volume is still in use but remove the container. After the last container using this volume was removed, the volume can be removed safely. 

### Use anonymous volumes

As mentioned above, we can use anonymous volumes instead of named volumes. This has the advantage that the user does not have to provide a name for the volume. 
[Anonymous volumes get automatically removed](https://docs.docker.com/engine/reference/commandline/rm/#remove-a-container-and-selectively-remove-volumes) if the linked container gets removed PROVIDED the volume is not shared with another container.

In [our public chat](https://matrix.to/#/#comit-dev:matrix.org?via=matrix.org) @thomaseizinger pointed out that variables get out of scope in the order they were created. 

## Proposed solution

Using anonymous volumes seems to be in favor for most usecases as sharing volumes between containers is an edge case. 
For the case where we want to share volumes between container, using anonymous volumes, would mean we need to [inspect the container](https://docs.docker.com/engine/reference/commandline/inspect/) and find the correct volume id.

Hence, I propose the following: 
* keep named volumes, i.e. use [`-v {named}:{internal}`](https://github.com/testcontainers/testcontainers-rs/blob/fe51e06fb2c44b1d3e3fd627cedea3fd582487ef/src/clients/cli.rs#L128-L130)
* add a function `inspect_volumes` to retrieve a list of attached volumes

[...]

PR #581, opened 2024-04-23
Set the binds option instead of volumes for mounts

### LINKED ISSUE #580: Problem mounting volumes after revamp
Hi, I was testing the new version after #575 and mounting is not working with custom images.

The problem seems to be because the runner sets the [`volumes`](https://github.com/testcontainers/testcontainers-rs/blob/d0799346adb6dc63d50d78f4327391c6ca4edf89/testcontainers/src/runners/async_runner.rs#L97-L102) config option:
```rust
// volumes
let vols: HashMap<String, HashMap<(), ()>> = runnable_image
    .volumes()
    .map(|(orig, dest)| (format!("{orig}:{dest}"), HashMap::new()))
    .collect();
config.volumes = Some(vols);
```

However that option is meant for anonymous volume mounts, where the keys are the path inside the container and the values are [an empty map](https://github.com/fussybeaver/bollard/blob/907037dd9b0d2a23d7224b06d4b37e9fe99af448/src/container.rs#L198-L201).
The `binds` hostconfig option is the one accepting a [list of mounts](https://github.com/fussybeaver/bollard/blob/907037dd9b0d2a23d7224b06d4b37e9fe99af448/codegen/swagger/src/models.rs#L2560-L2563). Replacing the above by something like this works:
```rust
// volumes
let binds: Vec<String> = runnable_image
    .volumes()
    .map(|(orig, dest)| format!("{orig}:{dest}"))
    .collect();
if !binds.is_empty() {
    config.host_config = config.host_config.map(|mut host_config| {
        host_config.binds = Some(binds);
        host_config
    });
}
```

ns-sjorgedeaguiar:
The volumes option is used for anonymous volumes, whereas the binds is the one that accepts a list of `{local}:{remote}` paths.

Fixes #580.


DDtKey (approved):

DDtKey:
Thank you!

DDtKey:
As a next step I'm going to address https://github.com/testcontainers/testcontainers-rs/issues/262

DDtKey:
I think as a next step we need to perform some refactoring here. It should be more obvious API, e.g like in [testcontainers-go](https://golang.testcontainers.org/features/files_and_mounts/)

PR #605, opened 2024-04-28
feat: allow cgroupns-mode and userns-mode to be configured 

dghilardi:
Allow cgroupns and userns to be defined when building RunnableImage

### DIFF
--- a/testcontainers/src/core/image/runnable_image.rs
+++ b/testcontainers/src/core/image/runnable_image.rs
@@ -22,6 +22,7 @@ pub struct RunnableImage<I: Image> {
     privileged: bool,
     shm_size: Option<u64>,
     cgroupns_mode: Option<CgroupnsMode>,
+    userns_mode: Option<String>,
 }
 
 /// Represents a port mapping between a local port and the internal port of a container.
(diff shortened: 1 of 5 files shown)


DDtKey (commented):
Thanks for the contribution!

I'd like to cover new configs with tests. Could we do that?

testcontainers/src/core/image/runnable_image.rs
```diff
@@ -38,6 +41,33 @@ pub enum Host {
...
+    type Error = ();
+
+    fn try_from(value: HostConfigCgroupnsModeEnum) -> Result<Self, Self::Error> {
+        match value {
+            HostConfigCgroupnsModeEnum::EMPTY => Err(()),
+            HostConfigCgroupnsModeEnum::PRIVATE => Ok(Self::Private),
+            HostConfigCgroupnsModeEnum::HOST => Ok(Self::Host),
+        }
+    }
+}
```
DDtKey, line 60: Do we really need this conversion? As far as I can see we need to convert `CgroupnsMode` to `HostConfigCgroupnsModeEnum` and not vice-versa 

Additionally, since this conversion is needed for `async_runner` - I would suggest moving the implementation there

dghilardi: Hi @DDtKey, thanks for the feedback
You are right, I removed the unused TryFrom and added some tests as you suggested.

DDtKey:
Thank you! 🙏

PR #619, opened 2024-04-30
Add exec_with

### LINKED ISSUE #617: Feature Request: Exec must return a struct that contains exit code, stdout, stderr
Current exec has this signature `pub async fn exec(&self, cmd: ExecCommand) {`. It returns nothing. 

In my tests I want to run a command and process resulting exit code & logs (both stderr & stdout).

Is it possible to add support for this feature?

Warchant:
Closes #617 - `exec_with` returns AsyncLogStream, which can be used by users to parse output stdout, stderr, or both (at once).

### DIFF
--- a/testcontainers/src/core/client.rs
+++ b/testcontainers/src/core/client.rs
@@ -87,17 +87,13 @@ impl Client {
         &self,
         container_id: &str,
         cmd: Vec<String>,
-        desired_log: DesiredLogStream,
+        attach_stdout: Option<bool>,
+        attach_stderr: Option<bool>,
     ) -> LogStreamAsync<'_> {
-        let (stdout, stderr) = match desired_log {
-            DesiredLogStream::Stdout => (true, false),
-            DesiredLogStream::Stderr => (false, true),
-        };
-
         let config = CreateExecOptions {
             cmd: Some(cmd),
-            attach_stdout: Some(stdout),
-            attach_stderr: Some(stderr),
+            attach_stdout,
+            attach_stderr,
             ..Default::default()
         };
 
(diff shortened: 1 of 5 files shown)


testcontainers/src/core/client.rs
```diff
@@ -87,17 +87,13 @@ impl Client {
         &self,
         container_id: &str,
         cmd: Vec<String>,
-        desired_log: DesiredLogStream,
```
DDtKey, line 90: Do we still need this enum with this approach?

testcontainers/src/core/containers/async_container.rs
```diff
@@ -184,26 +185,50 @@ where
         self.docker_client.docker_hostname().await
     }
 
-    pub async fn exec(&self, cmd: ExecCommand) {
-        let ExecCommand {
+    pub async fn exec_with(&self, cmd: ExecWithCommand) -> LogStreamAsync<'_> {
```
DDtKey, line 188: `LogStreamAsync` seems to be crate-level structure, we need to expose it, don't we?

testcontainers/src/core/containers/async_container.rs
DDtKey, line 1: That will definitely lead to breaking changes in near future.
- there is no way to distinguish `stdout` and `stderr`
- only UTF-8 logs will work (because of LogStreamAsync iml)

Also, we don't return an exit code. I think that also can be useful (we can get it with `inspect_exec(..)` )

DDtKey:


Let's also cover with some basic testing. 

And another question if we going to provide this api for sync(blocking) container? It will require a blocking log stream wrapper. Can be implemented in separate PR though

DDtKey:
I suggest this interface for now: #622

WDYT?

PR #626, opened 2024-05-09
fix: collect bridge IP address correctly

dlen:
I bumped into this issue while using https://github.com/testcontainers/testcontainers-rs-modules-community/tree/main/src/mysql 
and the latest version of testcontainers-rs.

`get_bridge_ip_address` was not returning the correct IP address value. When running a container on GNU/Linux systems
`NetworkSettings.bridge` might be an empty string therefore it is not useful as a default value.

This PR tries to provide a default value for such scenarios where:

- No `--bridge` flag has been explicitly provided when starting the container
- `NetworkSettings.bridge` is an empty string
- Container has not been launched with a specific network and therefore defaults to whatever is configured on the docker daemon

Thanks for all the efforts in this project!

### DIFF
--- a/testcontainers/src/core/containers/async_container.rs
+++ b/testcontainers/src/core/containers/async_container.rs
@@ -150,39 +150,31 @@ where
             .network_mode
             .unwrap_or_else(|| panic!("container {} has no network mode", self.id));
 
-        if network_mode != "bridge" {
-            panic!("container {} is not in bridge mode", self.id);
-        }
+        let network_settings = self
+            .docker_client
+            .inspect_network(&network_mode)
+            .await
+            .unwrap_or_else(|_| panic!("container {} network mode does not exist", self.id));
 
-        let network_settings = container_settings
+        network_settings
+            .driver
+            .unwrap_or_else(|| panic!("network {} is not in bridge mode", network_mode));
+
+        let container_network_settings = container_settings
             .network_settings
             .unwrap_or_else(|| panic!("container {} has no network settings", self.id));
 
-        let mut networks = network_settings
+        let mut networks = container_network_settings
             .networks
-            .unwrap_or_else(|| panic!("container {} has no any networks", self.id));
-
-        let bridge_name = self
-            .image
-            .network()
-            .clone()
-            .or_else(|| network_settings.bridge.filter(|b| !b.is_empty()))
-            .or_else(|| Some("bridge".to_owned()))
-            .unwrap_or_else(|| panic!("container {} has missing bridge name", self.id));
+            .unwrap_or_else(|| panic!("container {} has no networks", self.id));
 
         let ip = networks
-            .remove(&bridge_name)
+            .remove(&network_mode)
             .and_then(|network| network.ip_address)
-            .unwrap_or_else(|| {
-                panic!(
-                    "container {} has missing bridge IP {:?}",
-                    self.id,
-                    self.image.network()
-                )
-            });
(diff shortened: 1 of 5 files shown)


testcontainers/src/core/containers/async_container.rs
```diff
@@ -139,30 +140,53 @@ where
...
 
         let bridge_name = self
             .image
             .network()
             .clone()
-            .or(settings.bridge)
+            .or_else(|| {
+                network_settings
+                    .bridge
+                    .and_then(|b| if !b.is_empty() { Some(b) } else { None })
```
DDtKey, line 172: We can use `filter` I guess?

testcontainers/src/core/containers/async_container.rs
```diff
@@ -139,30 +140,53 @@ where
...
+            })
+            .or(Some("bridge".to_owned()))
```
DDtKey, line 174: Let's use `or_else` to avoid allocations every time?

DDtKey (commented):

The changes makes sense to me, I've left a couple of minor comments

dlen:
Wow @DDtKey thanks a lot for the prompt response!! I have pushed your suggestions

DDtKey:
Could you take a look at falling tests?

dlen:
Yes, np

dlen:
Found some issues with the old approach so I had to switch to determine the network attached to the container by using `HostConfig.NetworkMode` then inspect that network and find its driver to finally conclude if it is a bridged network in order to obtain the bridge ip. I think the test should pass on the CI now.

testcontainers/src/runners/async_runner.rs
```diff
@@ -384,6 +384,40 @@ mod tests {
...
+            .with_network("bridge")
+            .start()
+            .await;
+
+        assert!(!container
+            .get_bridge_ip_address()
+            .await
+            .to_string()
+            .is_empty())
+    }
```
DDtKey, line 404: Let's probably add a check with custom network (bridge mode)?

DDtKey:
I think we should consider this change as breaking one. I mean it's definitely fixes the described issue, but the behavior has changed.

Thus, I'm going to include this onto `0.17.0`.  
JFYI: I want to include couple of changes in the next major release, so it will probably will take some time before releasing (not too long)

dlen:
No problem. Thanks for your time man!

DDtKey (approved):

testcontainers/src/runners/async_runner.rs
```diff
@@ -384,6 +384,40 @@ mod tests {
         assert!(!client.network_exists("awesome-net-2").await)
     }
 
+    #[tokio::test]
+    async fn async_should_rely_on_network_mode_when_network_is_provided_and_settings_bridge_empty()
```
DDtKey, line 388: Hi @dlen!
I was about to merge the PR, but wanted to double-check the test coverage and current behavior in `main`. 
And understood that these tests are passing against it.

I'm just wondering if this could be covered better, because if we change this back - we don't see the crash, however we are aware of some issues.
Sorry for the late notice

dlen:
Yeah I could add more tests, no problem. However I won't be able to look
into it till later next week.

If you have something specific in mind let me know.

DDtKey:
Nothing specific at the moment, but I'll think of it a bit later.

The main goal is to make a test which fails against main, but passes with new changes. That would be ideal option, as we consider this a fix

But in fact, I don't rush you. We even can proceed with this and add tests separately

PR #652, opened 2024-06-11
refactor!: add `ImageExt` trait to avoid explicit conversion to `RunnableImage`

DDtKey:
This PR allows to override some image values without explicit conversion to `RunnableImage`.

For example:

```rs
// Instead of this:
use testcontainers::RunnableImage;
RunnableImage::from(image).with_tag("x").with_network("a").start();

// It can be done directly on image:
use testcontainers::ImageExt;
image.with_tag("x").with_network("a").start();
```

Also, it allows to simplify `GenericImage`. Because it was kinda duplicative code.


mervyn-mccreight (approved):
Nice work! 🚀

PR #695, opened 2024-07-07
feat!: support waiting from both standard streams

DDtKey:
Might be useful to avoid duplication in some cases

### DIFF
--- a/testcontainers/src/core/logs/stream.rs
+++ b/testcontainers/src/core/logs/stream.rs
@@ -51,6 +51,19 @@ impl LogStream {
             .boxed()
     }
 
+    /// Converts the log stream into a stream of raw bytes, regardless of the source.
+    pub(crate) fn into_bytes(self) -> RawLogStream {
+        self.inner
+            .filter_map(|record| async move {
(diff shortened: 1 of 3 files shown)


ns-sjorgedeaguiar:
@DDtKey on a related note, am I right that it's not possible to wait for the command to complete? I see that we can wait for messages to be present in stderr/stdout, or for a specific exit code, but there doesn't seem to be a away to wait until the command finishes regardless of the outcome. There's the timeout option but it sleeps for the whole duration, not circuit breaking when the command is finished.

In my case I wrote a wrapper to execute a SQL statement, and wanted to print whatever error occurred when there was some. The exit code or expected message depended on what the caller was doing (e.g. inserting, deleting, etc) and the exit code could be anything. My workaround was to keep polling until an exit code was available.

DDtKey:
Hi @ns-sjorgedeaguiar 
Yes, that's correct. however I think more appropriate place for the discussion is https://github.com/testcontainers/testcontainers-rs/issues/702

Where you also can find a workaround. But it's something we can support for sure

Issue #701, opened 2024-07-20
Support customize CMD for GenericImage


tisonkun:
I can see that we support customize entrypoint. Is there any blocker from supporting CMD also?

DDtKey:
Hi there 👋 

I think there is no blockers for this, and it's currently supported with [ImageExt::with_cmd](https://docs.rs/testcontainers/latest/testcontainers/core/trait.ImageExt.html#tymethod.with_cmd)
In general, `ImageExt` was introduced to unify the interface and make overrides simpler for any `Image`.

But we can extend `GenericImage` too.

PR #703, opened 2024-07-21
feat: support with_cmd for GenericImage

tisonkun:
This closes https://github.com/testcontainers/testcontainers-rs/issues/701.

@DDtKey I'm new to here so perhaps you can tell me some conveniences and best practices if I miss them :D

### DIFF
--- a/testcontainers/src/images/generic.rs
+++ b/testcontainers/src/images/generic.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 use crate::{
     core::{ports::ContainerPort, WaitFor},
     Image,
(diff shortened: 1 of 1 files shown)


tisonkun:
Seems some logic conflicts. Let me try to debug locally.

DDtKey (changes requested):
I remembered why the `cmd` isn't exposed as part of the `GenericContainer`.

It produces fully duplicative methods with `ImageExt` for no reason actually. Currently only essential options are provided as part of `GenericImage` api (which are not available within `ImageExt`)

Also, it would lead to unnecessary name collisions (not critical, but also - for no reason).

Let's say if we add `cmd`, why other options shouldn't be available? There is no benefit, but unnecessary confusion, actually

Moreover, the common way to go is usually implementation of custom `Image` (or use [provided](https://crates.io/crates/testcontainers-modules)) rather than using generic one.

I apologize for this, but I have to reject the PR unless there is a strong reason for this.

Thank you 🙏

tisonkun:
Sounds reasonable. Never mind :D

Issue #708, opened 2024-07-28
Unconditional `Url` parsing breaks `NamedPipeConnector` on Windows


blaenk:
On windows, the `DEFAULT_DOCKER_HOST` is correctly a named pipe path:

https://github.com/testcontainers/testcontainers-rs/blob/e354b7532b5d448ab37c9f95dab6c68c592cb62b/testcontainers/src/core/env/config.rs#L32-L34

However, as part of computing the `docker_host()` based on the configuration hierarchy, it is unconditionally parsed into a `Url`:

https://github.com/testcontainers/testcontainers-rs/blob/e354b7532b5d448ab37c9f95dab6c68c592cb62b/testcontainers/src/core/env/config.rs#L141

This has the effect of normalizing the Url into the following value (note the removed period from the beginning of `.pipe`):

```
npipe:////pipe/docker_engine
```

When bollard attempts to connect through that named pipe, it removes the `npipe://` prefix, expecting to be left with just `//.pipe/docker_engine`, but due to this artifact, it is instead left with `//pipe/docker_engine`.

https://github.com/fussybeaver/bollard/blob/865805f87e4066dbf1a283139e1c6148b62ccd80/src/docker.rs#L838-L843

This results in:

```
called `Result::unwrap()` on an `Err` value: Client(CreateContainer(HyperLegacyError { err: hyper_util::client::legacy::Error(Connect, Os { code: 53, kind: NotFound, message: "The network path was not found." }) }))
```

I confirmed this is a problem by forking this repo and hardcoding the value passed into the bollard client with the named path `"npipe:////./pipe/docker_engine"`

https://github.com/testcontainers/testcontainers-rs/blob/e354b7532b5d448ab37c9f95dab6c68c592cb62b/testcontainers/src/core/client/bollard_client.rs#L31-L36

I think we can still parse into a Url to check the scheme for example, but we probably shouldn't use the parsed Url itself to pass on to clients. Consider that the bollard and hyper clients just take an `&str` anyway. If there's an error with the url formatting, it'll come out eventually.

Separately, It might be worthwhile adding a matrix to also run tests against Windows.

DDtKey:
I think the main issue is normalization, and there is no way (currently?) to disable this behavior for `Url`.

I would like to have stricter and more explicit type used, but it seems related to https://github.com/servo/rust-url/issues/602

So I’m good with proposed solution

PR #709, opened 2024-07-28
fix: don't assume docker host is a url

blaenk:
This is one way to fix https://github.com/testcontainers/testcontainers-rs/issues/708

Although we still parse into a `Url`, we don't pass that value on to clients, since `Url` may mutate the input, which in the case of #708, breaks Windows `NamedPipeConnector`, which is the default way of connecting to the Docker daemon.

I also disabled the SCTP tests on Windows since it doesn't seem there's support for that on Windows.

### DIFF
--- a/testcontainers/src/core/client.rs
+++ b/testcontainers/src/core/client.rs
@@ -1,13 +1,5 @@
-use crate::core::{
-    client::exec::ExecResult,
-    env,
-    env::ConfigurationError,
-    logs::{
-        stream::{LogStream, RawLogStream},
-        LogFrame, LogSource, WaitingStreamWrapper,
-    },
-    ports::{PortMappingError, Ports},
-};
+use std::{io, str::FromStr};
+
 use bollard::{
     auth::DockerCredentials,
     container::{Config, CreateContainerOptions, LogOutput, LogsOptions, RemoveContainerOptions},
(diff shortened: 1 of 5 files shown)


testcontainers/src/core/client.rs
```diff
@@ -312,15 +312,15 @@ impl Client {
...
-        match docker_host.scheme() {
-            "tcp" | "http" | "https" => {
-                docker_host
-                    .host()
-                    .map(|host| host.to_owned())
-                    .ok_or_else(|| {
-                        ConfigurationError::InvalidDockerHost(docker_host.to_string()).into()
-                    })
-            }
+        let docker_host_url = Url::from_str(docker_host).expect("expected a valid Docker Host URL");
```
blaenk, line 315: This expectation was already occurring within `config.docker_host()`, FWIW.

This will not be a problem for `NamedPipeConnector`. The path always did parse without `Error`/panic, it's just that it would normalize to an erroneous Url. Parsing the Url remains fine so long as we don't pass it on to the clients directly. We still parse it here so we can differentiate the scheme, for example.

DDtKey:
Thank you for the contribution and well-described issue! 🚀 

Could you add a test for windows target, please? Otherwise it might be too easy to break again accidentally

testcontainers/src/core/client.rs
DDtKey: That makes sense, but could you replace `expect` with an error?
We prefer not to panic on library level and user is responsible for handling.

It used to be 

```rs
|url::ParseError| ConfigurationError::InvalidDockerHost(e.to_string()))?;
```

blaenk: Maybe I'm missing something? It seems that `config.docker_host()` always panicked:

https://github.com/testcontainers/testcontainers-rs/blob/e354b7532b5d448ab37c9f95dab6c68c592cb62b/testcontainers/src/core/env/config.rs#L141

I agree with you though and I've gone ahead and made the change in e1d4b14

DDtKey: It's for `DEFAULT_DOCKER_HOST` - it can't panic, because we know the value in advance.

But in the highlighted code the host might be a user's input

DDtKey:
Could you apply `rustfmt` and check falling tests?  

Overall, looks like LGTM 🙏

testcontainers/src/core/client.rs
blaenk: You're right! Thanks for pointing it out.

blaenk:
Done! Thanks for reviewing.

DDtKey (approved):

Issue #711, opened 2024-07-29
Cutting out a new release


stormshield-gt:
I will be very interested if you could please cut out a new ~~minor release~~ patch release.
I'm looking into the upgrade of `rustls`  to 0.23 which we just get through https://github.com/testcontainers/testcontainers-rs/pull/710 that update `bollard` to 0.17

DDtKey:
Hi @stormshield-gt 👋 

Not sure I got you.
Do you want us to rollout a new minor release with #710 included?

stormshield-gt:
> I wanted to prepare a major release soon, there are some breaking changes in main.

Did you mean minor release ? Or you are going 1.0.0 ? It any case it will definitely work for me.

> Otherwise we might consider a patch release with only this change, however it's not a fix

Because this crate has not reached 1.0.0 yet, I think it's OK to use patch release for all non-breaking changes. If we use  minor release even for non-breaking changes, it adds a burden to downstream users. 
For instance, 0.20 is not [semVer compatible](https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility) with 0.21 so it won't be updated automatically by `cargo update`, and users will have to manually update their *Cargo.toml* files.
`rustls` for instance use patch release for non-breaking changes.

DDtKey:
By major I meant minor part of semver. 
Just wanted to highlight that there are breaking changes and we use `minor` part for these.

It will be `0.21.0`

DDtKey:
Actually, we can't backport deps update. 
In general, upgrade of `rustls` to `0.23` is breaking change and may affect users - there are some [platform-related limitations in this version](https://aws.github.io/aws-lc-rs/index.html)

stormshield-gt:
Actually, it seems that [bollard wants to force the ring provider ](https://github.com/fussybeaver/bollard/blob/01413b7b0263ab74bd0026582b1a5825b8055504/Cargo.toml#L60) that doesn't have the NASM requirement on windows. 

But I think they didn't disable default features so maybe they are bringing both `ring` and `aws-lc-rs` provider, I will check that and open an issue there. I think they should re-export the provider as features, as did the rest of the ecosystem like [tokio-rustls](https://docs.rs/crate/tokio-rustls/latest/features) or [hyper-rustls](https://docs.rs/crate/hyper-rustls/latest/features).

[...]

PR #714, opened 2024-08-02
feat: allow configuring resource ulimits

blaenk:
Thanks again for working on this project!

This exposes a way to configure resource ulimits (the [`--ulimit`](https://docs.docker.com/reference/cli/docker/container/run/#ulimit) flag for `docker run`).

I'm relying on CI tests because I'm currently on Windows and didn't configure my environment to reflect https://github.com/testcontainers/testcontainers-rs/issues/711, since it seems I'd need to install nasm now, as I received some errors regarding it.

### DIFF
--- a/testcontainers/src/core/containers/request.rs
+++ b/testcontainers/src/core/containers/request.rs
@@ -6,6 +6,8 @@ use std::{
     time::Duration,
 };
 
+use bollard_stubs::models::ResourcesUlimits;
+
 use crate::{
     core::{
         logs::consumer::LogConsumer, mounts::Mount, ports::ContainerPort, ContainerState,
@@ -41,14 +43,6 @@ pub struct ContainerRequest<I: Image> {
 pub struct PortMapping {
     pub(crate) host_port: u16,
     pub(crate) container_port: ContainerPort,
-}
-
-/// Represents a resource ulimit.
-#[derive(Clone, Debug, Eq, PartialEq)]
-pub struct ResourcesUlimits {
-    pub(crate) name: Option<String>,
-    pub(crate) soft: Option<i64>,
-    pub(crate) hard: Option<i64>,
 }
 
 #[derive(parse_display::Display, Debug, Clone)]
@@ -102,10 +96,6 @@ impl<I: Image> ContainerRequest<I> {
 
     pub fn ports(&self) -> Option<&Vec<PortMapping>> {
         self.ports.as_ref()
-    }
-
-    pub fn ulimits(&self) -> Option<&Vec<ResourcesUlimits>> {
-        self.ulimits.as_ref()
     }
 
     pub fn privileged(&self) -> bool {
--- a/testcontainers/src/core/containers/request.rs
+++ b/testcontainers/src/core/containers/request.rs
@@ -6,8 +6,6 @@ use std::{
     time::Duration,
 };
 
-use bollard_stubs::models::ResourcesUlimits;
-
 use crate::{
     core::{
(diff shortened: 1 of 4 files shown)


DDtKey:
> I'm relying on CI tests because I'm currently on Windows and didn't configure my environment to reflect #711, since it seems I'd need to install nasm now, as I received some errors regarding it.

Yeah, unfortunately this requirement comes from `rustls` with `aws-lc-rs` driver. 

But `bollard` merged the fix not to require the driver except `ring`:
https://github.com/fussybeaver/bollard/pull/441

So, as workaround I may suggest using patch of `bollard` in your `Cargo.toml` to git one.
However I can't guarantee there is no other deps in your lock file which keeps `rustls` default features enabled.

testcontainers/src/core/containers/request.rs
```diff
@@ -95,6 +98,10 @@ impl<I: Image> ContainerRequest<I> {
         self.ports.as_ref()
     }
 
+    pub fn ulimits(&self) -> Option<&Vec<ResourcesUlimits>> {
```
DDtKey, line 101: We don't re-expose bollard's types right now (probably will in the future, with the access to client)

For now I'd prefer not to expose it with public methods. Any change of bollard would mean breaking change for us, when we could re-map changes into our structures in compatible way (depending on the changes)

Also, we don't re-export the type or crate, it may be not that convenient for users.

So for more we have two options:
- use our own type
- don't expose publicly, only internal use

DDtKey (commented):
I've left only one comment, please take a look. Overall looks like LGTM

testcontainers/src/core/containers/request.rs
blaenk: Makes sense! I created our own type. Let me know what you think.

The method now takes an `Option<i64>` for the `hard` limit since apparently that is optional, it'll use `soft` when it's missing. See https://docs.docker.com/reference/cli/docker/container/run/#ulimit

testcontainers/src/core/containers/request.rs
```diff
@@ -42,6 +43,14 @@ pub struct PortMapping {
     pub(crate) container_port: ContainerPort,
 }
 
+/// Represents a resource ulimit.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ResourcesUlimits {
+    pub(crate) name: Option<String>,
+    pub(crate) soft: Option<i64>,
+    pub(crate) hard: Option<i64>,
+}
```
DDtKey, line 52: Sorry, one more comment:

If we want to expose this one, we need to provide public getters. And also re-export it here: https://github.com/testcontainers/testcontainers-rs/blob/f8d4a3d3a63c338cc7baf264a6fb6cce88bbae25/testcontainers/src/core/containers/mod.rs#L7 (due to some historical organization of modules, might be changed in the future)

Alternative would be not to expose it (remove `ulimits` method)

blaenk: I'll just remove it altogether. Do you still want it to show up in the `Debug` implementation?

DDtKey (approved):

testcontainers/src/core/containers/request.rs
DDtKey: Yep, debug might be helpful. Thank you!

blaenk:
Thank you for the review!

No rush at all, but I'm curious when you can imagine a release including it.

DDtKey:
I think within couple of days since it’s comparable change.
I just need to check if there is anything else I wanted to finish before proceeding with the new release

DDtKey:
It's been released!

blaenk:
I appreciate the heads up thank you!

PR #721, opened 2024-08-16
fix: docker socket detection on unix

mominul:
Docker Desktop on macOS creates the socket in the user's home directory from version 4.13.0.[[1](https://github.com/docker/for-mac/issues/6529)][[2](https://github.com/testcontainers/testcontainers-java/issues/6165)]

And the Docker team [has the intention to move away](https://github.com/docker/for-mac/issues/6529#issuecomment-1710091807) from the root-owned `/var/run/docker.sock`

There is a option named `Allow the default Docket Socket to be used (requires password)` in the Docker Desktop, but the created symbolic link `/var/run/docker.sock` [doesn't persist](https://github.com/lando/lando/issues/3533#issuecomment-1464252377) between OS restart or OS upgrade.

Without this patch or creating a symbolic link, macOS devs have to face the ```called `Result::unwrap()` on an `Err` value: Client(Init(SocketNotFoundError("/var/run/docker.sock")))``` error.

I have used `Cow` to avoid the need of cloning the fields of `Config` struct. I am open to suggestion or other approaches.

Thanks!

### DIFF
--- a/testcontainers/src/core/env/config.rs
+++ b/testcontainers/src/core/env/config.rs
@@ -1,9 +1,10 @@
-use dirs::{home_dir, runtime_dir};
 use std::{
     borrow::Cow,
     path::{Path, PathBuf},
     str::FromStr,
 };
+
+use dirs::{home_dir, runtime_dir};
 
 use crate::core::env::GetEnvValue;
 
--- a/testcontainers/src/core/env/config.rs
+++ b/testcontainers/src/core/env/config.rs
@@ -1,9 +1,8 @@
+use dirs::{home_dir, runtime_dir};
 use std::{
     borrow::Cow,
-    env::var,
     path::{Path, PathBuf},
-    process::Command as StdCommand,
-    str::{from_utf8, FromStr},
+    str::FromStr,
 };
 
 use crate::core::env::GetEnvValue;
@@ -133,7 +132,6 @@ impl Config {
     ///     1. `${XDG_RUNTIME_DIR}/.docker/run/docker.sock`.
     ///     2. `${HOME}/.docker/run/docker.sock`.
     ///     3. `${HOME}/.docker/desktop/docker.sock`.
-    ///     4. `/run/user/${UID}/docker.sock`, where `${UID}` is the user ID of the current user.
     ///  6. The default Docker socket including schema will be returned if none of the above are set.
     pub(crate) fn docker_host(&self) -> Cow<'_, str> {
         self.tc_host
@@ -144,34 +142,28 @@ impl Config {
                 if cfg!(unix) {
                     check_path_exists("/var/run/docker.sock".into())
                         .or_else(|| {
-                            var("XDG_RUNTIME_DIR").ok().and_then(|dir| {
-                                check_path_exists(format!("{dir}/.docker/run/docker.sock",))
+                            runtime_dir().and_then(|dir| {
+                                check_path_exists(format!(
+                                    "{}/.docker/run/docker.sock",
+                                    dir.display()
+                                ))
                             })
                         })
                         .or_else(|| {
-                            check_path_exists(format!(
-                                "{}/.docker/run/docker.sock",
-                                var("HOME").unwrap()
-                            ))
+                            home_dir().and_then(|dir| {
+                                check_path_exists(format!(
+                                    "{}/.docker/run/docker.sock",
+                                    dir.display()
+                                ))
+                            })
                         })
                         .or_else(|| {
-                            check_path_exists(format!(
-                                "{}/.docker/desktop/docker.sock",
-                                var("HOME").unwrap()
-                            ))
-                        })
-                        .or_else(|| {
-                            // Get the user ID of the current user by running `id -u`
-                            StdCommand::new("id")
-                                .arg("-u")
-                                .output()
-                                .ok()
-                                .and_then(|output| {
-                                    from_utf8(&output.stdout).map(|s| s.trim().to_string()).ok()
-                                })
-                                .and_then(|uid| {
-                                    check_path_exists(format!("/run/user/${uid}/docker.sock"))
-                                })
+                            home_dir().and_then(|dir| {
+                                check_path_exists(format!(
+                                    "{}/.docker/desktop/docker.sock",
+                                    dir.display()
+                                ))
+                            })
                         })
                         .map(|p| format!("unix://{p}"))
                         .map(Cow::Owned)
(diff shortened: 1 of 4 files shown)


DDtKey:
Thank you @mominul for the contribution and detailed description! 🚀

testcontainers/src/core/env/config.rs
```diff
@@ -126,11 +128,21 @@ impl Config {
...
+            .unwrap_or_else(|| {
+                // Docker Desktop on macOS creates the socket in the user's home directory
+                // from version 4.13.0.
+                // https://github.com/docker/for-mac/issues/6529
+                if cfg!(target_os = "macos") {
+                    format!("unix://{}/.docker/run/docker.sock", var("HOME").unwrap()).into()
+                } else {
+                    DEFAULT_DOCKER_HOST.into()
+                }
+            })
```
DDtKey, line 145: I think to make it more universal and reliable, we could check the socket existence, something similar to what `Testcontainers for Go` do: https://github.com/testcontainers/testcontainers-go/blob/c44b1b2ed935455fa222f122a7474efe309f641c/docs/features/configuration.md?plain=1#L78-L86

It might be the case not only for macos

testcontainers/src/core/env/config.rs
```diff
@@ -126,11 +128,21 @@ impl Config {
     ///  2. `DOCKER_HOST` environment variable.
     ///  3. Docker host from the `docker.host` property in the `~/.testcontainers.properties` file.
     ///  4. Else, the default Docker socket will be returned.
-    pub(crate) fn docker_host(&self) -> &str {
+    pub(crate) fn docker_host(&self) -> Cow<'_, str> {
```
DDtKey, line 131: we will also need to update documentation of the function and in `lib.rs`

DDtKey:
Hi @mominul 👋 
I'm just wondering if you're going to take a look at comments and probably implement the changes? (No pressure at all, trying understand the state). Perhaps we will extract suggested changes into separate PR if you don't have time

mominul:
Hello @DDtKey!

Sorry for the delay in implementing the changes; I'll try to implement them by next week. 

Thanks! 😄

testcontainers/src/core/env/config.rs
mominul: Done!

testcontainers/src/core/env/config.rs
mominul: Done!

mominul:
@DDtKey I have addressed your suggestions, can you have a look at it please?

testcontainers/src/core/env/config.rs
```diff
@@ -125,12 +128,58 @@ impl Config {
...
+                            check_path_exists(format!(
+                                "{}/.docker/run/docker.sock",
+                                var("HOME").unwrap()
+                            ))
+                        })
+                        .or_else(|| {
+                            check_path_exists(format!(
+                                "{}/.docker/desktop/docker.sock",
+                                var("HOME").unwrap()
+                            ))
```
DDtKey, line 161: Suggesting the following approach:
1. don't panic, but ignore the branch (it's unlikely to panic, but still)
2. we use `dirs::home_dir()`, just a bit less hardcoded variables and repetitions

```suggestion
                            dirs::home_dir().ok().and_then(|path| {
                                 check_path_exists(format!("{path}/.docker/desktop/docker.sock"))
                            })
```

testcontainers/src/core/env/config.rs
```diff
@@ -125,12 +128,58 @@ impl Config {
...
+                        .or_else(|| {
+                            var("XDG_RUNTIME_DIR").ok().and_then(|dir| {
+                                check_path_exists(format!("{dir}/.docker/run/docker.sock",))
+                            })
+                        })
+                        .or_else(|| {
+                            check_path_exists(format!(
+                                "{}/.docker/run/docker.sock",
+                                var("HOME").unwrap()
+                            ))
```
DDtKey, line 155: and the same here

testcontainers/src/core/env/config.rs
```diff
@@ -125,12 +128,58 @@ impl Config {
...
+                                .arg("-u")
+                                .output()
+                                .ok()
+                                .and_then(|output| {
+                                    from_utf8(&output.stdout).map(|s| s.trim().to_string()).ok()
+                                })
+                                .and_then(|uid| {
+                                    check_path_exists(format!("/run/user/${uid}/docker.sock"))
+                                })
+                        })
```
DDtKey, line 175: We could consider to use [uzers crate](https://crates.io/crates/uzers) which wraps libc call.
Or even ignore this particular option, users can override it with config options 🤔 
I mean we can support this option on demand, for now we didn't have such requests.

However, we can keep this impl too

testcontainers/src/core/env/config.rs
```diff
@@ -125,12 +128,58 @@ impl Config {
...
+                            // Get the user ID of the current user by running `id -u`
+                            StdCommand::new("id")
+                                .arg("-u")
+                                .output()
+                                .ok()
+                                .and_then(|output| {
+                                    from_utf8(&output.stdout).map(|s| s.trim().to_string()).ok()
+                                })
+                                .and_then(|uid| {
+                                    check_path_exists(format!("/run/user/${uid}/docker.sock"))
```
DDtKey, line 173:
```suggestion
                                    check_path_exists("/run/user/${uid}/docker.sock".to_string())
```

testcontainers/src/core/env/config.rs
```diff
@@ -150,6 +199,15 @@ impl Config {
     }
 }
 
+/// Checks if the path exists and returns it if it does.
+fn check_path_exists(path: String) -> Option<String> {
```
DDtKey, line 203: couple of alternative names:
- `get_existing_path`
- `filter_existing_path`
- `validate_path`


`check` usually doesn't imply "return/wrap" semantic

testcontainers/src/core/env/config.rs
mominul: Done!

testcontainers/src/core/env/config.rs
mominul: Done!

testcontainers/src/core/env/config.rs
mominul: I have removed this path fallback.

testcontainers/src/core/env/config.rs
mominul: Done!

mominul:
@DDtKey Can you have another look at this, please? I have removed the user id dependent socket path fallback until we get demand to add it.

DDtKey (approved):

PR #722, opened 2024-08-19
chore: add `#![forbid(unsafe_code)]` to the library

CommanderStorm:
The motivation is to make [`cargo-geiger`](https://github.com/geiger-rs/cargo-geiger) be sure that there is no unsafe code


DDtKey (approved):

DDtKey (approved):
Thank you! 👍

Issue #720, opened 2024-08-08
[Feature request]: Allow specifiying the working directory


nicmr:
Equivalent to the docker run `-w, --workdir` flag.
This would be extremely useful for some workloads and allow for feature parity with other testcontainer implementations.
I might try to contribute this myself if I can find the time - I will leave a comment in this thread if I start working on it.

For reference:
- Go implementation has added this feature with https://github.com/testcontainers/testcontainers-go/pull/1899 
- Java implementation added this feature with https://github.com/testcontainers/testcontainers-java/commit/e2b61d45c702af93375c32ff4132d07c84f179d0

DDtKey:
Hi @nicmr 👋

Thank you for the request and references! 🙏

This shouldn’t be a problem to extend [ContainerRequest](https://docs.rs/testcontainers/latest/testcontainers/core/struct.ContainerRequest.html) and [ImageExt](https://docs.rs/testcontainers/latest/testcontainers/core/trait.ImageExt.html)

I will be glad to accept a PR

nicmr:
Hi @DDtKey, thank you for the hints. I've taken a look at the types you linked and `bollard` and I think I have a good idea of how to implement this. I'll try to submit a PR this weekend :)

nicmr:
Closing this one as it was implemented with #724. Thank you for the guidance and the review @DDtKey. Also kudos for having such a clean codebase, it was easy to understand how things work under the hood.

PR #724, opened 2024-08-25
feat: add `working_dir` to `ContainerRequest`,`ImageExt`

nicmr:
Implements #720 by adding support for setting the working directory on `ImageExt` and `ContainerRequest`. Also adds a test for this feature. This is equivalent to the `-w, --workdir` flag of `docker run`.

The abbreviation `working_dir` (as opposed to e.g. `working_directory`) was chosen to use the same name as the `bollard` crate.

### DIFF
--- a/testcontainers/src/core/containers/request.rs
+++ b/testcontainers/src/core/containers/request.rs
@@ -224,6 +224,7 @@ impl<I: Image + Debug> Debug for ContainerRequest<I> {
             .field("cgroupns_mode", &self.cgroupns_mode)
             .field("userns_mode", &self.userns_mode)
             .field("startup_timeout", &self.startup_timeout)
+            .field("working_dir", &self.working_dir)
             .finish()
     }
 }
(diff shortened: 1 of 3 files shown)


testcontainers/src/core/containers/request.rs
```diff
@@ -35,6 +35,7 @@ pub struct ContainerRequest<I: Image> {
     pub(crate) cgroupns_mode: Option<CgroupnsMode>,
     pub(crate) userns_mode: Option<String>,
     pub(crate) startup_timeout: Option<Duration>,
+    pub(crate) working_dir: Option<String>,
```
DDtKey, line 38: Let's also add the field to `impl<I: Image + Debug> Debug for ContainerRequest<I> {`

DDtKey (approved):
Thank you very much for the contribution! 🚀

testcontainers/src/core/containers/request.rs
nicmr: Thank you, I missed that one :)

DDtKey (approved):

Issue #578, opened 2024-04-22
[Feature request]: add capabiltities support


stormshield-gt:
It would be great to have [capabilities](https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities) support.
My current use case is for the` hashicorp/vault` image that require the  `IPC_LOCK` capability

nicmr:
@DDtKey I've started implementing this, shouldn't be too different from #724 , but this time for `cap_add` and `cap_drop`

nicmr:
@DDtKey @stormshield-gt can we close this since #726 was merged?

DDtKey:
Sure, thank you for implementing this! 🙏

PR #726, opened 2024-08-30
feat: add `cap_add` and `cap_drop` support

nicmr:
Hi :wave: This PR implements #578 and adds tests for the added functionality.
Specifically, this PR adds support for adding and dropping capabilities for containers.
Have a nice weekend!

### DIFF
--- a/testcontainers/src/core/image/image_ext.rs
+++ b/testcontainers/src/core/image/image_ext.rs
@@ -79,6 +79,12 @@ pub trait ImageExt<I: Image> {
     /// Sets the container to run in privileged mode.
     fn with_privileged(self, privileged: bool) -> ContainerRequest<I>;
 
+    /// Adds the capabilities to the container
+    fn with_cap_add(self, capabilities: Vec<String>) -> ContainerRequest<I>;
+
+    /// Drops the capabilities from the container's capabilities
+    fn with_cap_drop(self, capabilities: Vec<String>) -> ContainerRequest<I>;
+
     /// cgroup namespace mode for the container. Possible values are:
     /// - [`CgroupnsMode::Private`]: the container runs in its own private cgroup namespace
     /// - [`CgroupnsMode::Host`]: use the host system's cgroup namespace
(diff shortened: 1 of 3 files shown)


DDtKey (approved):

DDtKey (approved):

testcontainers/src/core/image/image_ext.rs
```diff
@@ -206,6 +212,26 @@ impl<RI: Into<ContainerRequest<I>>, I: Image> ImageExt<I> for RI {
...
 
+    fn with_cap_add(self, capabilities: impl IntoIterator<Item = String>) -> ContainerRequest<I> {
+        let mut container_req = self.into();
+        container_req
+            .cap_add
+            .get_or_insert_with(Vec::new)
+            .extend(capabilities);
+
+        container_req
+    }
```
DDtKey, line 223: In general, I'm not sure we need the "extend" semantic. We don't use it for other methods.
I think it's acceptable to replace if method called again (and perhaps more obvious)

nicmr: I see your point. With some `with_` functions we do have the additive semantic of `extend` though, e.g. mounts and environment variables (with the difference that the newly added functions currently support adding multiple values at once instead of one by one, hence the extend instead of push/insert).
Intuitively, I would say it could be the least surprising to the user if it maps to how the docker CLI handles it, since most users are likely to be familiar with it.
- Is it possible to specify the flag multiple times to combine the permissions additively? If yes, extend, if not, replace.
- Is it possible to add/drop multiple permissions with a single instance of the flag? If yes, take `impl IntoIter<Item = String>`, if not, take `String`.
What do you think of that approach?
Of course I can also add documentation for this behaviour to the functions, regardless of the implementation we go with

DDtKey: Yeah, that’s all good points.
Only `with_cmd` actually overwrites the list (and it generally makes sense, I wanted it to be aligned with docker api).

AFAIR docker accepts 1 value per flag for cap-add/drop, but they are additive for sure. 

So we can go with the way how we expose ENV variables and mounts 🤔 Also, I guess it’s common case when only 1 capability is needed

Correct me if I’m wrong

DDtKey: Hi @nicmr 👋 

Let me know if you can’t (or don’t want) to finalize this last thing, please 🙏

nicmr: Sorry, I was on vacation, I'm going to finalize it this week. Thank you for your patience!

nicmr: Hi @DDtKey, sorry for the repeated delay, life got in the way. I've changed the signature to accept a single `impl Into<String>`. Thank you for the feedback so far!

DDtKey (approved):

PR #730, opened 2024-09-10
feat: support copy files to container

guenhter:
This PR supports copying files into the container like this:

```rust
let container = GenericImage::new("alpine", "latest")
        .with_wait_for(WaitFor::seconds(2))
        .with_copy_to("/tmp/somefile", "foobar".to_string().into_bytes())
        .start()
        .await?;
```

### DIFF
--- a/testcontainers/Cargo.toml
+++ b/testcontainers/Cargo.toml
@@ -38,6 +38,7 @@ signal-hook = { version = "0.3", optional = true }
 thiserror = "1.0.60"
 tokio = { version = "1", features = ["macros", "fs", "rt-multi-thread"] }
 tokio-stream = "0.1.15"
+tokio-tar = "0.3.1"
 tokio-util = { version = "0.7.10", features = ["io"] }
 url = { version = "2", features = ["serde"] }
 
--- a/testcontainers/src/core/client.rs
+++ b/testcontainers/src/core/client.rs
@@ -1,8 +1,14 @@
-use std::{io, str::FromStr};
+use std::{
+    io::{self},
+    str::FromStr,
+};
 
 use bollard::{
     auth::DockerCredentials,
-    container::{Config, CreateContainerOptions, LogOutput, LogsOptions, RemoveContainerOptions},
+    container::{
+        Config, CreateContainerOptions, LogOutput, LogsOptions, RemoveContainerOptions,
+        UploadToContainerOptions,
+    },
     errors::Error as BollardError,
     exec::{CreateExecOptions, StartExecOptions, StartExecResults},
     image::CreateImageOptions,
@@ -16,6 +22,7 @@ use url::Url;
 
 use crate::core::{
     client::exec::ExecResult,
+    copy::{CopyToContaienrError, CopyToContainer},
     env,
     env::ConfigurationError,
     logs::{
@@ -81,6 +88,10 @@ pub enum ClientError {
     InitExec(BollardError),
     #[error("failed to inspect exec command: {0}")]
     InspectExec(BollardError),
+    #[error("failed to upload data to container: {0}")]
+    UploadToContainerError(BollardError),
+    #[error("failed to prepare data for copy-to-container: {0}")]
+    CopyToContaienrError(CopyToContaienrError),
 }
 
 /// The internal client.
@@ -274,6 +285,32 @@ impl Client {
             .start_container::<String>(container_id, None)
             .await
             .map_err(ClientError::StartContainer)
+    }
+
+    pub(crate) async fn copy_to_container(
+        &self,
+        container_id: impl Into<String>,
+        copy_to_container: &CopyToContainer,
+    ) -> Result<(), ClientError> {
+        let container_id: String = container_id.into();
+        let target_directory = copy_to_container
+            .target_directory()
+            .map_err(ClientError::CopyToContaienrError)?;
+
+        let options = UploadToContainerOptions {
+            path: target_directory,
+            no_overwrite_dir_non_dir: "false".into(),
+        };
+
+        let tar = copy_to_container
+            .tar()
+            .await
+            .map_err(ClientError::CopyToContaienrError)?;
+
+        self.bollard
+            .upload_to_container::<String>(&container_id, Some(options), tar)
+            .await
+            .map_err(ClientError::UploadToContainerError)
     }
 
     pub(crate) async fn pull_image(&self, descriptor: &str) -> Result<(), ClientError> {
--- a/testcontainers/src/core/copy.rs
+++ b/testcontainers/src/core/copy.rs
@@ -1,0 +1,97 @@
+use std::{
+    io,
+    path::{self, Path, PathBuf},
+};
+
+#[derive(Debug, Clone)]
+pub struct CopyToContainer {
+    pub target: String,
+    pub source: CopyDataSource,
+}
+
+#[derive(Debug, Clone)]
+pub enum CopyDataSource {
+    File(PathBuf),
+    Data(Vec<u8>),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum CopyToContaienrError {
+    #[error("io failed with error: {0}")]
+    IoError(io::Error),
+    #[error("failed to get the path name: {0}")]
+    PathNameError(String),
+}
+
+impl CopyToContainer {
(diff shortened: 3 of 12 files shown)


testcontainers/src/core/client.rs
```diff
@@ -276,6 +290,55 @@ impl Client {
...
+        header.set_path(file_name).map_err(ClientError::TarError)?;
+        header.set_size(data.len() as u64);
+        header.set_mode(0o0644);
+        header.set_cksum();
+        ar.append(&mut header, data)
+            .map_err(ClientError::TarError)?;
+        let bytes = ar.into_inner().map_err(ClientError::TarError)?;
+
+        Ok(Bytes::copy_from_slice(bytes.as_slice()))
+    }
```
DDtKey, line 340: `tar` uses blocking api, we should avoid this by spawning blocking task at least.

Also it's not related to the `Client` itself

DDtKey (commented):
Hi there 👋 

Thank you a lot for the contribution 👍 

However it would be nice to rationalize the need and discuss the proposal before putting an effort 🙏 

So, I guess `mount`functionality doesn't cover your needs? You want a full independent copy of files, right? So we will have both mounting and copy features (like in [Java version](https://java.testcontainers.org/features/files/))

testcontainers/src/core/image/image_ext.rs
```diff
@@ -54,6 +57,13 @@ pub trait ImageExt<I: Image> {
     /// Adds a mount to the container.
     fn with_mount(self, mount: impl Into<Mount>) -> ContainerRequest<I>;
 
+    /// Copies some source into the container as file
+    fn with_copy_to<P: AsRef<std::path::Path>>(
+        self,
+        target: P,
```
DDtKey, line 63: Btw, I assume we shouldn't use `Path` as an argument for `target`. Let's say you're on Windows, but target is Unix-based container

guenhter:
Hi,

 the rational behind is specially when using Docker In Docker environments with the docker-socket mounted. Here, copying files to the container instead of mounting them is way easier and less error prone. This is one of the main reasons why this is needed.

testcontainers/src/core/client.rs
guenhter: I'll have a look at that.

testcontainers/src/core/image/image_ext.rs
guenhter: This is a very valid point. `String` it is then :)

testcontainers/src/core/client.rs
guenhter: If you don't mind, I then just just this crate for taring the files: https://github.com/dignifiedquire/async-tar

DDtKey: Yes, sure `async-tar` relies on `async-std`. We use tokio (at least for now, we don't provide a way to choose).

So I'd suggest to use https://crates.io/crates/tokio-tar (the same author, but for tokio)

DDtKey:
> the rational behind is specially when using Docker In Docker environments with the docker-socket mounted. Here, copying files to the container instead of mounting them is way easier and less error prone. This is one of the main reasons why this is needed.

I actually think it's a useful feature, so I don't mind for sure.
It's just always a shame and sad that a person has made a lot of effort, but the PR is not accepted for some other reason

That's why I prefer to have at least a short issue to have visibility

guenhter:
Completely understandable. Next time, I'll raise an issue first for discussion.

testcontainers/Cargo.toml
```diff
@@ -16,6 +16,8 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
+async-std = "1.13.0"
+async-tar = "0.5.0"
```
DDtKey, line 20: Please, take a look at https://github.com/testcontainers/testcontainers-rs/pull/730#discussion_r1752678907

We can use tokio-tar

guenhter: Oh sorry. I overlooked that. yes sure. I'll change that

guenhter: Done

testcontainers/src/core/copy.rs
```diff
@@ -0,0 +1,11 @@
...
+pub struct CopyToContainer {
+    pub target: String,
+    pub source: CopyDataSource,
+}
+
+#[derive(Debug, Clone)]
+pub enum CopyDataSource {
+    File(String),
+    Data(Vec<u8>),
+}
```
DDtKey, line 11: This one is nice, what I could suggest:

- we can use `Path` for source, because we going to copy file from the host/caller system
- we can implement `From<Path>` (file) and `From<T: Into<Vec<u8>>` (data)
- use `impl Into<CopyDataSource>` for arguments in `ImageExt`

It will provide more flexible way to interact with API
WDYT?

testcontainers/src/core/client.rs
```diff
@@ -81,6 +90,12 @@ pub enum ClientError {
     InitExec(BollardError),
     #[error("failed to inspect exec command: {0}")]
     InspectExec(BollardError),
+    #[error("failed to copy to container: {0}")]
+    CopyToContainer(BollardError),
+    #[error("failed to tar data for container: {0}")]
+    TarError(io::Error),
+    #[error("failed to get the path name: {0}")]
+    PathNameError(String),
```
DDtKey, line 98: I'd suggest to have one `CopyToContainerError` to encapsulate all the related errors into single one.

And here we can use transparent error

testcontainers/src/core/client.rs
```diff
@@ -276,6 +291,66 @@ impl Client {
...
+                ar.append_data(&mut header, file_name, data.as_slice())
+                    .await
+                    .map_err(ClientError::TarError)?;
+            }
+        }
+
+        let bytes = ar.into_inner().await.map_err(ClientError::TarError)?;
+
+        Ok(Bytes::copy_from_slice(bytes.as_slice()))
+    }
```
DDtKey, line 352: we actually can move this impl into `CopyDataSource`. It doesn't use the client and separate functionality in general

So it will be quite clear call:

`source.tar(target)`

WDYT?

testcontainers/src/core/copy.rs
guenhter: Very good idea. I imlemented a few `From<...>` but left await `From<T: Into<Vec<u8>>` because it didn't work right away.

testcontainers/src/core/client.rs
guenhter: That is a very good approach. I now put all errors related to prepare the data for container upload into a separate Error. Only the one `BollardError` which comes from the `docker.upload...` was handled separately, as it seems that all other bollard errors are also there.

testcontainers/src/core/client.rs
guenhter: That is a fantastic idea. I moved that function and also half of the parts of the other function to the `copy.rs` file. That feels now way better. Thx

testcontainers/src/core/client.rs
```diff
@@ -276,6 +291,66 @@ impl Client {
...
+                ar.append_file(target, &mut f)
+                    .await
+                    .map_err(ClientError::TarError)?;
+            }
+            CopyDataSource::Data(data) => {
+                let path = Path::new(&target);
+                let file_name = match path.file_name() {
+                    Some(v) => v,
+                    None => return Err(ClientError::PathNameError(target)),
+                };
```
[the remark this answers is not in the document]
guenhter: Hm, do you think this is a problem? On Windows machines, both path styles (windows and linux) are supported by `Path`. If the code runs on linux, only the linux way to handle path is supported by `Path`.

So,

- Linux -> Linux: should be fine
- Windows -> Linux: should be fine
- Windows -> Windows should be fine
- Linux -> Windows would be a problem, but can this even happen?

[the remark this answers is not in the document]
DDtKey: Yeah, good point.

> Linux -> Windows would be a problem, but can this even happen?

It's generally possible but unlikely.

I think we're ok to go with current solution until there is a demand. See no reason to block on it.

DDtKey (approved):
Thank you for the contribution and patience during the review process ❤️

guenhter:
Thanks for all the help. I realy appreciate your effort.

PR #735, opened 2024-09-17
feat: Support copying directories to container

guenhter:
Now not only files but also directories can be copied.

The only open thing here (marked with a TODO) is, if a windows container is used, that absolute paths are not supported (yet). For me personally this is at the moment ok because I don't use Windows containers and not even have a Windows machine to implement that. Maybe this can be done by someone (I guess it is just one line or so) when it is needed.


guenhter:
I already changed the PR title. The CI should succeed now...

DDtKey (commented):

I'll take a look soon, sorry for delay

DDtKey (approved):
Looks good to me, don't think windows containers should be a blocker, we can re-consider this on-demand

guenhter:
Thx. That sound very pragmatic to me.

guenhter:
@DDtKey are there plans to crate a new release so this feature can be used?

DDtKey:
Yes, we will release the new version this week. I only wanted to check if I have anything else to include

DDtKey:
I'm preparing release, just pushed small fix - found a typo in error definition.
It wIll be available soon, thank you for all efforts! 🙏

PR #736, opened 2024-09-18
refactor: replace dirs crate with etcetera

joonas:
I recently learned that [`dirs`](https://crates.io/crates/dirs) depends on [`option-ext`](https://crates.io/crates/option-ext) via [`dir-sys`](https://crates.io/crates/dirs-sys) (as of `dir-sys` 0.4.1 that came in as part of `dirs` 5.0.1), which is MPL 2.0 licensed.

[It does not appear that the author is interested reconsidering their stance on licensing](https://github.com/dirs-dev/dirs-sys-rs/issues/21#issuecomment-1531528155), which unfortunately creates some complications for projects under the CNCF umbrella (as well as many corporate projects), and so I thought I would propose switching `dirs` out for `etcetera` since it is MIT or Apache 2.0 licensed.

As you can see from the linked thread, this is a fairly common limitation for adoption, so in the interest of enabling further adoption, I believe it would be in the best interest for `testcontainers-rs` to reconsider this dependency, especially considering how trivial it is to replace 🙂 

FWIW, I have done testing on my end to ensure that the changes I'm proposing produce similar results under Mac OS and Linux, but since I don't have direct access to Windows I can't say with 100% certainty that they do though from reading through the code across both libraries, I feel fairly confident that they should.

### DIFF
--- a/testcontainers/Cargo.toml
+++ b/testcontainers/Cargo.toml
@@ -21,9 +21,9 @@ bollard = { version = "0.17.0", features = ["ssl"] }
 bollard-stubs = "=1.45.0-rc.26.0.1"
 bytes = "1.6.0"
 conquer-once = { version = "0.4", optional = true }
-dirs = "5.0.1"
 docker_credential = "1.3.1"
 either = "1.12.0"
+etcetera = "0.8.0"
 futures = "0.3"
 log = "0.4"
 memchr = "2.7.2"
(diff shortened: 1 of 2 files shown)


DDtKey (approved):
That's really a good catch!
Thank you 🚀 

I find this reasonable and probably we should consider usage [cargo-about](https://github.com/EmbarkStudios/cargo-about) in CI to prevent accidental usage

PR #746, opened 2024-10-12
docs: negative failure message instead of a postive one

Ghamza-Jd:
The message inside the expect is shown on failure, so updated the example to reflect that


DDtKey (approved):

DDtKey:
Thank you!

PR #747, opened 2024-10-13
docs: add examples of `ImageExt` usage and cover `GenericImage`

hovinen:
I struggled a bit when upgrading to testcontainers-rs 0.23, since the interfaces had changed somewhat since the last version I used. The documentation didn't explain exactly how to use the methods in `ImageExt` in conjunction with those of `GenericImage`.

This change adds a bit more content to the documentation:

* A few more lines in the README file to exercise the configuration methods in `ImageExt`,
* Some documentation for `GenericImage` showing how it is used along with `ImageExt`.

The latter in particular includes a doctest. If the API is changed in the future so that further changes are needed, the doctest will fail and can be updated accordingly. This will make it easier for downstream users to understand how to port their code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,13 +22,15 @@ The crate provides an API for working with containers in a test environment.
 - Blocking API (under `blocking` feature)
 
 ```rust
-use testcontainers::{core::{IntoContainerPort, WaitFor}, runners::SyncRunner, GenericImage};
+use testcontainers::{core::{IntoContainerPort, WaitFor}, runners::SyncRunner, GenericImage, ImageExt};
 
 #[test]
 fn test_redis() {
     let container = GenericImage::new("redis", "7.2.4")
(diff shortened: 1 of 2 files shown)


DDtKey:

We actually had a migration guide as part of 0.18, there were big changes: https://github.com/testcontainers/testcontainers-rs/releases/tag/0.18.0

But that's definitely not obvious place when we have several versions ahead. 
Perhaps it makes sense to add a reference in readme, or preferably move it to our [docs website](https://rust.testcontainers.org/)

DDtKey (approved):

hovinen:
Thanks for the quick approval!

It seems that the PR can't be merged yet due to some flakiness in the tests. I also don't see a way to retry the affected actions. Could you re-trigger them? Thank you.

Issue #742, opened 2024-09-26
Reuse containers


tisonkun:
Like https://java.testcontainers.org/features/reuse/

I currently use `bollard` directly to reuse containers between tests, while noticed that `testcontainers` encapsulation is valuable and ergonomic, like all the code to support `get_host_port_ipv4`.

So if `testcontainers` can support reuse containers, users like me can benefit from both features.

tisonkun:
My current code in gist - https://gist.github.com/tisonkun/c2ac3154bd6bea003c6097b195a9e32d

DDtKey:
Hi 👋 
Thank you for raising the issue!

Yes, it makes sense to support and should be simple

gurinderu:
How do you want to make a reuse ID? Should it be calculated automatically from the source code & call site meta or provided by hand?

gurinderu:
Also, it is not clear to understand when to stop the container, lets say we run nextest with the same reuse id in a serial way.  Nextest use process-per-test strategy and we don't have ryuk

DDtKey:
Reusable containers are not expected to be cleaned. 

Documentation from dotnet impl:
> Enabling reuse will disable the resource reaper, meaning the resource will not be cleaned up

and also from Java:
> Reusable containers are not suited for CI usage and as an experimental feature not all Testcontainers features are fully working (e.g., resource cleanup or networking).

You can check [java](https://java.testcontainers.org/features/reuse/) and [dotnet](https://dotnet.testcontainers.org/api/resource_reuse/) docs, because it describes how it works.

tisonkun:
> Reusable containers are not suited for CI usage

Just a comment to this statement: We reuse containers in the CI environment, and the runner will destroy everything once the job is finished. So it works well for our CI usage :D

DDtKey:
Yes, actually it should work if your runners are properly configured and don't store the state between different jobs. E.g DnD approach will work

the-wondersmith:
@DDtKey I've actually been working on implementing support for this. Would you accept a PR for the feature?

DDtKey:
@the-wondersmith I will be happy to review the PR.

Feel free to open 🙂

the-wondersmith:
> @the-wondersmith I will be happy to review the PR.
> 
> Feel free to open 🙂

Will do, as soon as I'm able 😁

DDtKey:
[...]

PR #756, opened 2024-11-20
feat: add support for user-configured labels

the-wondersmith:
PR adds support for applying user-configured labels to containers started by `testcontainers`.

> [!NOTE]
> This PR is a precursor to future support of [reusable containers](https://github.com/testcontainers/testcontainers-rs/issues/742)

### DIFF
--- a/testcontainers/src/core/image/image_ext.rs
+++ b/testcontainers/src/core/image/image_ext.rs
@@ -48,13 +48,18 @@ pub trait ImageExt<I: Image> {
     /// Sets the network the container will be connected to.
     fn with_network(self, network: impl Into<String>) -> ContainerRequest<I>;
 
+    /// Adds the specified label to the container.
+    ///
+    /// **Note**: all keys in the `org.testcontainers.*` namespace should be regarded
+    /// as reserved by `testcontainers` internally, and should not be expected or relied
+    /// upon to be applied correctly if supplied as a value for `key`.
+    fn with_label(self, key: impl Into<String>, value: impl Into<String>) -> ContainerRequest<I>;
+
     /// Adds the specified labels to the container.
     ///
-    /// **Note**: in addition to all keys in the `com.testcontainers.*` namespace, there
-    /// are certain labels that are used by `testcontainers` internally which will always
-    /// be unconditionally overwritten, and so should not be expected or relied upon to
-    /// be applied correctly if they are included in `labels`. Currently, they are:
-    /// - `managed-by`
+    /// **Note**: all keys in the `org.testcontainers.*` namespace should be regarded
+    /// as reserved by `testcontainers` internally, and should not be expected or relied
+    /// upon to be applied correctly if they are included in `labels`.
     fn with_labels(
         self,
         labels: impl IntoIterator<Item = (impl Into<String>, impl Into<String>)>,
@@ -174,6 +179,14 @@ impl<RI: Into<ContainerRequest<I>>, I: Image> ImageExt<I> for RI {
             network: Some(network.into()),
             ..container_req
         }
+    }
+
+    fn with_label(self, key: impl Into<String>, value: impl Into<String>) -> ContainerRequest<I> {
+        let mut container_req = self.into();
+
+        container_req.labels.insert(key.into(), value.into());
+
+        container_req
     }
 
     fn with_labels(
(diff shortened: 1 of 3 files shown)


testcontainers/src/core/image/image_ext.rs
```diff
@@ -48,6 +48,18 @@ pub trait ImageExt<I: Image> {
...
+    ///
+    /// **Note**: in addition to all keys in the `com.testcontainers.*` namespace, there
+    /// are certain labels that are used by `testcontainers` internally which will always
+    /// be unconditionally overwritten, and so should not be expected or relied upon to
+    /// be applied correctly if they are included in `labels`. Currently, they are:
+    /// - `managed-by`
+    fn with_labels(
+        self,
+        labels: impl IntoIterator<Item = (impl Into<String>, impl Into<String>)>,
+    ) -> ContainerRequest<I>;
```
DDtKey, line 61: Since it's append semantic, I'd suggest to add a singular form `with_label` (we may have both or just singular actually). Also singular methods seems aligned [with Java impl](https://java.testcontainers.org/features/advanced_options/#container-labels)

I mean a common use-case is to pass couple of labels in-place, but passing many of them also convinient with iterator so we can have both

WDYT?

testcontainers/src/runners/async_runner.rs
```diff
@@ -63,8 +63,17 @@ where
             .map(|(key, value)| format!("{key}:{value}"))
             .collect();
 
+        let labels = HashMap::<String, String>::from_iter(
+            container_req
+                .labels()
+                .iter()
+                .map(|(key, value)| (key.into(), value.into()))
+                .chain([("managed-by".to_string(), "testcontainers".to_string())]),
```
DDtKey, line 71: Actually, to avoid any collisions  and possible conflicts with user's code. I'd suggest to use predefined prefix for **all** our labels, including `managed-by`

- `org.testcontainers.managed-by` 

So we can claim that `org.testcontainers.*` prefix is dedicated for usage by testcontainers and it will be unlikely to collide

WDYT?

DDtKey (commented):

I find this pretty small and simple PR looking good to me, just a few comments to discuss and LGTM!  ✅

testcontainers/src/core/image/image_ext.rs
the-wondersmith: @DDtKey An excellent suggestion. Implemented and pushed 😁

testcontainers/src/runners/async_runner.rs
the-wondersmith: I don't have a strong enough argument *against* it, honestly, other than "the semantics / intent / meaning of `managed-by` or `app.docker.io/managed-by` *feels* clearer than `org.testcontainers.managed-by`", but "✨vibes✨" is absolutely not a valid reason to argue in this context.

Plus, as I was thinking about it, the counter-point advocate part of my brain floated it as a really nice / clean way to configure reaper behavior in the future, e.g.:

```rust
let managed_by = if !reusable && !ryuk {
    // reaper behavior controlled by standard `Drop` impl
    "testcontainers"
} else if reusable && !ryuk {
    // reaper behavior explicitly controlled by user code
    let managed_by = env!("CARGO_PKG_NAME");
} else if !reusable && ryuk {
    // reaper behavior controlled by `ryuk`
    let managed_by = "ryuk";
} else {
    // etc...
};

let labels = HashMap::<String, String>::from_iter(
    container_req
        .labels()
        .iter()
        .map(|(key, value)| (key.into(), value.into()))
        .chain([("org.testcontainers.managed-by".into(), managed_by.into())]),
);
```

So... yeah. Sold. Implemented and pushed 😁

the-wondersmith:
> Thank you a lot for the contribution! 🙏

Absolutely my pleasure 😁

> I find this pretty small and simple PR looking good to me, just a few comments to discuss and LGTM! ✅

I replied to both comments, and pushed the suggested changes. If the responses (and changes) make sense and to you look good, I'll leave the `Resolve Conversation` button to your discretion 😅

DDtKey (approved):

DDtKey:
Perfect, LGTM ✅ 

I'll take a look at the second PR soon, thanks one more time! ❤️

PR #757, opened 2024-11-22
feat: reusable containers

the-wondersmith:
PR implements reusable containers inline with [the existing](https://java.testcontainers.org/features/reuse/) reusable containers feature(s) of `testcontainers` for other languages.

> [!IMPORTANT]
> This PR follows from (and is branched from) #756. PR 756 *must* be merged prior to merging this PR

### DIFF
--- a/testcontainers/src/core/client.rs
+++ b/testcontainers/src/core/client.rs
@@ -447,19 +447,31 @@ impl Client {
         .collect::<HashMap<_, _>>();
 
         let options = Some(ListContainersOptions {
-            filters,
             all: false,
             size: false,
-            limit: Some(1),
+            limit: None,
+            filters: filters.clone(),
         });
 
-        Ok(self
+        let containers = self
             .bollard
             .list_containers(options)
             .await
-            .map_err(ClientError::ListContainers)?
+            .map_err(ClientError::ListContainers)?;
+
+        if containers.len() > 1 {
+            log::warn!(
+                "Found {} containers matching filters: {:?}",
+                containers.len(),
+                filters
+            );
+        }
+
+        Ok(containers
             .into_iter()
-            .next()
+            // Use `max_by_key()` instead of `next()` to ensure we're
+            // returning the id of most recently created container.
+            .max_by_key(|container| container.created.unwrap_or(i64::MIN))
             .and_then(|container| container.id))
     }
 }
--- a/testcontainers/src/core/client.rs
+++ b/testcontainers/src/core/client.rs
@@ -1,4 +1,5 @@
 use std::{
+    collections::HashMap,
     io::{self},
     str::FromStr,
 };
@@ -6,8 +7,8 @@ use std::{
 use bollard::{
     auth::DockerCredentials,
     container::{
-        Config, CreateContainerOptions, LogOutput, LogsOptions, RemoveContainerOptions,
-        UploadToContainerOptions,
+        Config, CreateContainerOptions, ListContainersOptions, LogOutput, LogsOptions,
+        RemoveContainerOptions, UploadToContainerOptions,
     },
     errors::Error as BollardError,
     exec::{CreateExecOptions, StartExecOptions, StartExecResults},
@@ -66,6 +67,8 @@ pub enum ClientError {
     #[error("failed to map ports: {0}")]
     PortMapping(#[from] PortMappingError),
 
+    #[error("failed to list containers: {0}")]
+    ListContainers(BollardError),
     #[error("failed to create a container: {0}")]
     CreateContainer(BollardError),
     #[error("failed to remove a container: {0}")]
@@ -416,6 +419,48 @@ impl Client {
         };
 
         Some(bollard_credentials)
+    }
+
+    /// Get the `id` of the first running container whose `name`, `network`,
+    /// and `labels` match the supplied values
+    #[cfg_attr(not(feature = "reusable-containers"), allow(dead_code))]
+    pub(crate) async fn get_running_container_id(
+        &self,
+        name: Option<&str>,
+        network: Option<&str>,
+        labels: &HashMap<String, String>,
+    ) -> Result<Option<String>, ClientError> {
+        let filters = [
+            Some(("status".to_string(), vec!["running".to_string()])),
+            name.map(|value| ("name".to_string(), vec![value.to_string()])),
+            network.map(|value| ("network".to_string(), vec![value.to_string()])),
+            Some((
+                "label".to_string(),
+                labels
+                    .iter()
+                    .map(|(key, value)| format!("{key}={value}"))
+                    .collect(),
+            )),
+        ]
+        .into_iter()
+        .flatten()
+        .collect::<HashMap<_, _>>();
+
+        let options = Some(ListContainersOptions {
+            filters,
+            all: false,
+            size: false,
+            limit: Some(1),
+        });
+
+        Ok(self
+            .bollard
+            .list_containers(options)
+            .await
+            .map_err(ClientError::ListContainers)?
+            .into_iter()
+            .next()
+            .and_then(|container| container.id))
     }
 }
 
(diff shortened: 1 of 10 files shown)


the-wondersmith:
@DDtKey Dug into the test failures, and it *looks* like it's just `cargo hack` *not* including `experimental` when it tries to activate the `reusable-containers` feature.

I updated the CI action to fix it, but (for obvious reasons) GHA won't run actions from forks / PR branches (usually).

Not sure how to proceed here, got any thoughts?

the-wondersmith:
> Actually, I think we can avoid usage of `experimental` feature. It's already under separate feature and we can just claim in documentation that it's "experimental" functionality. It's not enabled by default and users anyway should read something about the feature before enabling it.

Fair enough. Removed the `experimental` feature and reverted the CI changes 🙂

DDtKey:
I'll take a deeper look at the PR today, sorry again for the delay

testcontainers/src/runners/async_runner.rs
```diff
@@ -20,6 +20,25 @@ use crate::{
...
+/// For reference: without a unique-per-session identifier, containers created by
+/// previous test sessions that were marked `reuse`, (or where the test suite was
+/// run with the `TESTCONTAINERS_COMMAND` environment variable set to `keep`), that
+/// *haven't* been manually cleaned up could be incorrectly returned from methods
+/// like [`Client::get_running_container_id`](Client::get_running_container_id),
+/// as the container name, labels, and network would all still match.
+#[cfg(feature = "reusable-containers")]
+pub(crate) fn session_id() -> &'static ulid::Ulid {
+    TESTCONTAINERS_SESSION_ID.get_or_init(ulid::Ulid::new)
+}
```
DDtKey, line 41: Not sure I get the `session_id` intent here - I mean it's definitely can be useful (for example to sort them if many)

But why do we want to filter by `session_id` label? 
The idea of reusable containers is to be able to use in next test run again

E.g: 
```sh
cargo test  # `session1`: creates containers
# do something
cargo test # `session2`: reuses containers of `session1`
```

Moreover, I might want to run parallel test suites with the same containers (I explicitly enabled re-usability)

testcontainers/src/core/client.rs
```diff
@@ -417,6 +420,48 @@ impl Client {
...
+        });
+
+        Ok(self
+            .bollard
+            .list_containers(options)
+            .await
+            .map_err(ClientError::ListContainers)?
+            .into_iter()
+            .next()
+            .and_then(|container| container.id))
```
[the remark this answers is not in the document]
the-wondersmith: I don't disagree at all; just to clarify though (cause it's super easy to miss 😅) - there's a [limit](https://github.com/testcontainers/testcontainers-rs/blob/23bdbfd0662813fae1c6e0f6c13466df9cfef9a6/testcontainers/src/core/client.rs#L453) of `1` applied to the filter options passed to the docker socket. In theory, there can only ever be either zero or one results. I 100% agree with the defensive coding spirit though, and am obviously happy to refactor accordingly.

Way-forward-wise, do you think we can trust in `bollard` (and docker's) filtration logic and Keep It Simple™ (i.e. leave it as is)? Alternatively, do you think we should remove the limit and add something like:

```rust
{

  // ...

  let mut containers = self
      .bollard
      .list_containers(options)
      .await
      .map_err(ClientError::ListContainers)?
      .into_iter();

  let container_id = containers.next().and_then(|container| container.id);

  if let Some(container) = containers.next() {
      warn!("insert some warning here")
  }

  Ok(container_id)

}
```

testcontainers/src/runners/async_runner.rs
the-wondersmith: Ah - this *originally* came about after I ran into some trouble with the test suite failing and subsequently trying to reuse some containers *incorrectly* and then causing failures on *subsequent* runs.

I think a unique session identifier *is* a good idea / feature, memory is failing me as to why I didn't this up the same parallel test suite edge case you did though 🤔.

In either case, what do you think about keeping the ability, but making it's inclusion in filtration user-configurable? e.x.:

```rust
let reusable_container = GenericImage::new("alpine", "latest").with_reuse(true).start().await?;

let session_unique_container = GenericImage::new("alpine", "latest").with_reuse(true).current_session(true).start().await?;
```

That, or change the type of `ContainerRequest.reuse` from `bool` to something like `Option<ReuseDirective>`, e.x.:

```rust
pub enum ReuseDirective {
    Always,
    CurrentSession,
}
```

DDtKey: I think `enum` option is a perfect fit here actually, because I see all of these options useful

testcontainers/src/core/client.rs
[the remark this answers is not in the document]
the-wondersmith: I like it. One small point - remove `limit` and in our check `Always` == latest, `CurrentSession` == container w/ the current session id, right?

testcontainers/src/runners/async_runner.rs
the-wondersmith: Yup, the more I thought about it after leaving the comment, the more I liked the idea.

Implemented it, pushed the changes. Please take a look at your convenience and let me know what you think 😁

testcontainers/src/core/client.rs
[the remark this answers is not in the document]
the-wondersmith: Applied the suggestion along with the changes discussed [below](https://github.com/testcontainers/testcontainers-rs/pull/757#discussion_r1879041342).

Let me know what you think 😁

testcontainers/src/core/containers/request.rs
```diff
@@ -234,27 +244,41 @@ impl PortMapping {
...
+            }
+            #[cfg(feature = "reusable-containers")]
+            {
+                |repr: &mut std::fmt::DebugStruct<'_, '_>| -> std::fmt::Result {
+                    repr.field("reusable", &self.reuse).finish()
+                }
+            }
+        };
+
+        with_feature_flag_fields(
```
DDtKey, line 260: Tbh, I prefer another pattern for simple use-cases:

```
let debug_struct = f.debug_struct(..); // all common ones are omitted 

// Utilize shadowing if the feature is enabled
#[cfg(feature = "reusable-containers")]
let debug_struct = debug_struct.field(“reusable”, ..);

debug_struct.finish()
```

testcontainers/src/core/image/image_ext.rs
DDtKey, line 1: Not sure about `boolean` and `string` conversions. I mean it's ambiguous and less expendable.
"true" may confuse if we improve `CurrentSession` behavior.

Let's probably stick to enum initially and won't provide conversions for now. Also it's more user-friendly - IDE/LSP can suggest enum options, but not custom strings. 
+ harder to provide backward compatibility


It's stricter policy, and we can always relax once have better understanding and feedback

DDtKey (approved):
I like what you've built here. Thank you a lot and sorry for all the delays 🙏 

Only few minor comments and LGTM

Next milestones for this feature are: 
- env-variable control
- proper reaper for "current session" (ryuk)

testcontainers/src/core/containers/request.rs
the-wondersmith: FTR, same. There's a problem with lifetimes in this instance though, and the borrow-checker doesn't seem to be able to be made happy any other way 🥲

Screenshot 2024-12-16 at 9 32 52 AM

Screenshot 2024-12-16 at 9 33 03 AM

testcontainers/src/core/image/image_ext.rs
the-wondersmith: Made the constraint on `with_reuse` method stricter, pushed the change 🙂

testcontainers/src/core/containers/request.rs
the-wondersmith: **_WAIT_**, disregard that - brain wasn't working 🤦. Took a quick reading comprehension lesson and tried again:

Screenshot 2024-12-16 at 9 50 24 AM

DDtKey (approved):

PR #774, opened 2025-03-04
feat: docker-compose support

DDtKey:
## Description

Initial attempt to introduce docker-compose support.

### Status

- work-in-progress

### DIFF
--- a/testcontainers/src/compose/client.rs
+++ b/testcontainers/src/compose/client.rs
@@ -1,0 +1,64 @@
+use std::{fmt, path::PathBuf};
+
+use crate::core::async_container::raw::RawContainer;
+
+pub(super) mod containerised;
+pub(super) mod local;
+
+pub(super) enum ComposeClient {
+    Local(local::LocalComposeCli),
+    Containerised(containerised::ContainerisedComposeCli),
+}
+
+impl ComposeClient {
+    pub(super) fn new_local(compose_files: Vec<PathBuf>) -> Self {
+        ComposeClient::Local(local::LocalComposeCli::new(compose_files))
+    }
+
+    pub(super) async fn new_containerised(compose_files: Vec<PathBuf>) -> Self {
+        ComposeClient::Containerised(
+            containerised::ContainerisedComposeCli::new(compose_files).await,
+        )
+    }
+}
+
+pub(super) struct UpCommand {
+    pub(super) project_name: String,
+    pub(super) wait_timeout: std::time::Duration,
+}
+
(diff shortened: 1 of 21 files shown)


testcontainers/src/core/containers/async_container/raw.rs
DDtKey, line 1: You can notice there is a code extracted into `RawContainer` - it's not strong opinion, but attempt to expose the same functionality for containers created by compose wrapper (e.g logs, exec commands and etc) and for currently existing async-container.

Worth noting that I'd prefer not to expose `RawContainer` for now and just utilize delegation. At least for the first version

We also could consider adding a trait (e.g `compose.get_container("service-name") -> impl ContainerOperations`), but we can achieve backward compatibility for `AsyncContainer` via delegation.

testcontainers/src/compose/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+    async fn test_local_docker_compose() {
+        let path_to_compose = PathBuf::from(format!(
+            "{}/tests/test-compose.yml",
+            env!("CARGO_MANIFEST_DIR")
+        ));
+        let docker_compose = DockerCompose::with_local_client(&[path_to_compose.as_path()]);
+        docker_compose.up().await;
+        let client = reqwest::get("http://localhost:8081").await.unwrap();
+        tokio::time::sleep(std::time::Duration::from_secs(15)).await;
+    }
```
DDtKey, line 131: Both tests work already, but they should be "serial" to avoid a port conflict.
Just not finished because there is no stable interface to access containers from compose

testcontainers/src/compose/client.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub(super) volumes: bool,
+}
+
+pub(super) trait ComposeInterface {
+    async fn up(&self, command: UpCommand) -> Result<(), std::io::Error>;
+    async fn down(&self, command: DownCommand) -> Result<(), std::io::Error>;
+}
+
+impl ComposeInterface for ComposeClient {
+    async fn up(&self, command: UpCommand) -> Result<(), std::io::Error> {
```
DDtKey, line 42: As an option, it should return a list of "raw" containers. 
So we can expose method `compose.get_container(service_name) -> &RawContainer` (or something like that)

```suggestion
    async fn up(&self, command: UpCommand) -> Result<Vec<RawContainer>, std::io::Error> {
```

DDtKey:
Closed in favor of updated branch - #864

PR #776, opened 2025-03-10
Please add cpu and memory limits

bitemyapp:


DDtKey:
Sorry, I somehow missed the PR.
Thank you for the contribution.

This change introduces quite a lot of options, I don't even think `testcontainers` in other languages allows to change the defaults for these options.  Not that common use-case for testing I guess.

However - the docker client is usually exposed, so it's possible to update the container's config via [`update_container`](https://docs.rs/bollard/latest/bollard/struct.Docker.html#method.update_container)

jayakasadev:
This is definitely a useful change for testing with containers in constrained environments or when containers require licenses to run above certain resource limits.

Edit: 
This is functionality that's already provided in [other languages](https://node.testcontainers.org/features/containers/#with-resources-quota).

bitemyapp:
Rebased at #890

PR #779, opened 2025-03-22
feat: add option to stop containers with a timeout

o087D:
Previously containers were stopped with no timeout (SIGKILL) which was interfering with clean shutdown of processes being tested. 

This commit adds a new API of stop_with_timeout(timeout_secconds) in addition to the original stop(). The original stop() interface calls stop_with_timeout using the default '0'.

EDIT:

It would seem my original diagnosis was incorrect and the default timeout is 10 seconds, but my CI system was somehow setting the default to 1 second. I will fix the docs and put in the correct default. I think there is still some value in here - apologies for the time waste getting to this point!

### DIFF
--- a/testcontainers/src/core/containers/async_container.rs
+++ b/testcontainers/src/core/containers/async_container.rs
@@ -183,7 +183,7 @@ where
         let network_settings = self.docker_client.inspect_network(&network_mode).await?;
 
         network_settings.driver.ok_or_else(|| {
-            TestcontainersError::other(format!("network {} is not in bridge mode", network_mode))
+            TestcontainersError::other(format!("network {network_mode} is not in bridge mode"))
         })?;
 
         let container_network_settings = container_settings
--- a/testcontainers/src/core/containers/async_container.rs
+++ b/testcontainers/src/core/containers/async_container.rs
@@ -280,11 +280,27 @@ where
         Ok(())
     }
 
-    /// Stops the container (not the same with `pause`).
+    /// Stops the container (not the same with `pause`) using the default 10 second timeout
     pub async fn stop(&self) -> Result<()> {
-        log::debug!("Stopping docker container {}", self.id);
+        self.stop_with_timeout(None).await?;
+        Ok(())
+    }
 
-        self.docker_client.stop(&self.id).await?;
+    /// Stops the container with timeout before issuing SIGKILL (not the same with `pause`).
+    ///
+    /// Set Some(-1) to wait indefinitely, None to use system configured default and Some(0)
+    /// to forcibly stop the container immediately - otherwise the runtime will issue SIGINT
+    /// and then wait timeout_seconds seconds for the process to stop before issuing SIGKILL.
+    pub async fn stop_with_timeout(&self, timeout_seconds: Option<i64>) -> Result<()> {
+        log::debug!(
+            "Stopping docker container {} with {} second timeout",
+            self.id,
+            timeout_seconds
+                .map(|t| t.to_string())
+                .unwrap_or("'system default'".into())
+        );
+
+        self.docker_client.stop(&self.id, timeout_seconds).await?;
         Ok(())
     }
 
--- a/testcontainers/src/watchdog.rs
+++ b/testcontainers/src/watchdog.rs
@@ -34,7 +34,7 @@ static WATCHDOG: Lazy<Mutex<Watchdog>> = Lazy::new(|| {
                     .unwrap_or_default()
                 {
                     signal_docker
-                        .stop(&container_id)
+                        .stop(&container_id, None)
                         .await
                         .expect("failed to stop container");
                     signal_docker
--- a/testcontainers/src/core/client.rs
+++ b/testcontainers/src/core/client.rs
@@ -162,9 +162,16 @@ impl Client {
             .map_err(ClientError::RemoveContainer)
     }
 
-    pub(crate) async fn stop(&self, id: &str) -> Result<(), ClientError> {
+    pub(crate) async fn stop(
+        &self,
+        id: &str,
+        timeout_seconds: Option<i64>,
+    ) -> Result<(), ClientError> {
         self.bollard
-            .stop_container(id, None)
+            .stop_container(
+                id,
+                timeout_seconds.map(|t| bollard::container::StopContainerOptions { t }),
+            )
             .await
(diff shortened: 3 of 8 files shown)


DDtKey:

I've fixed PR name
Could you, please, fix clippy & compilation warnings and apply rustfmt formatting (nightly)? (see CI)

o087D:
Apologies for that, not sure what happened with the linter and forgot to enable all features!

DDtKey (approved):

testcontainers/src/core/containers/async_container.rs
```diff
@@ -282,9 +282,22 @@ where
 
     /// Stops the container (not the same with `pause`).
     pub async fn stop(&self) -> Result<()> {
-        log::debug!("Stopping docker container {}", self.id);
+        self.stop_with_timeout(0).await?;
```
mervyn-mccreight, line 285: Shouldn't it be `-1` here according to the documentation of `stop_with_timeout` to stay the same?

What would be the difference between `0` and `-1` here?

testcontainers/src/watchdog.rs
```diff
@@ -34,7 +34,7 @@ static WATCHDOG: Lazy<Mutex<Watchdog>> = Lazy::new(|| {
                     .unwrap_or_default()
                 {
                     signal_docker
-                        .stop(&container_id)
+                        .stop(&container_id, 0)
```
mervyn-mccreight, line 37: Same question as here: https://github.com/testcontainers/testcontainers-rs/pull/779/files#r2014926568

testcontainers/src/core/containers/async_container.rs
o087D: It is actually undefined as far as I can tell.

Bollard uses https://github.com/fussybeaver/bollard/blob/master/src/container.rs#L305 which derived default, and implements stop as https://github.com/fussybeaver/bollard/blob/master/src/container.rs#L1337. This calls the build command in https://github.com/fussybeaver/bollard/blob/master/src/docker.rs#L1335 and which generates the query path using https://github.com/fussybeaver/bollard/blob/master/src/uri.rs#L24.

The option was set to None in the original testcontainers-rs stop command.

So, the query path should be empty if None was set.

The Docker API is defined as https://docs.docker.com/reference/api/engine/version/v1.48/#tag/Container/operation/ContainerStop - the stop time is not required.

The OpenAPI specification is:

```
  /containers/{id}/stop:
    post:
      summary: "Stop a container"
      operationId: "ContainerStop"
      responses:
        204:
          description: "no error"
        304:
          description: "container already stopped"
        404:
          description: "no such container"
          schema:
            $ref: "#/definitions/ErrorResponse"
          examples:
            application/json:
              message: "No such container: c2ada9df5af8"
        500:
          description: "server error"
          schema:
            $ref: "#/definitions/ErrorResponse"
      parameters:
        - name: "id"
          in: "path"
          required: true
          description: "ID or name of the container"
          type: "string"
        - name: "signal"
          in: "query"
          description: |
            Signal to send to the container as an integer or string (e.g. `SIGINT`).
          type: "string"
        - name: "t"
          in: "query"
          description: "Number of seconds to wait before killing the container"
          type: "integer"
      tags: ["Container"]
```

My assumption is that the default integer is 0 and that -1 may change how things are processed.

o087D: Actually, my bad - the -1 was from an example somewhere, I have it in my written notes. But it must have been wrong, or maybe an older implementation did things differently.

Digging (much) deeper the actual answer is https://github.com/moby/moby/blob/master/api/types/container/config.go#L18.

```
	// Timeout (optional) is the timeout (in seconds) to wait for the container
	// to stop gracefully before forcibly terminating it with SIGKILL.
	//
	// - Use nil to use the default timeout (10 seconds).
	// - Use '-1' to wait indefinitely.
	// - Use '0' to not wait for the container to exit gracefully, and
	//   immediately proceeds to forcibly terminating the container.
	// - Other positive values are used as timeout (in seconds).
```

I have just tried it locally and it is taking 10s before stopping - inspecting the CI logs the runners are exiting containers after 1 second, not 100% sure how that is possible.

However using the above timeout patch does work to configure to value to a specific timeout value.

This does swing this issue to being a me problem rather than something which would have value to many people. I will fix my docs and put in the 10s default.

DDtKey: This a good catch actually:

> The default timeout can be specified using the --stop-timeout option when creating the container. If no default is configured for the container, the Daemon determines the default, and is 10 seconds for Linux containers, and 30 seconds for Windows containers.

Which is applicable for stop request:

https://docs.docker.com/reference/cli/docker/container/stop/

DDtKey: Since it depends on the platform and docker engine, perhaps we can pass `None` instead? Preserving the default behavior

o087D:
I think this is now there! there is a new clippy lint I dont think I want to get into:

```
warning: large size difference between variants
  --> testcontainers/src/core/wait/mod.rs:32:1
   |
32 | / pub enum WaitFor {
33 | |     /// An empty condition. Useful for default cases or fallbacks.
34 | |     Nothing,
35 | |     /// Wait for a certain message to appear in the container's logs.
36 | |     Log(LogWaitStrategy),
   | |     -------------------- the second-largest variant contains at least 48 bytes
...  |
44 | |     Http(HttpWaitStrategy),
   | |     ---------------------- the largest variant contains at least 272 bytes
45 | |     /// Wait for the container to exit.
46 | |     Exit(ExitWaitStrategy),
47 | | }
   | |_^ the entire enum is at least 272 bytes
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
   = note: `#[warn(clippy::large_enum_variant)]` on by default
help: consider boxing the large fields to reduce the total size of the enum
   |
44 -     Http(HttpWaitStrategy),
44 +     Http(Box<HttpWaitStrategy>),
   |

warning: `testcontainers` (lib) generated 1 warning
```

I also want to test this on my problematic container just to make sure

DDtKey (approved):
Really appreciate it!

Sorry for the delay, we probably should have fix the lints in separate PR before.

But I don't mind merging as part of this PR

DDtKey (approved):

PR #781, opened 2025-03-24
build: propagate ring/aws-lc-rs feature flags to bollard

kaitlinsm:
In order for consumers to be able to switch between `ring` and `aws-lc-rs` cleanly without `testcontainers_rs` bringing in the wrong library, add feature flags that are propagated through to `bollard` to switch between the two ssl providers.

### DIFF
--- a/testcontainers/Cargo.toml
+++ b/testcontainers/Cargo.toml
@@ -17,7 +17,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 async-trait = { version = "0.1" }
-bollard = { version = "0.18.1", features = ["ssl"] }
+bollard = { version = "0.18.1"}
 bollard-stubs = "=1.47.1-rc.27.3.1"
 bytes = "1.6.0"
 conquer-once = { version = "0.4", optional = true }
@@ -44,7 +44,10 @@ ulid = { version = "1.1.3", optional = true }
 url = { version = "2", features = ["serde"] }
 
 [features]
-default = []
+default = ["ssl"]
+ssl = ["bollard/ssl"]
+aws-lc-rs = ["bollard/aws-lc-rs"]
+ssl_providerless = ["bollard/ssl_providerless"]
 blocking = []
 watchdog = ["signal-hook", "conquer-once"]
 http_wait = ["reqwest"]
--- a/testcontainers/Cargo.toml
+++ b/testcontainers/Cargo.toml
@@ -44,10 +44,10 @@ ulid = { version = "1.1.3", optional = true }
 url = { version = "2", features = ["serde"] }
 
 [features]
-default = ["ssl"]
-ssl = ["bollard/ssl"]
+default = ["ring"]
+ring = ["bollard/ssl"]
 aws-lc-rs = ["bollard/aws-lc-rs"]
-ssl_providerless = ["bollard/ssl_providerless"]
+ssl = ["bollard/ssl_providerless"]
 blocking = []
 watchdog = ["signal-hook", "conquer-once"]
 http_wait = ["reqwest"]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
         with:
           tool: cargo-hack
       - name: Build
-        run: cargo hack build --each-feature --keep-going
+        run: cargo hack build --at-least-one-of ring,aws-lc-rs,ssl --feature-powerset --depth 2 --keep-going
 
   test:
     name: Test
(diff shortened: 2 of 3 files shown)


testcontainers/Cargo.toml
```diff
@@ -44,7 +44,10 @@ ulid = { version = "1.1.3", optional = true }
 url = { version = "2", features = ["serde"] }
 
 [features]
-default = []
+default = ["ssl"]
+ssl = ["bollard/ssl"]
+aws-lc-rs = ["bollard/aws-lc-rs"]
+ssl_providerless = ["bollard/ssl_providerless"]
```
DDtKey, line 50: both `ssl` and `aws-lc-rs` enables `ssl_providerless` on `bollard` level

Looks like there is no reason to expose such feature

testcontainers/Cargo.toml
```diff
@@ -44,7 +44,10 @@ ulid = { version = "1.1.3", optional = true }
 url = { version = "2", features = ["serde"] }
 
 [features]
-default = []
+default = ["ssl"]
+ssl = ["bollard/ssl"]
+aws-lc-rs = ["bollard/aws-lc-rs"]
```
DDtKey, line 49: I'm also not really sure if we should follow bollard's convention here, because in fact it's just a switch between providers - `ring` and `aws-lc-rs` 🤔 

Bollard is the most important dependency though. However their choice of feature-names may have historical reasons (e.g backward compatible change)

WDYT? Perhaps `ring` and `aws-lc-rs` (or maybe even prefixed with `ssl-*`)?

DDtKey:
Also test suite relies on cargo hack. We need to adjust CI to have ssl feature enabled

DDtKey:
Thank you for the contribution 👍

testcontainers/Cargo.toml
kaitlinsm: Ah, good spot.

The way I read the docs https://docs.rs/bollard/latest/bollard/#feature-flags

> ssl_providerless: enable SSL support through [Rustls](https://github.com/rustls/rustls) without installing a [CryptoProvider](https://docs.rs/rustls/0.23.12/rustls/crypto/struct.CryptoProvider.html). You are responsible to do so.

suggested to me that the three feature flags (`ssl|aws-lc-rs|ssl_providerless`) are  mutually exclusive - but actually reading the [bollard Cargo.toml](https://github.com/fussybeaver/bollard/blob/526f9b8bdbdf58ebbed898d184fbfb4fcef29e69/Cargo.toml#L40) shows that that's wrong.

I'll remove the extra flag.

testcontainers/Cargo.toml
kaitlinsm: I did think about that one a bit, and initially decided to "propagate" the flags to the library we're consuming here.

The `rustls` ecosystem's crates have feature flags named after the libraries:
* [hyper-rustls](https://github.com/rustls/hyper-rustls?tab=readme-ov-file#crate-features)
* [rustls](https://docs.rs/rustls/latest/rustls/#crate-features)

A bit of digging shows that bollard has [used the ssl flag for over 9 years](https://github.com/fussybeaver/bollard/blame/v0.0.43/Cargo.toml) - only [adding support for aws-lc-rs](https://github.com/fussybeaver/bollard/commit/90cab8a8beb71ecd77c552850f7ba1fcd00227f5) more recently.

So `ssl` was "let's use SSL via rustls' default config", and the ring/aws-lc-rs split has happened a lot later than that.

I think `ring|aws-lc-rs` - to follow suit with `rustls` instead of `bollard` is the way to go.

DDtKey: Yes, that makes total sense to me. 
Let's align with rustls ecosystem since we don't have any inheritance here

Thank you!

testcontainers/Cargo.toml
kaitlinsm: If we just allow either `ring` or `aws-lc-rs`, it will prevent a consumer of `testcontainers` from being able to configure their own ssl provider.

The thing that's a bit unpleasant about this is that the feature flags are getting really verbose, but it feels like if I don't add the flexibility now, I'll be leaving a time bomb for later when the next ssl library comes along.

I'd suggest the following feature flags for `testcontainers`:

```
default = ["ring"]
ring = ["bollard/ssl"]
aws-lc-rs = ["bollard/aws-lc-rs"]
# use this feature flag if you want to use an SSL Provider _other_ than `ring` or `aws-lc-rs`
ssl = ["bollard/ssl_providerless"]

DDtKey: Agree, makes sense

DDtKey (approved):

PR #782, opened 2025-03-30
feat: added CommandStrategy with SuccessfulCommand enum

### LINKED ISSUE #702: WaitFor command exit with 0
Some image, like Postgres, would have a best match wait strategy like `pg_isready` returns 0.

Maybe we can implement such a wait strategy for testing with command?

sutt0n:
Potentially resolves #702 

This introduces a new wait strategy, `CommandStrategy`. The goal here is to wait for a successful command (exit code 0) prior to marking a container as ready. This polls the container state to ensure it's running first, and then executes the given command. Once the command has an exit code of 0, then the strategy has completed.

There are options to fail fast, as well as customize the exit code, just to expose flexibility for the strategy.

Open to changes / updates, if required -- this language is not my day-to-day.

### DIFF
--- a/testcontainers/src/core/wait/command_strategy.rs
+++ b/testcontainers/src/core/wait/command_strategy.rs
@@ -1,0 +1,110 @@
+use std::time::Duration;
+
+use crate::{
+    core::{client::Client, error::WaitContainerError, wait::WaitStrategy, ExecCommand},
+    ContainerAsync, Image,
+};
+
+#[derive(Debug, Clone)]
+pub struct CommandStrategy {
+    expected_code: i64,
+    poll_interval: Duration,
+    command: ExecCommand,
+    fail_fast: bool,
+}
+
+impl CommandStrategy {
+    /// Create a new `CommandStrategy` with default settings.
+    pub fn new() -> Self {
+        Self {
+            command: ExecCommand::default(),
+            expected_code: 0,
+            poll_interval: Duration::from_millis(100),
+            fail_fast: false,
+        }
+    }
+
+    /// Creates a new `CommandStrategy` with default settings and a preset command to execute.
+    pub fn command(command: ExecCommand) -> Self {
+        CommandStrategy::default().with_exec_command(command)
+    }
+
+    /// Set the fail fast flag for the strategy, meaning that if the command's first run does not
+    /// have the expected exit code, the strategy will exit with failure. If the flag is not set,
+    /// the strategy will continue to poll the container until the expected exit code is reached.
+    pub fn with_fail_fast(mut self, fail_fast: bool) -> Self {
+        self.fail_fast = fail_fast;
+        self
+    }
+
+    /// Set the command for executing the command on the container.
+    pub fn with_exec_command(mut self, command: ExecCommand) -> Self {
+        self.command = command;
+        self
+    }
+
+    /// Set the poll interval for checking the container's status.
+    pub fn with_poll_interval(mut self, poll_interval: Duration) -> Self {
+        self.poll_interval = poll_interval;
+        self
+    }
+}
+
+impl WaitStrategy for CommandStrategy {
+    async fn wait_until_ready<I: Image>(
+        self,
+        client: &Client,
+        container: &ContainerAsync<I>,
+    ) -> crate::core::error::Result<()> {
+        loop {
+            let container_state = client
+                .inspect(container.id())
+                .await?
+                .state
+                .ok_or(WaitContainerError::StateUnavailable)?;
+
+            let is_running = container_state.running.unwrap_or_default();
+
+            if is_running {
+                let exec_result = client
+                    .exec(container.id(), self.command.clone().cmd)
+                    .await?;
+
+                let inspect_result = client.inspect_exec(&exec_result.id).await?;
+                let mut running = inspect_result.running.unwrap_or(false);
+
+                loop {
+                    if !running {
+                        break;
+                    }
+
+                    let inspect_result = client.inspect_exec(&exec_result.id).await?;
+                    let exit_code = inspect_result.exit_code;
+                    running = inspect_result.running.unwrap_or(false);
+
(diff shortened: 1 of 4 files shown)


testcontainers/src/core/wait/command_strategy.rs
DDtKey, line 1: Can't this be achieved with [`exec_after_start`](https://docs.rs/testcontainers/0.23.3/testcontainers/core/trait.Image.html#method.exec_after_start) with `ExitCode` wait condition ([it waits for exit](https://github.com/testcontainers/testcontainers-rs/blob/8168a824d6bb12ef03681640f03e167f34a3b14d/testcontainers/src/core/containers/async_container.rs#L242-L258))? Additionally, this way allows to rely on a container's state

Or we could consider extending existing wait conditions if it's not enough

DDtKey (commented):

I've left a comment, please take a look

testcontainers/src/core/wait/command_strategy.rs
sutt0n: At second glance, it feels like duplicate functionality. This was probably an oversight on my end by lack of context.

I can test what you've provided and report back findings - that may aid in either resolving or progressing a solution for this issue.

DDtKey: We don't have to be fully consistent with other languages. You can find a lot of differences between Go, Java and Node, for example (but we try to be aligned when possible)

Also, we should compare `exec_after_start` with strategies here, not the `WaitFor` itself.

However we can consider such strategy for sure, more like "alias" or "short/easier way".
But we need to avoid maintaining of duplicative logic/code.  Let me explain:

What are we introducing here that is missing in `exec_after_start`?
If it's about "easier way" - we need to find a way to compose it more correctly (reuse code?)
But if there is a difference in functionality - then it should be very clear to users why to use one over the other

testcontainers/src/core/wait/mod.rs
```diff
@@ -44,6 +48,8 @@ pub enum WaitFor {
     Http(HttpWaitStrategy),
     /// Wait for the container to exit.
     Exit(ExitWaitStrategy),
+    /// Wait for a certain command to exit with a successful code.
+    SuccessfulCommand(CommandStrategy),
```
DDtKey, line 52: I'd make it more general rather than specific.
We always can expose a constructor method like `successful_command`

```suggestion
    /// Wait for a certain command to exit with a successful code.
    Command(CommandStrategy),
```

testcontainers/src/core/wait/command_strategy.rs
DDtKey: Let's consider the current implementation to make it more clear if it's clear enough 

```rs
.with_wait_for(WaitFor::SuccessfulCommand(
        CommandStrategy::command(
            ExecCommand::new(["pg_isready"]).with_cmd_ready_condition(CmdWaitFor::exit_code(0)),
        ),
    ))
```

What confuses me a little bit here:

- `with_cmd_ready_condition(CmdWaitFor::exit_code(0)` - expects exit-code = 0
  -  `SuccessfulCommand` - also expects exit-code = 0
  - What if I'd change `with_cmd_ready_condition` to expect non-zero code or something else?
  - Do I really need to specify this two times as a user?
  
  
Perhaps, instead I'd suggest to support something like `fail_fast`/`retry` for `ExecCommand`
And add very simple strategy, like `Command(ExecCommand)`

So it will look like:
```rs
.with_wait_for(
  WaitFor::command(
    ExecCommand::new(["pg_isready"]).with_cmd_ready_condition(CmdWaitFor::exit_code(0))
  )
)
```

sutt0n: Apologies for the delay, been a busy week here with the weather we've been getting in Arkansas. 😅 

I've pushed an update that I believe is something that aligns... I was hesitant on the `CmdWaitFor::exit_code(0)` in `WaitFor::command()`, but added it anyway.

...however, now that I think about it, we may want to not force that specific exit code, but allow users to specify one on their own. Thoughts?

testcontainers/src/core/wait/mod.rs
sutt0n: Thank you for the constructive feedback! I've made these changes.

testcontainers/src/core/wait/command_strategy.rs
```diff
@@ -0,0 +1,110 @@
...
+    ContainerAsync, Image,
+};
+
+#[derive(Debug, Clone)]
+pub struct CommandStrategy {
+    expected_code: i64,
+    poll_interval: Duration,
+    command: ExecCommand,
+    fail_fast: bool,
+}
```
DDtKey, line 14: It seems unchanged - we still have `expected_code` here which may contradict to `ExecCommand`

The idea was to change how it's organized and implemented
https://github.com/testcontainers/testcontainers-rs/pull/782#discussion_r2023388158

sutt0n: Apologies for this overlook -- updated!

sutt0n:
Force pushed to fix the commit message... hard to write Rust while holding a sleeping baby. 😅

testcontainers/src/core/wait/command_strategy.rs
sutt0n: @DDtKey Is there anything else you'd like fixed on this PR?

sutt0n: @DDtKey Wanting to follow up here, is there anything else you'd like to see addressed on this PR that I may have missed?

sutt0n: @DDtKey Hey just wanting to touch base and ensure that this PR doesn't get too stale.

DDtKey:
Sorry again for the long delay.

@sutt0n, do you think [this solution](https://github.com/testcontainers/testcontainers-rs/pull/816) would address your use case as well?
I find it more natural and aligned with the Docker interface, while still offering nearly the same functionality — the ability to wait for an arbitrary command to succeed during startup.

Issue #433, opened 2022-11-08
Support pausing and resuming a container


lcmgh:
Hi!

In my integration test I like to test my apps behavior in case the database becomes unavailable. Currently I am stopping the db but of course I also want to test the recover of the database with its former state. Even if don't mind loosing the former db state I still have to create a new db pool because port changes on every container spawn and the already injected one can never recover.

Two options come to my mind to solve the issue:
1) Allow defining an explicit port
2) Allow to pause and continue container runs (my preferred option) 

Thanks

DDtKey:
Are you (probably) looking for adding docker's `pause`/`unpause` to testcontainers api?

thomaseizinger:
I have never heard of `docker pause` but it seems like a reasonable thing to implement so I am gonna accept a PR that adds a `pause` and `unpause` API to `Container`.

FibreFoX:
Stupid question: wouldn't it make more sense to just use [`stop`](https://docs.rs/testcontainers/latest/testcontainers/struct.Container.html#method.stop) and [`start`](https://docs.rs/testcontainers/latest/testcontainers/struct.Container.html#method.start) instead of introducing a new pause/unpause, to have it more "in line" with the Java implementation like documented on the website https://www.testcontainers.org/test_framework_integration/manual_lifecycle_control/?

I think it feels "more natural" when having both ecosystems (Rust + Java) using the same method language.

DDtKey:
`stop + start` recreates container (with graceful shutdown logic by `sigterm` signal or `sigkill` in case of hanging), so the state of such container will be dropped. 

While `pause`/`unpause` doesn't. 
One more case is introducing some controlled delays. 

Just a different behavior and could be useful for testing purposes.

Anyway, it's [official docker API](https://docs.docker.com/engine/reference/commandline/pause/) for suspending processes. So I think we should avoid this "just to align"

FibreFoX:
@DDtKey Thanks for clarification, I wasn't aware that `stop + start` actually recreate the containers instead just stopping and re-starting the container. And sure, this is a very special scenario.

If I understand correct, the [testcontainers-java](https://github.com/testcontainers/testcontainers-java) should have an open issue to feature-request this too, right?

mervyn-mccreight:
> One more case is introducing some controlled delays.

[...]

PR #785, opened 2025-04-10
feat: support `pause` and `unpause` container

### LINKED ISSUE #715: Missing feature for pausing containers
`ContainerAsync` `stop()` docs says:

```
Stops the container (not the same with `pause`).
```

However this `pause()` doesn't exist.

olitha:
This adds the pause/unpause calls for ContainerAsync/Container because I wanted to have an integration test keeping the same port for a mosquitto container.

fixes #715, #433


DDtKey (approved):

PR #787, opened 2025-05-04
feat: allow `security_opt` and `readonly_rootfs` to be configured

gurinderu:
I want to implement a Talos test container, and there is no way to change some arguments.

```
docker run --rm -it \
  --name tutorial \
  --hostname talos-cp \
  --read-only \
  --privileged \
  --security-opt seccomp=unconfined \
  --mount type=tmpfs,destination=/run \
  --mount type=tmpfs,destination=/system \
  --mount type=tmpfs,destination=/tmp \
  --mount type=volume,destination=/system/state \
  --mount type=volume,destination=/var \
  --mount type=volume,destination=/etc/cni \
  --mount type=volume,destination=/etc/kubernetes \
  --mount type=volume,destination=/usr/libexec/kubernetes \
  --mount type=volume,destination=/opt \
  -e PLATFORM=container \
  ghcr.io/siderolabs/talos:v1.9.5
```

### DIFF
--- a/testcontainers/src/core/image/image_ext.rs
+++ b/testcontainers/src/core/image/image_ext.rs
@@ -169,6 +169,12 @@ pub trait ImageExt<I: Image> {
 
     /// Sets the user that commands are run as inside the container.
     fn with_user(self, user: impl Into<String>) -> ContainerRequest<I>;
+
+    /// Sets the container's root filesystem to be mounted as read-only
+    fn with_readonly_rootfs(self, readonly_rootfs: Option<bool>) -> ContainerRequest<I>;
+
+    /// Sets security options for the container
+    fn with_security_opt(self, security_opt: Option<Vec<String>>) -> ContainerRequest<I>;
 }
 
 /// Implements the [`ImageExt`] trait for the every type that can be converted into a [`ContainerRequest`].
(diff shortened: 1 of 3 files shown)


testcontainers/src/core/image/image_ext.rs
```diff
@@ -169,6 +169,12 @@ pub trait ImageExt<I: Image> {
 
     /// Sets the user that commands are run as inside the container.
     fn with_user(self, user: impl Into<String>) -> ContainerRequest<I>;
+
+    /// Sets the container's root filesystem to be mounted as read-only
+    fn with_readonly_rootfs(self, readonly_rootfs: Option<bool>) -> ContainerRequest<I>;
+
+    /// Sets security options for the container
+    fn with_security_opt(self, security_opt: Option<Vec<String>>) -> ContainerRequest<I>;
```
DDtKey, line 177: I think we can drop `Option` from arguments similarly to other methods.

I'd say `None` is needed as a default option, but isn't really needed as part of API here

DDtKey, line 177: Also, it would be nice to keep aligned "append" semantic  for `security_opt` as we do for capabilities for example

1. Accept `opt: impl Into<String>`
2. Append to the list
```
        container_req
            .security_opts
            .get_or_insert_with(Vec::new)
            .push(capability.into());
```

DDtKey: This also may prevent us from having `Some( vec![] )` (empty vec inside the Some)

DDtKey (commented):
Thank you for the contribution! 🚀 

I've left some minor comments, please take a look

testcontainers/src/core/image/image_ext.rs
gurinderu: Fixed

gurinderu: Fixed

DDtKey (approved):

kiview:
@DDtKey In other languages (Java, Go, .Net and Node.js should be relatively consistent here) we have the general pattern, of not allowing *all properties* to be configured in the main public API, but instead relying on a mechanism to drop onto the Docker API abstraction level, hence exposing all available option here.

* https://java.testcontainers.org/features/advanced_options/#using-createcontainercmdmodifier
* https://golang.testcontainers.org/features/common_functional_options/#docker-type-modifiers
* `WithCreateParameterModifier` in .NET

Maybe worthwhile to consider for tc-rust as well.

PR #790, opened 2025-05-19
feat: add `Container*::is_running`

tversteeg:
Simple, but for my use-case, very useful check if a container is still live.


DDtKey (approved):
Thank you for the contribution! 🚀

Issue #718, opened 2024-08-07
Support for passing platform


poplexity:
Similar to https://github.com/testcontainers/testcontainers-rs/issues/500 it would be great to have feature parity with the CLI.  The suggested workaround does work but is quite inconvenient, we have migrated one project to use bollard directly such that we could specify the platform when launching the container.

shelajev:
After looking at the other implementations, this could perhaps be a more general request. We could expose a lower level API, similar to `withCreateCmdModifier` in testcontainers-java, and that could be used to configure various options: 
* platform
* CPU and memory limits for repeatable environments for performance
* runtimes, to enable GPU support like in the Ollama module

laz-001:
just honor `DOCKER_DEFAULT_PLATFORM`

DDtKey:
As [it was discussed](https://github.com/testcontainers/testcontainers-rs/pull/800#discussion_r2170483838) it's not enough to simply rely on `DOCKER_DEFAULT_PLATFORM` as `testcontainers` supports remote hosts (and [Testcontainers Cloud](https://testcontainers.com/cloud/)) that may differ from the local setup. 

So it's been addressed with hybrid approach: 
 - #800 auto detection of `DOCKER_DEFAULT_PLATFORM`
   -  but with fallback to default platform in #837 if image pull failed
 - Ability to configure platform explicitly in the code via `ImageExt` in #838

laz-001:
> As [it was discussed](https://github.com/testcontainers/testcontainers-rs/pull/800#discussion_r2170483838) it's not enough to simply rely on `DOCKER_DEFAULT_PLATFORM`

Potentially not enough in some scenarios, but still essential. Just to clarify, NOT honoring `DOCKER_DEFAULT_PLATFORM` is a bug in itself (i hit on this bug in a scenario i don't remember anymore...)

DDtKey:
What are you trying to say? Did anyone argued that it should not be honored? Issue was accepted initially
Ignoring other scenarios also would be a bug as it's common enough to use remote hosts and etc, even for DinD use-cases in CI

Message above was intended to provide the context of how it works now

laz-001:
> What are you trying to say?

Oh, this twisting. Cu around.

PR #800, opened 2025-06-26
feat: use DOCKER_DEFAULT_PLATFORM env var to specify platform

### LINKED ISSUE #500: is there any way to pass "platform" parameter when running docker images?
for example,

```bash

```

trevormcdonald:
When the env var is set the platform is passed for both image and container creation in parity with the docker CLI.

This is especially useful on apple silicon when developing for linux environments.

fixes #500, #718


DDtKey:

testcontainers/src/core/env/config.rs
DDtKey, line 1: The main issue I see is that testcontsiners can utilize remote docker (or testcontsiners cloud)

So relying on `DOCKER_DEFAULT_PLATFORM` might not be appropriate way, because your local setup might differ from the one used with testcontsiners.

In that case, there should be a way to override the variable only for testcontsiners at least

DDtKey (changes requested):

testcontainers/src/core/env/config.rs
DDtKey: Docker host can be currently overridden by properties file, so  I guess it can be solved by special config there. 
However, the format of properties must be aligned with implementations in other languages 🤔

Let me think a little bit, perhaps we should accept the PR and allow overriding as follow-up

Because another important fact is that testcontainers respect `DOCKER_HOST`

trevormcdonald: I appreciate the effort! I looked around a bit at the state of other language implementations and noticed the java version will fall back to a specific platform if an initial pull fails: https://github.com/testcontainers/testcontainers-java/blob/b2b404a9a87ed70d0672fe5bb9ac1e3a463af94f/core/src/main/java/org/testcontainers/images/RemoteDockerImage.java#L165

That may work for at least my specific case, will test it soon if that is a desired alternative.

mervyn-mccreight (approved):

testcontainers/src/core/env/config.rs
DDtKey: @trevormcdonald Just following up to see if you’ve had any success with your suggestion above?

trevormcdonald: sorry for the radio silence! I've implemented the fallback platform approach over here: https://github.com/testcontainers/testcontainers-rs/pull/837

PR #802, opened 2025-07-02
chore: remove unused Python files

amab8901:
Pipfile is for Python. This project has no Python. `cargo t` passes locally after removing pipfile. Same with `requirements.txt` and `runtime.txt`. Doesn't seem sensible to have these files since the project is based on Rust rather than python, and these files don't seem to be used in any way. So the only thing these files do is create confusion


DDtKey:
Just to clarify, we do use Python for building our documentation website. As you can see, the CI actions are currently failing.

PR #803, opened 2025-07-02
test: add integration test for connection between client and server

amab8901:
add integration test for connection between client container and server container. This is useful to have as an example implementation to help users figure out how to write their own integration test that relies on connection between containers

it passes `cargo t` and `cargo t --features blocking`


DDtKey:
Hi 👋 

We have switched test-suite to utilize new builder API. If you're interested in getting this merged, could you update the PR?

Thank you!

amab8901:
update the PR in what way?

PR #804, opened 2025-07-02
refactor: testimages and tests to use builder API

mbodmer:
This PR is based on feat/build-image from PR: https://github.com/testcontainers/testcontainers-rs/pull/801


DDtKey:
We can make a chain of PRs, this can point to branch of https://github.com/testcontainers/testcontainers-rs/pull/801. So it will show clear diff

mbodmer:
> We can make a chain of PRs, this can point to branch of #801. So it will show clear diff

how would i do that? Merge this PR into mbodmer:/feat/build-image from #801 ? This is not possible, as i cannot choose this branch as the target branch of the PR.

mbodmer:
I think I need help here:
https://github.com/testcontainers/testcontainers-rs/actions/runs/16045667198/job/45298950753?pr=804#step:7:369

locally this works, but on the buildserver there seems to be a different environment

DDtKey:
Right, that's your fork's branch
Never mind, it will be clearer once we merge builder

mbodmer:
Could you please start the build again? Thanks

mbodmer:
rebased to the current state of #801

mbodmer:
Again, rebased to the current state of https://github.com/testcontainers/testcontainers-rs/pull/801

mervyn-mccreight:
@mbodmer the builder api pr has been merged, let's tackle this one next :)

mbodmer:
A good feature for future improvement would be having an ignore list or better recognize .dockerignore/.containerignore files, when adding a directory to the build context.

testcontainers/src/runners/sync_runner.rs
```diff
@@ -118,6 +119,20 @@ mod tests {
         runtime().block_on(client.network_exists(name)).unwrap()
     }
 
+    fn get_server_container(msg: Option<WaitFor>) -> GenericImage {
```
mervyn-mccreight, line 122: This is always called with `None`. I think since this returns `GenericImage` adding the `WaitFor` can be done afterwards anyways.

mervyn-mccreight (approved):

testcontainers/src/runners/sync_runner.rs
mbodmer: I've just followed the same pattern already existing. But you are right, the parameter is never used and I've thus removed it now

Issue #808, opened 2025-07-04
Methods `.stdout()` and `.stderr()` of `ExecResult` are not `Send`


amab8901:
# Description

`.stdout()` and `.stderr()` methods on `ExecResult` return futures that are unsafe to send between threads

# Code to reproduce:

<details>
<summary>main.rs</summary>

``` rust
const fn main() {}

#[cfg(test)]
mod tests {
    use std::sync::Arc;

    use testcontainers::{ContainerAsync, GenericImage, core::ExecCommand, runners::AsyncRunner};

    #[tokio::test]
    pub(crate) async fn test_lol() {
        let container = GenericImage::new("influxdb", "2.7.12")
            .start()
            .await
            .unwrap();

        let arc_container = Arc::new(container);

        // tokio::spawn(trigger_warning(arc_container));
        trigger_warning(arc_container).await;
    }

    pub(crate) async fn trigger_warning(container: Arc<ContainerAsync<GenericImage>>) {
        let mut exec_result = container.exec(ExecCommand::new(vec!["ls"])).await.unwrap();
        let x = exec_result.stdout_to_vec().await.unwrap();
        println!("{x:?}");
    }
}

```

</details>

<details>
<summary>Cargo.toml</summary>

``` toml

[package]
name = "nice"
version = "0.1.0"
edition = "2024"

[dependencies]
bollard = { version = "0.18.1" }
bytes = "1.10.1"
testcontainers = "0.24.0"
tokio = "1.45.1"
tracing = "0.1.41"
tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }

[workspace]
resolver = "3"
members = ["client", "server"]


[lints.clippy]
let_and_return = "allow"
similar_names = "allow"
too_many_arguments = "allow"
missing_errors_doc = "allow"
pub_with_shorthand = "allow"
missing_docs_in_private_items = "allow"
field_scoped_visibility_modifiers = "allow"
arbitrary_source_item_ordering = "allow"
self_named_module_files = "allow"
single_char_lifetime_names = "allow"
single_call_fn = "allow"
allow_attributes_without_reason = "allow"
unused_trait_names = "allow"
blanket_clippy_restriction_lints = "allow"
allow_attributes = "allow"
decimal_literal_representation = "allow"
implicit_return = "allow"
question_mark_used = "allow"
multiple_inherent_impl = "allow"
needless_pass_by_ref_mut = "warn"
shadow_reuse = "warn"
[...]

PR #809, opened 2025-07-04
fix: add Send trait to stderr & stdout

amab8901:
Solves:
https://github.com/testcontainers/testcontainers-rs/issues/808


DDtKey:

PR #815, opened 2025-07-15
feat: allow TESTCONTAINERS_HOST_OVERRIDE to override docker host

Hugoch:
This PR adds the option to customize Docker hostname detection with the `TESTCONTAINERS_HOST_OVERRIDE` environment variable. 
It replicates the behavior of the Java testcontainer package (https://java.testcontainers.org/features/configuration/)

### DIFF
--- a/testcontainers/src/core/client.rs
+++ b/testcontainers/src/core/client.rs
@@ -421,6 +421,13 @@ impl Client {
     }
 
     pub(crate) async fn docker_hostname(&self) -> Result<url::Host, ClientError> {
+        // allow TESTCONTAINERS_HOST_OVERRIDE to override the docker host
+        if let Some(host_override) =
+            <env::Os as env::GetEnvValue>::get_env_value("TESTCONTAINERS_HOST_OVERRIDE")
+        {
+            return url::Host::parse(&host_override)
+                .map_err(|_| ConfigurationError::InvalidDockerHost(host_override).into());
+        }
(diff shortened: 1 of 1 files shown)


testcontainers/src/core/client.rs
```diff
@@ -421,6 +421,13 @@ impl Client {
     }
 
     pub(crate) async fn docker_hostname(&self) -> Result<url::Host, ClientError> {
+        // allow TESTCONTAINERS_HOST_OVERRIDE to override the docker host
+        if let Some(host_override) =
+            <env::Os as env::GetEnvValue>::get_env_value("TESTCONTAINERS_HOST_OVERRIDE")
+        {
+            return url::Host::parse(&host_override)
+                .map_err(|_| ConfigurationError::InvalidDockerHost(host_override).into());
+        }
```
DDtKey, line 430: There is more appropriate place for this - https://github.com/testcontainers/testcontainers-rs/blob/d652b383ae75c0b1e56362751e4fb8a7b80d8bb2/testcontainers/src/core/env/config.rs#L127-L137

Moreover config already has `tc_host` which is prioritized, but we read it only from properties file. 
What we can do is to extend [this part](https://github.com/testcontainers/testcontainers-rs/blob/d652b383ae75c0b1e56362751e4fb8a7b80d8bb2/testcontainers/src/core/env/config.rs#L103-L119) and read tc_host from env variables too

DDtKey (commented):

PR #818, opened 2025-07-16
ci: distribute feature-set test execution

mervyn-mccreight:
Use cargo-hacks `--partition` option to distribute feature-set test execution to a set of runners with the goal to speed up feedback cycles.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
   merge_group:
   push:
-    branches: [ main ]
+    branches: [main]
 
 concurrency:
   group: ${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -43,6 +43,12 @@ jobs:
         toolchain:
           - stable
           - nightly
+        partition:
+          - 1/5
+          - 2/5
+          - 3/5
+          - 4/5
+          - 5/5
```
DDtKey, line 51: I think 4 would be enough, WDYT?

mervyn-mccreight: Sure, 5 was just an arbitrary number to be honest

shdhgfejbcd:
Dhgrhvfth

shdhgfejbcd:
> Dhgrhvfth

Fhgggjjfhhgf

PR #820, opened 2025-07-23
doc: rustdoc for builder API

mbodmer:
For a start I have now documented the relevant parts of the builder API here.

I have added no_run to examples as the needed resources are missing, they should be built however to ensure correctness.


mervyn-mccreight (approved):
Thanks for taking the time to add the documentation! 🚀

DDtKey (approved):
Thank you for wrapping up this with documentation! ❤️

Issue #821, opened 2025-07-25
[feature request] Support for Host Port Exposure (host.testcontainers.internal)


DCjanus:
The Testcontainers for Java library offers a powerful feature that allows containers to connect back to services running on the host machine. This is achieved by exposing host ports, which are then accessible within the container via the special hostname `host.testcontainers.internal`.

You can see the documentation for the Java feature here: [Exposing host ports to the container](https://java.testcontainers.org/features/networking/#exposing-host-ports-to-the-container)

This functionality is incredibly useful for integration tests where a containerized application needs to communicate with a mock server, a local database instance, or any other service running on the test host.

It would be a great addition to `testcontainers-rs` to implement similar functionality.

### How it works in testcontainers-java

Testcontainers for Java uses a sophisticated, multi-layered approach to make this work seamlessly across different environments (Docker Desktop, Linux, Docker-in-Docker):

1.  **Native Support (Higher Priority):** It leverages native Docker features like the special DNS name `host.docker.internal` (on Docker Desktop) or the `host-gateway` keyword (on modern Linux Docker) by adding an entry to the container's `/etc/hosts` file (e.g., `--add-host=host.testcontainers.internal:host-gateway`).

2.  **Fallback Mechanism (High Compatibility):** When native support is unavailable, it automatically starts a lightweight SSHD container (`testcontainers/sshd`). It then uses SSH remote port forwarding to tunnel traffic from the application container to the host machine. The `host.testcontainers.internal` hostname is resolved to the SSHD container's IP address within the shared Docker network.

3.  **Dynamic Gateway Discovery:** In complex scenarios like Docker-in-Docker, it can even discover the correct host gateway IP by launching a temporary container and inspecting its network route (`ip route`).

### Proposed Feature for `testcontainers-rs`

We propose adding a similar feature to `testcontainers-rs` that would allow users to:

1.  Declare host ports to be exposed to containers.
2.  Access these ports from within the container using a consistent hostname, like `host.testcontainers.internal`.

This would significantly enhance the library's capabilities for a wide range of integration testing scenarios. 

If this feature request is accepted, I would be interested in contributing to its development when I have the time.

Thank you for considering this feature request.

DDtKey:
Thank you for filling the feature-request!

This is a useful feature and contributions are welcome!

Some references to extend the context of how Java version handles this:

[...]

PR #830, opened 2025-08-09
feat: add host port mapping support for container-to-host communication

mcuste:
## Summary

  - Relevant feature request: #821
  - Add `with_exposed_host_port()` API to enable containers to connect back to host services
  - Implement automatic host mapping detection using native Docker features:
    - `host-gateway` on Linux (Docker 20.10+)
    - `host.docker.internal` on Docker Desktop (macOS/Windows)
  - Add comprehensive tests covering various scenarios including parallel connections
    - Unit tests for version detection and mapping logic
    - Integration tests for host service connectivity
    - Parallel connection tests to ensure thread safety
    - All existing tests continue to pass
  - Include example demonstrating the feature usage

  ## Background

  This feature, requested on #821, enables containers to communicate with services running on the host machine by exposing host ports through a standardized hostname `host.testcontainers.internal`. This is particularly useful for integration tests where   containers need to call back to test services or mock servers running on the host.

  The implementation leverages native Docker features that have been stable for years:
  - `host-gateway` support was added in Docker 20.10 (released October 2020, ~5 years ago)
  - `host.docker.internal` has been available on Docker Desktop since early versions

  Any Docker installation older than 20.10 would be significantly outdated, so the current implementation focuses on
  these native approaches rather than complex fallback mechanisms.

  As for the DinD case, I will look into it in the future if I can find the time.

### DIFF
--- a/testcontainers/src/core/host_port_mapping.rs
+++ b/testcontainers/src/core/host_port_mapping.rs
@@ -1,0 +1,169 @@
+use std::collections::BTreeMap;
+
+use crate::core::{client::Client, containers::request::Host, error::TestcontainersError};
+
+/// Checks if we're running on Docker Desktop (macOS or Windows).
+/// Docker Desktop provides `host.docker.internal` by default.
+fn is_docker_desktop() -> bool {
+    let is_desktop = cfg!(any(target_os = "macos", target_os = "windows"));
+    if is_desktop {
+        log::trace!("Platform: Docker Desktop, host.docker.internal available");
+    } else {
+        log::trace!("Platform: not Docker Desktop");
+    }
+    is_desktop
+}
+
+/// Checks if a Docker version string is at least the specified major.minor version
+fn is_docker_version_at_least(version_str: &str, min_major: u32, min_minor: u32) -> bool {
+    // Docker version strings can be like "20.10.17", "24.0.0", etc.
+    let parts: Vec<&str> = version_str.split('.').collect();
+    if parts.len() < 2 {
+        return false;
+    }
+
+    let major: u32 = parts[0].parse().unwrap_or(0);
+    let minor: u32 = parts[1].parse().unwrap_or(0);
+
+    major > min_major || (major == min_major && minor >= min_minor)
+}
+
+/// Checks if the current Docker installation supports the `host-gateway` feature.
+/// This is available in Docker 20.10+ on Linux.
+async fn supports_host_gateway(client: &Client) -> bool {
+    if !cfg!(target_os = "linux") {
+        log::trace!("Platform: not Linux, host-gateway unsupported");
+        return false;
+    }
+
+    log::trace!("Platform: Linux, checking Docker version for host-gateway");
+
+    // Check Docker version to see if it supports host-gateway
+    match client.docker_version().await {
+        Ok(Some(version_str)) => {
+            let supported = is_docker_version_at_least(&version_str, 20, 10);
+            if !supported {
+                log::warn!("Docker version {} detected on Linux - host-gateway may not be supported (requires 20.10+)", version_str);
+            }
+            supported
+        }
+        Ok(None) => {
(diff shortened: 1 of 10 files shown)


mcuste:
Currently one of the tests fail on CI: `cargo test --no-default-features --features http_wait,properties-config`

Tried it locally but can't really reproduce, and the changes I made shouldn't effect those tests. The error is `IOError { err: Custom { kind: Other, error: "bytes remaining on stream" } }`, so probably it failed due to the Docker in CI environment, which will probably pass on retry.

DDtKey:
Yes, it seems to be flakiness. 

Thanks for the contribution! 🚀 

I will try to take a look soon!
cc @mervyn-mccreight

testcontainers/src/core/host_port_mapping.rs
```diff
@@ -0,0 +1,169 @@
...
+        }
+        Err(err) => {
+            // If we can't get version info, assume it's supported but warn
+            log::warn!(
+                "Failed to detect Docker version ({}), assuming host-gateway is supported on Linux",
+                err
+            );
+            true
+        }
+    }
```
DDtKey, line 65: This looks like good candidate to be a method of client as main argument is `&Client`

But in order to split the logic and decouple from the `Client` I'd suggest to define some trait and impl it for the `Client` (e.g `impl HostMapping for Client { ...}`)

testcontainers/src/core/host_port_mapping.rs
```diff
@@ -0,0 +1,169 @@
...
+/// Checks if a Docker version string is at least the specified major.minor version
+fn is_docker_version_at_least(version_str: &str, min_major: u32, min_minor: u32) -> bool {
+    // Docker version strings can be like "20.10.17", "24.0.0", etc.
+    let parts: Vec<&str> = version_str.split('.').collect();
+    if parts.len() < 2 {
+        return false;
+    }
+
+    let major: u32 = parts[0].parse().unwrap_or(0);
+    let minor: u32 = parts[1].parse().unwrap_or(0);
```
DDtKey, line 26: Let's rely on https://crates.io/crates/semver for that?

testcontainers/src/core/host_port_mapping.rs
```diff
@@ -0,0 +1,169 @@
+use std::collections::BTreeMap;
+
+use crate::core::{client::Client, containers::request::Host, error::TestcontainersError};
+
+/// Checks if we're running on Docker Desktop (macOS or Windows).
+/// Docker Desktop provides `host.docker.internal` by default.
+fn is_docker_desktop() -> bool {
+    let is_desktop = cfg!(any(target_os = "macos", target_os = "windows"));
```
DDtKey, line 8: This won’t actually work as is because we can’t rely on the operating system of the user for that.

Testcontainers support remote Docker hosts, so the OS you’re running tests on may be different from the remote's one.

DDtKey (changes requested):
I already started reviewing piece by piece and then realized conceptual issue: https://github.com/testcontainers/testcontainers-rs/pull/830#discussion_r2265479792 - the implementation relies only on `target_os` which isn't enough (It won't work for remote docker, testcontainers cloud and etc)

Tbh I think that's the blocker for the whole implementation we need to address 🤔 

See details in #821 and references in comments - that's generally why `sshd` is used in other languages as a fallback mechanism 

❓We could consider adding initial support for non-remote hosts only (with explicit documentation) to split this into several steps, but we have to ensure it can be extended without breaking things

DDtKey:
So, I'm generally ok to split work into 2-3 steps:
- introduce initial support limited to local docker installations (+ have flexible enough interface)
- support fallback mechanisms, remote docker hosts
- ensure DinD works

But let's discuss what can be achieved in the first iteration

DDtKey:
Closing in favor of #846

Thanks for all the efforts anyway! Appreciate it!

PR #832, opened 2025-08-11
feat: add support for devices request

### LINKED ISSUE #831: Expose host's underlying GPUs
I am using the library and would like to write tests that use GPU. I looked around and didn't see code, nor issue, for that. Is there a way to use GPU's in the current state of implementation?

Context:
https://docs.docker.com/engine/containers/resource_constraints/#expose-gpus-for-use

diegopso:
This should expose the Bollard API to make device requests and enable usage of host's GPUs for tests.

I didn't find a good way to integrate with CI since the test I wrote depends on an actual GPU. I wrapped the feature on a feature flag and then skipped testing that feature flag in CI.

Closes #831

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
         with:
           tool: cargo-hack
       - name: Tests
-        run: cargo hack test --feature-powerset --depth 2 --clean-per-run --partition ${{ matrix.partition }}
+        run: cargo hack test --feature-powerset --exclude-features device-requests --depth 2 --clean-per-run --partition ${{ matrix.partition }}
 
   fmt:
     name: Rustfmt check
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
         with:
           tool: cargo-hack
       - name: Tests
-        run: cargo hack test --feature-powerset --exclude-features device-requests --depth 2 --clean-per-run --partition ${{ matrix.partition }}
+        run: cargo hack test --feature-powerset --depth 2 --clean-per-run --partition ${{ matrix.partition }}
 
   fmt:
     name: Rustfmt check
--- a/testcontainers/Cargo.toml
+++ b/testcontainers/Cargo.toml
@@ -60,6 +60,7 @@ watchdog = ["signal-hook", "conquer-once"]
 http_wait = ["reqwest"]
 properties-config = ["serde-java-properties"]
 reusable-containers = []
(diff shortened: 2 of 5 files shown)


diegopso:
I fixed the PR name. 

The test errors seem to me as if there is a problem with concurrent tests that either use the same network name or port number. Sometimes just rerunning the jobs fixes them, like [here](https://github.com/diegopso/testcontainers-rs/actions/runs/16878692489/job/47809908976). Once running locally, I had to manually delete one of the networks that was left over after the tests.

One possible solution would be to add some randomness to these values. I think it is unrelated to this PR though.

diegopso:
Now it looks like a log consumer was hanging forever until the Job was evicted. Also seems unrelated to this PR.

mervyn-mccreight:
Thanks for the contribution! 🚀 

Hmm, I found that it would be possible with access to large runners: https://docs.github.com/en/actions/concepts/runners/larger-runners

@DDtKey do you know if we have access to them?

> The test errors seem to me as if there is a problem with concurrent tests that either use the same network name or port number. 

Yes, the teams have been unstable lately, we need to put some effort into making them stable again. Sorry for the inconvenience.

DDtKey:
I don’t think larger runners are the solution - they only would hide the issue.

I believe many `testcontainers` users may have even more tests and containers in their flows.

Considering the errors in the CI - we need to think about auto-retries on the library level for transient errors in order to reduce flakiness.

.github/workflows/ci.yml
```diff
@@ -67,7 +67,7 @@ jobs:
         with:
           tool: cargo-hack
       - name: Tests
-        run: cargo hack test --feature-powerset --depth 2 --clean-per-run --partition ${{ matrix.partition }}
+        run: cargo hack test --feature-powerset --exclude-features device-requests --depth 2 --clean-per-run --partition ${{ matrix.partition }}
```
DDtKey, line 70: Do we really need exclusion? It would affect other developers who runs tests locally / in cloud

Instead, we could consider not failing such tests, but run only when it's possible. And would be better to print a warning if we can't run the test

mervyn-mccreight:
With large runners I meant to solve the problem that this needs a GPU to be testable. AFAIK in the set of large runners there also is a GPU runner.

The instabilities are a separate issue - one we should solve without the need for larger runners as you just said, so I'm with you there :)

DDtKey:
@mervyn-mccreight  I see, sorry for misunderstanding. I'm not sure - I'll need to check. But I think no reason to block the PR.

@diegopso Thanks for the contribution! 🚀

PR #837, opened 2025-08-26
feat: fallback to platform linux/amd64 when image pull fails

### LINKED ISSUE #500: is there any way to pass "platform" parameter when running docker images?

trevormcdonald:
this introduces similar behavior to testcontainers-java, seen here: https://github.com/testcontainers/testcontainers-java/blob/20ca76649c14cb03841b3c373dbc26be0fb83467/core/src/main/java/org/testcontainers/images/RemoteDockerImage.java#L165

fixes #500, #718

### DIFF
--- a/testcontainers/src/core/client.rs
+++ b/testcontainers/src/core/client.rs
@@ -454,10 +454,46 @@ impl Client {
             .bollard
             .create_image(Some(pull_options), None, credentials);
         while let Some(result) = pulling.next().await {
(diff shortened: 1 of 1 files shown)


mervyn-mccreight:
@DDtKey WDYT behavior wise?

DDtKey (approved):
It corresponds to Java's version and generally make sense as I initially mentioned in #800 - so LGTM

Thanks for the contribution! 🙏

PR #838, opened 2025-09-14
feat: platform config passing through ImageExt trait

### LINKED ISSUE #500: is there any way to pass "platform" parameter when running docker images?

mominul:
Alternative approach to #800 #837

I think this approach gives more flexibility API-wise. I am happy to add tests if the maintainers think this is a more suitable approach.

Fixes #500, #718

Thanks in advance!


DDtKey:
Hi @mominul 👋 

Thanks for the contribution!

I think we may want to have both, but platform detection relying on default Docker's methods (`DOCKER_DEFAULT_PLATFORM`) looks like useful and needed feature.
As it's typical use-case for users and run the same code on different platforms (e.g developer with Apple Silicon, CI on linux + amd64 and etc)

Exposing a way to configure platform from the code looks reasonable as well, but it's more like addition, not a replacement of #800 & #837

What do you think about co-existence of both approaches? (with the following priory: explicitly configured in the code  > auto-detection > fallback)

mominul:
@DDtKey That would be great! I have rebased my changes upon #800 & #837 Can you have a look if you are okay with it? The CI failures are unrelated to these changes I think.

mominul:
Friendly ping ✋ 

Is anything remaining from my end?

PR #846, opened 2025-09-27
feat: expose host ports between host and containers via ssh sidecar

DCjanus:
Close https://github.com/testcontainers/testcontainers-rs/issues/821

### DIFF
--- a/testcontainers/Cargo.toml
+++ b/testcontainers/Cargo.toml
@@ -28,6 +28,7 @@ log = "0.4"
 memchr = "2.7.2"
 parse-display = "0.9.0"
 pin-project-lite = "0.2.14"
+russh = { version = "0.54.3", default-features = false, features = ["ring", "rsa"] }
 reqwest = { version = "0.12.5", features = [
     "rustls-tls",
     "rustls-tls-native-roots",
--- a/testcontainers/src/core/containers/async_container.rs
+++ b/testcontainers/src/core/containers/async_container.rs
@@ -3,6 +3,7 @@ use std::{fmt, net::IpAddr, pin::Pin, str::FromStr, sync::Arc, time::Duration};
 use tokio::io::{AsyncBufRead, AsyncReadExt};
 use tokio_stream::StreamExt;
 
+use super::host::HostPortExposure;
 use crate::{
     core::{
         async_drop,
@@ -42,6 +43,7 @@ pub struct ContainerAsync<I: Image> {
     #[allow(dead_code)]
     network: Option<Arc<Network>>,
     dropped: bool,
+    host_port_exposure: Option<HostPortExposure>,
     #[cfg(feature = "reusable-containers")]
     reuse: crate::ReuseDirective,
 }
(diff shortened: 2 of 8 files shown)


testcontainers/Cargo.toml
```diff
@@ -28,6 +28,7 @@ log = "0.4"
 memchr = "2.7.2"
 parse-display = "0.9.0"
 pin-project-lite = "0.2.14"
+russh = { version = "0.54.3", default-features = false, features = ["ring", "rsa"] }
```
DDtKey, line 31: I think we need to keep it under a feature - looks like a good candidate.
Isolated functionality and we can enable feature by default later.

It's common request to be able to use testcontainers without `ring` or `aws-lc-rs`.

DCjanus: I tried modifying it to use a feature gate, but it turned out to be much more complex than I expected. I might need some time to experiment on another branch.

Additionally, I just realized that I forgot to add an HTML in `docs/` introduction for this new feature. I will add it later. Thank you for your attention.

DCjanus:
Hi @DDtKey, thanks again for the earlier review—your feedback made it much easier to tighten things up. Since that round I’ve:

  - Wrapped the host port exposure implementation behind its feature gate and turned russh into an optional dependency. Because russh insists on exactly one TLS backend feature (ring or aws-lc-rs) being active, the
  new feature now effectively requires enabling precisely one of those; otherwise the crate will refuse to compile.
  - Guarded every runtime touchpoint (async container lifecycle, request wiring, runner, and the integration tests) with #[cfg(feature = "host-port-exposure")], so default builds stay free of the SSH tunnel code
  paths.
  - Updated CI to run both the “feature disabled” matrix and the “feature enabled with ring/aws-lc-rs” permutations. That coverage makes sure we catch the compile-time constraint above whenever someone mixes
  feature flags.
  - Added a networking guide that explains how to enable host port exposure, calls out the TLS backend requirement, and gives code samples, and then routed the page through MkDocs so users can actually find it.

DCjanus:
It seems the failing test is unrelated to this patch and may just be flaky. The same CI job passes on my fork, so I don’t have a repro tied to these changes. 

Since I don’t have permissions to retry the job here, I may have to wait for someone who does to rerun it.

DDtKey (approved):
Awesome, thanks a lot for all your efforts! 🚀

DDtKey:
JFYI, TIL that weak references in Cargo.toml result in the dependency being added to Cargo.lock. This is an issue in certain cases and is not really expected.

Reference: https://github.com/rust-lang/cargo/issues/10801

DCjanus:
> JFYI, TIL that weak references in Cargo.toml result in the dependency being added to Cargo.lock. This is an issue in certain cases and is not really expected.
> 
> Reference: [rust-lang/cargo#10801](https://github.com/rust-lang/cargo/issues/10801)

During local testing, I encountered this phenomenon, but I thought it was expected

Issue #843, opened 2025-09-22
`rustls/ring` is always active even with `aws-lc-rs` feature


dsykes16:
The `buildkit` feature of `bollard` includes the `ssl` feature, which includes `rustls/ring`. This results in `ring` being brought into dependencies even if the `aws-lc-rs` feature is selected. A workaround is to call `CryptoProvider::install_default()`, but that is less than ideal since `ring` shouldn't be brought into the dependencies at all if `aws-lc-rs` is chosen.

```
bollard #docker
An asynchronous Docker daemon API
version: 0.19.1 (latest 0.19.2)
license: Apache-2.0
rust-version: unknown
documentation: https://docs.rs/bollard
homepage: https://github.com/fussybeaver/bollard
repository: https://github.com/fussybeaver/bollard
crates.io: https://crates.io/crates/bollard/0.19.1
features:
  default                = [http, pipe]
...
  aws-lc-rs              = [ssl_providerless, rustls/aws-lc-rs]
  bollard-buildkit-proto = [dep:bollard-buildkit-proto]
  buildkit               = [chrono, num, rand, tokio/fs, tokio-stream, tokio-util/io, tonic, tower-service, ssl, bollard-stubs/buildkit, bollard-buildkit-proto, dep:async-stream, dep:bitflags]
...
  ssh                    = [hyper-util, openssh, tower-service]
  ssl                    = [ssl_providerless, rustls/ring]
...
```

DDtKey:
Hi @dsykes16 👋 

Thanks for the report!

Looks like this regression introduced in #801 (cc @mbodmer) as previously we have added ability to disable `default-features` and tls at all https://github.com/testcontainers/testcontainers-rs/issues/799#issuecomment-3032590885

That's unfortunate, we should have released builder api under a feature and make `buildkit` enabled only if builder feature enabled..

Since it's been already released, I think we need to make a `builder-api` feature enabled by default, so if default-features are disabled - `ring` won't be required. `ring` enabled anyway with `default-features`, so this sounds reasonable to me - disable default ones and enable `aws-lc-rs` is preferred - the main issue is that there won't be a way to enable `aws-lc-rs` + builder api without `ring` and that's probably a question to `bollard`)

However it's not always guaranteed

> but that is less than ideal since ring shouldn't be brought into the dependencies at all if aws-lc-rs is chosen.

[...]

PR #847, opened 2025-09-28
fix: update bollard to version supporting buildkit_providerless

mbodmer:
fixes: https://github.com/testcontainers/testcontainers-rs/issues/843

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,8 +31,10 @@ jobs:
       - uses: taiki-e/install-action@v2
         with:
           tool: cargo-hack
-      - name: Build
-        run: cargo hack build --feature-powerset --depth 2 --keep-going
+      - name: Build (without host-port-exposure)
+        run: cargo hack build --feature-powerset --depth 2 --keep-going --exclude-features host-port-exposure
+      - name: Build (host-port-exposure with TLS backend)
+        run: cargo hack build --feature-powerset --depth 2 --keep-going --include-features host-port-exposure,ring,aws-lc-rs --at-least-one-of ring,aws-lc-rs
 
   test:
     name: Test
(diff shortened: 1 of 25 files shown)


DDtKey:
@mbodmer may I ask you to check if the test failures are relevant? Looks like all of them are failing and it doesn't look like flakiness at first glance

mbodmer:
I am using my branch in productive tests. There is no problem with it. It looks like the buildserver takes too long to start the containers. Probably too much load on it.

DDtKey:
@mbodmer something definitely changed with this PR because all tests are failing, and retries haven’t helped. The errors are also different from the ones we know might be flaky.

```
  ---- runners::sync_runner::tests::sync_should_return_error_when_non_bridged_network_selected stdout ----
  
  thread 'runners::sync_runner::tests::sync_should_return_error_when_non_bridged_network_selected' panicked at testcontainers/src/runners/sync_runner.rs:130:14:
  called `Result::unwrap()` on an `Err` value: Client(BuildImage { descriptor: "simple_web_server:latest", err: DockerStreamError { error: "debian:bookworm-slim: failed to resolve source metadata for docker.io/library/debian:bookworm-slim: no active session for 01K88X0FRZ8933A0DZATYJGY4P: context deadline exceeded" } })
```

Many of users will utilize `testcontainers` in default CI setup, and it doesn't look good 
I'd agree if we would see this in current main, but we see this only against `providerless` version 🤔

DDtKey:
@mbodmer it wasn't working for me locally too
It worked only if image already present, but building was failing 

The issue is wrong execution branch selected,  there was one missing place where `buildkit` feature was required instead of `providerless` one
I've prepared a fix in bollard: https://github.com/fussybeaver/bollard/pull/597

DDtKey:
It works here with the patch: https://github.com/testcontainers/testcontainers-rs/pull/856

mbodmer:
Thank you for looking into it, i had a short look, but could not make sense of the behaviour. Would have needed more time. Thank you for looking into it!

PR #848, opened 2025-10-02
feat: add container hostname support

kevinalh:
The Docker container create command supports the --hostname parameter ([1]). This is also supported through the Bollard library as part of the the ContainerCreateBody struct ([2]). This commit adds the option to configure this parameter through the ImageExt trait.

This was previously proposed for an older version of testcontainers-rs ([3]), however it was rejected.

The usefulness of this feature is to allow programs inside the container to report back a hostname that makes sense in a test environment. For instance, ClickHouse has the hostName function ([4]) which will report this configured value. By default Docker assigns the Docker container ID, which in an integration testing context isn't that useful.

[1]: https://docs.docker.com/reference/cli/docker/container/create/
[2]: https://docs.rs/bollard/latest/bollard/secret/struct.ContainerCreateBody.html
[3]: https://github.com/testcontainers/testcontainers-rs/pull/539
[4]: https://clickhouse.com/docs/sql-reference/functions/other-functions#hostname

### DIFF
--- a/testcontainers/src/core/containers/request.rs
+++ b/testcontainers/src/core/containers/request.rs
@@ -27,6 +27,7 @@ pub struct ContainerRequest<I: Image> {
     pub(crate) image_tag: Option<String>,
     pub(crate) container_name: Option<String>,
     pub(crate) network: Option<String>,
+    pub(crate) hostname: Option<String>,
     pub(crate) labels: BTreeMap<String, String>,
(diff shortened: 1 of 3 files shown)


arturocuya (changes requested):
mucho chatgpt

DDtKey (approved):
Looks good to me and I definitely can imagine plenty of use cases for this

Thanks for the contribution! 🙏

PR #856, opened 2025-10-26
feat: support build options - `no_cache`, `skip_if_exists` and `buildargs`

DDtKey:


mbodmer:
So there were problems when multiple builds of the same image were executed at the same time? Looks good, thanks.

DDtKey:
It wasn't main problem, but something I realized during testing (that we can provide a way to avoid parallel builds for same images, quite often in CI/tests it's enough to build 1 time - but it's opt-in option)

Main issue was the wrong code branch and credentials handling in bollard due to wrong feature check

PR #879, opened 2025-11-23
perf: update ferroid for better performance during encode/decode

s0l0ist:
This PR updates ferroid to the latest release which increases performance of encode/decode by another 10-15%. It also introduces a more ergonomic `ULID::now()` method.


DDtKey:
Thanks!

PR #890, opened 2025-12-18
feat: add CPU and memory limits

bitemyapp:
Why is everything i64

### DIFF
--- a/testcontainers/src/core/containers/request.rs
+++ b/testcontainers/src/core/containers/request.rs
@@ -49,6 +49,64 @@ pub struct ContainerRequest<I: Image> {
     pub(crate) startup_timeout: Option<Duration>,
     pub(crate) working_dir: Option<String>,
     pub(crate) log_consumers: Vec<Box<dyn LogConsumer + 'static>>,
+
+    /// The length of a CPU period in microseconds. Default is 100000, this configures how
+    /// CFS will schedule the threads for this container. Normally you don't adjust this and
+    /// just set the CPU quota or nano CPUs. You might want to set this if you want to increase
+    /// or reduce context-switching the container is subjected to.
+    pub(crate) cpu_period: Option<i64>,
+
+    /// Microseconds of CPU time that the container can get in a CPU period.
+    /// Most users will want to set CPU quota to their desired CPU count * 100000.
+    /// For example, to limit a container to 2 CPUs, set CPU quota to 200000.
+    /// This is based on the default CPU period of 100000.
+    /// If CPU quota is set to 0, the container will not be limited.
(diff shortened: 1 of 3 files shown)


bitemyapp:
This is a rebase of #776

testcontainers/src/core/containers/request.rs
```diff
@@ -49,6 +49,64 @@ pub struct ContainerRequest<I: Image> {
...
+    /// echo 1 > memory.oom_control
+    /// ```
+    /// This operation is only allowed to the top cgroup of sub-hierarchy.
+    /// If OOM-killer is disabled, tasks under cgroup will hang/sleep
+    /// in memory cgroup's OOM-waitqueue when they request accountable memory.
+    /// https://lwn.net/Articles/432224/
+    pub oom_kill_disable: Option<bool>,
+
+    /// Tune a container's PIDs limit. Set `0` or `-1` for unlimited, or `null` to not change.
+    pub pids_limit: Option<i64>,
```
DDtKey: That's quite a lot of options
It feels like it would be nice group some of them to structs (e.g `CpuLimits`, `MemoryLimits` - naming is questionable though) and accept typed arguments within `ImageExt`

This would provide us more granularity and the same time less overloaded trait

mdelapenya: Testcontainers Go maintainer here. Instead of adding so many options, we usually prefer passing a function modifier that has access to the different internal docker types, so the user can, for advanced cases, change all of them if needed: https://golang.testcontainers.org/features/creating_container/#advanced-settings

Please take a look at the config, host config and endpoint settings modifiers.

Of course, I'm not sure how mature is the Rust docker client nowadays.

bitemyapp: I was following the pattern of the existing Rust API. If the Rust maintainers want to refactor that, that's fine, but it's out of scope here. I've been using this fork for the better part of a year because I needed control over these parameters.

mdelapenya: Agreed. That will enable more advanced use cases without increasing the API surface that much

bitemyapp: I'd really like to unfork my dependency. If you want to change the API you're free to do that at any time.

DDtKey: I will try to dedicate time and implement it If you don't have capacity to perform the requested changes here.

DDtKey:
Closing due to https://github.com/testcontainers/testcontainers-rs/pull/890#discussion_r2644474136

Thanks for your efforts! It wasn’t lost, but it definitely triggered a positive improvement!

Issue #893, opened 2025-12-19
Cleanup or restart a stopped container that is supposed to be reused


tisonkun:
See the following context.

However, this can't handle the case where an existing container is stopped. It would then failed with:

```
called `Result::unwrap()` on an `Err` value: Client(CreateContainer(DockerResponseServerError { status_code: 409, message: "Conflict. The container name \"/postgres\" is already in use by container \"5d2940fb2bd042cf859598932aae34aa5004b19d38797165806282b4ada66157\". You have to remove (or rename) that container to be able to reuse that name." }))
thread 'test_simple_pubsub' panicked at tests/toolkit/src/container.rs:150:10:
called `Result::unwrap()` on an `Err` value: Client(CreateContainer(DockerResponseServerError { status_code: 409, message: "Conflict. The container name \"/postgres\" is already in use by container \"5d2940fb2bd042cf859598932aae34aa5004b19d38797165806282b4ada66157\". You have to remove (or rename) that container to be able to reuse that name." }))
```

I resolve this issue in my bollard based solution by removing a stopped container with the same name, but not sure how to make it more smoothly in testcontainers.

Anyway, this can be a follow-up to improved:

```rust
async fn maybe_remove_container(&self, docker: &Docker, container_name: &str) {
    let options = Some(RemoveContainerOptions {
        v: true,
        force: true,
        link: false,
    });

    // best effort to remove existing container; ignore errors
    let _ = docker.remove_container(container_name, options).await;
}
```

_Originally posted by @tisonkun in https://github.com/testcontainers/testcontainers-rs/pull/887#discussion_r2575699650_
            

> stopped container

This is common when a lasting test container gets stopped during the dev machine reboot.

_Originally posted by @tisonkun in https://github.com/testcontainers/testcontainers-rs/pull/887#discussion_r2575701031_

tisonkun:
I'm glad to investigate it but I can't commit the time here. So if any other would dive into this issue, feel free to contribute as well!

DDtKey:
I believe it's closed by #896

It's been merged to main and will be released soon

tisonkun:
Great! Once it releases I can test it in my project then :D

DDtKey:
Hi!

It's been released

I would appreciate it if you could try it out and let me know if it works for you.

Thanks!

tisonkun:
@DDtKey it generally works. But I noticed that we have use `bollard` with its `buildkit_providerless` which polls in several dependencies:

```
[...]

Issue #707, opened 2024-07-24
Single container start for whole integration test file


Tockra:
Hello,

I'm writing integration tests in Rust, where I want to test the HTTP endpoints of my application. For this purpose, I need to mock the Keycloak login server, as I did not build a test mode into my application. To achieve this, I decided to start a real Keycloak server for my integration tests to ensure everything works as expected. The same I did for the database.

To do this, I need to start a Docker container with a specific image and run some setup scripts. However, starting a Docker container for each test is time-consuming, specially the keycloak auth container needs 10 seconds to receive connections, so I want to start a single auth container for all tests in a file. So 10 test methods needs only to wait one time for keycloak and not 10 times (10 x 10 seconds = 100 seconds test execution).

Previously, I found a solution that worked for a long time but now does not:
```rust
use ctor::{ctor, dtor};
use lazy_static::lazy_static;
use log::debug;
use mongodb::{
    bson::{doc, oid::ObjectId},
    options::{ClientOptions, UpdateModifications},
    Client, Collection,
};
use serde::Serialize;
use std::{env, thread};
use testcontainers::runners::AsyncRunner;
use testcontainers::{core::Mount, ContainerRequest, GenericImage, ImageExt};
use tokio::sync::Notify;

use common::{channel, execute_blocking, Channel, ContainerCommands};
#[path = "../common/mod.rs"]
pub mod common;

lazy_static! {
    static ref MONGODB_IN: Channel<ContainerCommands> = channel();
    static ref MONGODB_CONNECTION_STRING: Channel<String> = channel();
    static ref RUN_FINISHED: Notify = Notify::new();
}

#[ctor]
fn on_startup() {
    thread::spawn(|| {
        execute_blocking(start_mongodb());
        // This needs to be here otherwise the MongoDB container did not call the drop function before the application stops
        RUN_FINISHED.notify_one();
    });
}

#[dtor]
fn on_shutdown() {
    execute_blocking(clean_up());
}

async fn clean_up() {
    MONGODB_IN.tx.send(ContainerCommands::Stop).unwrap();

    // Wait until Docker is successfully stopped
    RUN_FINISHED.notified().await;
    debug!("MongoDB stopped.")
}

async fn start_mongodb() {
    let mongodb = get_mongodb_image().start().await.unwrap();
    let port = mongodb.get_host_port_ipv4(27017).await.unwrap();
    debug!("MongoDB started on port {}", port);
[...]

PR #896, opened 2025-12-21
feat: support for reusable containers that have been stopped

edrevo:
This PR allows the `reusable-containers` feature to reuse stopped containers.

## Motivation

I am trying to write integration tests with ClickHouse and Postgres, but I don't want to spin up one container per test. Moving the containers to a static OnceCell works, but unfortunately static variables are not dropped so the containers are not destroyed.

Integrating Ryuk as a solution to works, but the Ryuk container is not able to remove itself. Ryuk will successfully remove all test containers, and then shutdown. In order to avoid generating a Ryuk container per test run, I'd like to use reusable containers. Unfortunately, the current code fail as soon as Ryuk shuts down.

I am actually happy that Ryuk shuts down since that means it won't hog resources. What I want is for the reusable containers to be able to reuse my stopped Ryuk container, since that is the cleanest solution I have found for test container management on OnceCells.

### DIFF
--- a/testcontainers/src/core/client.rs
+++ b/testcontainers/src/core/client.rs
@@ -767,6 +767,7 @@ impl Client {
         network: Option<&str>,
         labels: &HashMap<String, String>,
     ) -> Result<Option<ContainerInfo>, ClientError> {
+        use bollard::models::ContainerSummaryStateEnum;
         let filters = [
             name.map(|value| ("name".to_string(), vec![value.to_string()])),
             network.map(|value| ("network".to_string(), vec![value.to_string()])),


DDtKey:
I believe it closes #893!

It's actually the desired behavior IMO, thanks for the contribution! 🚀

DDtKey:
> I am trying to write integration tests with ClickHouse and Postgres, but I don't want to spin up one container per test. Moving the containers to a static OnceCell works, but unfortunately static variables are not dropped so the containers are not destroyed

Integration with Ryuk and reusable-containers is the correct direction and we have plans for that for a while. And I really glad to accept PRs towards this direction.

However, as a **workaround** for your use-case you bc old try to use `Weak` + `Arc` combination, so parallel tests will reuse same container - see [this example](https://github.com/testcontainers/testcontainers-rs/issues/707#issuecomment-2248314261)

edrevo:
Thanks for the quick review! I took a look at the CI failures and it looks due to rate limits on Docker Hub's side.

DDtKey:
Don't worry about rate limits, we're aware of it and we will resolve it independently

DDtKey:
Could you address clippy warning?

Generally LGTM!

DDtKey:
Thank you!

PR #897, opened 2025-12-21
feat: add User-Agent header 'tc-rust/<version>' to Docker API requests

### LINKED ISSUE #576: Add User-Agent with value `tc-rust/<version>`
First of all we need this feature supported by bollard, see [the comment](https://github.com/fussybeaver/bollard/issues/395#issuecomment-2027595434) (seems it worth a separate feature-request)

- version should be the version from `Cargo.toml` (`env!("CARGO_PKG_VERSION")`)

benodiwal:
Closes #576 

Uses bollard's `connect_with_custom_transport` to inject the header for HTTP, Unix socket, and Windows named pipe transports.

Limitation: SSL/TLS connections don't include the header (would require reimplementing bollard's TLS transport).

### DIFF
--- a/testcontainers/Cargo.toml
+++ b/testcontainers/Cargo.toml
@@ -17,10 +17,8 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 async-trait = { version = "0.1" }
-bollard = { version = "0.19.4", features = ["buildkit_providerless"] }
+bollard = { version = "0.20.0", features = ["buildkit_providerless", "time"] }
 http = "1.1"
-hyper = { version = "1.3", features = ["client", "http1"] }
-hyper-util = { version = "0.1", features = ["client-legacy", "http1", "tokio"] }
 bytes = "1.6.0"
 conquer-once = { version = "0.4", optional = true }
 docker-compose-types = { version = "0.22", optional = true }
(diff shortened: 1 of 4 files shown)


DDtKey:
Thanks for the contribution!

Tbh, custom transports just to add a header seems overkill, especially since TLS connections are not supported due to this. One of the use cases (and probably the most important for headers) of testcontainers is to run containers against remote Docker or [Testcontainers Cloud](https://testcontainers.com/cloud/), and in such setups, TLS is usually utilized.

Custom transport is great capability for some use-cases, but it feels like more customizations could be exposed on the bollard side or at least default transports could be exposed so that users can simply wrap it. 
It would be really nice to just pass a callback that modifies the request, like: `Option<Box<dyn Fn(BollardRequest) -> BollardRequest>>` (just an example, of course).

benodiwal:
@DDtKey I’ve had this change merged in https://github.com/fussybeaver/bollard/pull/619. Once it’s released, I’ll apply the corresponding change in this PR.

DDtKey:
@benodiwal thanks a lot for the efforts! 🚀 
This looks just right!

benodiwal:
@DDtKey we have released `v0.20.0` version of `bollard` with the required change, this can be looked again.

DDtKey (approved):

DDtKey:
Great news! Thank you for all your efforts!

benodiwal:
@DDtKey I guess the ci failures are unrelated to changes in this PR.

PR #901, opened 2025-12-28
build: try to reconcile deps

tisonkun:
This refers to https://github.com/testcontainers/testcontainers-rs/issues/893#issuecomment-3694438610.

See why we need "buildkit_providerless".


tisonkun:
Used in build image - can be make this feature optional? In my use case I need only pull test image from registry, no buildkit required.

DDtKey:
We can make it under feature gate, but enabled by default for compatibility

PR #904, opened 2025-12-30
feat: open stdin option for container request

0xF0D0:
I've added open_stdin option on container request.

Ping me if any tests should be added.

### DIFF
--- a/testcontainers/src/core/containers/request.rs
+++ b/testcontainers/src/core/containers/request.rs
@@ -60,6 +60,7 @@ pub struct ContainerRequest<I: Image> {
     pub(crate) health_check: Option<Healthcheck>,
     #[cfg(feature = "device-requests")]
     pub(crate) device_requests: Option<Vec<DeviceRequest>>,
+    pub(crate) open_stdin: Option<bool>,
 }
 
 /// Represents a port mapping between a host's external port and the internal port of a container.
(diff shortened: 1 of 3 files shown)


DDtKey:
Hi @0xF0D0 and thanks for the contribution! 

May I ask you to provide some details about your use-case?  Do you try to use some cli-based container?
Doesn't `exec` functionality cover your needs?

0xF0D0:
Not quite actually. 
First things first, what I'm trying to do here is using testcontainer-rs is writing integration tests with containers while mimicking network by stdin, stdout. From each container's perspective, input packets are passed from stdin, and send output packets to stdout.

so AFAIK `exec` does not hook every stdout so I had to turn this stdin option.

Here's my sample code

```rust
        let docker_client = testcontainers::core::client::docker_client_instance()
            .await
            .unwrap();
        let result = docker_client
            .attach_container(
                &node_name,
                Some(AttachContainerOptions {
                    stdin: true,
                    detach_keys: None,
                    logs: false,
                    stream: true,
                    stdout: false,
                    stderr: false,
                }),
            )
            .await
            .unwrap();

        let mut stdin = result.input;
```
this stdin does not work unless I turn this open_stdin option.

I was also wondering if could add stdin just like [stdout](https://docs.rs/testcontainers/latest/testcontainers/core/struct.Container.html#method.stdout) but seemed like it would change far more codes than this pr so I felt I should ask how you feel about this change.

0xF0D0:
@DDtKey Hi, is there anything that I can follow up? :)

DDtKey:
@0xF0D0 I apologize for the delay, I’ve been experiencing limited bandwidth.

I’ll take a look at it soon. Thanks!

Issue #911, opened 2026-01-19
compare mounts


amab8901:
<img width="2045" height="108" alt="Image" src="https://github.com/user-attachments/assets/c2762e3b-62f2-41d5-b363-0b3d27c3da32" />

I want to check whether the container already has the mount before pushing the mount. This is to prevent duplicate mounts, which otherwise triggers error:

<img width="2067" height="59" alt="Image" src="https://github.com/user-attachments/assets/2d19f6f4-f89e-4cd9-8d21-687244dfc443" />

Solution: add PartialEq to Mount

PR #912, opened 2026-01-19
feat(mount): add PartialEq

amab8901:
fixes https://github.com/testcontainers/testcontainers-rs/issues/911

### DIFF
--- a/testcontainers/src/core/mounts.rs
+++ b/testcontainers/src/core/mounts.rs
@@ -1,6 +1,6 @@
 /// Represents a filesystem mount.
 /// For more information see [Docker Storage](https://docs.docker.com/storage/) documentation.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct Mount {
     access_mode: AccessMode,
     mount_type: MountType,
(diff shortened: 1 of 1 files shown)


testcontainers/src/core/mounts.rs
```diff
@@ -1,6 +1,6 @@
 /// Represents a filesystem mount.
 /// For more information see [Docker Storage](https://docs.docker.com/storage/) documentation.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
```
DDtKey: That might not work as expected because equality here isn’t just about comparing all fields. For instance, if access, permissions or tmpfs size are different but point to the same target path, should it be considered equal or not? It depends on the use case - but in most of cases you most likely only interested in unique mount per target or (source, target)

amab8901: the way I implemented it now works for my use case (I'm using a fork of testconainers as a dependency to my project until this PR is merged).

If someone has a use case where they only care about some fields but not others, then a method can be added for those cases, like `Mount::eq_target_path(other: Mount)`

PR #935, opened 2026-04-11
fix: pass Docker Hub credentials for images without a registry prefix

zfarrell:
We kept hitting Docker Hub rate limits (`429 Too Many Requests: You have reached your unauthenticated pull rate limit`) when running tests, despite being logged in via `docker login`. Traced it to `credentials_for_image` — it calls `descriptor.split_once('/')` to extract the registry, which returns `None` for standard Docker Hub images like `postgres:16` (no slash). So credentials are never looked up and pulls go unauthenticated.

This adds a `resolve_registry()` helper that defaults to `index.docker.io` when the image has no explicit registry prefix, matching how the Docker CLI resolves it.


DDtKey (approved):

DDtKey:

PR #943, opened 2026-05-04
chore(deps)!: update bollard 0.20 to 0.21

loxapollon:
update bollard 0.20 to 0.21 that is all and adapts the MountType change


DDtKey (approved):

Though, just informing that we always consider bumps of bollard as "breaking change" as it's essential underlying library and we expose its modules for re-usability.

Not ideal, but allows to keep flexibility for our users while preventing any possible breaking changes in publicly exposed interfaces.

PR #948, opened 2026-06-02
feat: support pinning images by digest

### LINKED ISSUE #411: Support specifying images uaing sha256 digests

bergkvist:
Add the ability to pin a Docker image to an immutable content digest (`sha256:...`) instead of relying solely on a mutable tag, which can be overwritten in the registry.

- `Image::digest()` returns an optional digest, letting image implementations ship a default pin.
- `ImageExt::with_digest()` overrides the digest per run, taking precedence over the image's own `digest()`.
- When a digest resolves, the reference passed to Docker becomes `name:tag@digest`: Docker resolves by digest while the tag is kept for readability.

Closes #411

### DIFF
--- a/docs/quickstart/community_modules.md
+++ b/docs/quickstart/community_modules.md
@@ -59,3 +59,26 @@ fn create_redis() -> ContainerRequest<Redis> {
         .with_env_var(("REDIS_PASSWORD", "my_secret_password"))
 }
 ```
+
+### Pinning an image by digest
+
+For reproducible pulls you can pin a module to an immutable content digest with
+[`with_digest`](https://docs.rs/testcontainers/latest/testcontainers/core/trait.ImageExt.html#tymethod.with_digest).
+Unlike a tag, a digest can't be overwritten in the registry, so the exact same
+image is used on every run. The digest must include the algorithm prefix (e.g.
+`sha256:...`); the reference sent to Docker becomes `name:tag@digest`, and Docker
+resolves the image by digest while the tag is kept for readability:
+
+```rust
+use testcontainers_modules::{
+    redis::Redis,
+    testcontainers::{ContainerRequest, ImageExt},
+};
+
+/// Pin the Redis module to a specific image digest
+fn create_pinned_redis() -> ContainerRequest<Redis> {
+    Redis::default()
+        .with_tag("6.2-alpine")
+        .with_digest("sha256:e2c2f1f4c0a8b6d4e3f9a1b7c5d2e8f0a4b6c8d0e2f4a6b8c0d2e4f6a8b0c2d4")
+}
+```
--- a/testcontainers/src/core/containers/request.rs
+++ b/testcontainers/src/core/containers/request.rs
@@ -28,6 +28,7 @@ pub struct ContainerRequest<I: Image> {
     pub(crate) overridden_cmd: Vec<String>,
     pub(crate) image_name: Option<String>,
     pub(crate) image_tag: Option<String>,
+    pub(crate) image_digest: Option<String>,
     pub(crate) container_name: Option<String>,
     pub(crate) platform: Option<String>,
     pub(crate) network: Option<String>,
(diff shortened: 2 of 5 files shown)


DDtKey (commented):
Thanks for the contribution!

It looks good to me, though one thing to note and I want to think about: tag remains required with this interface 🤔 

But in fact, you can pull an image just by name and digest, like with docker cli:
```
docker pull <image-name>@sha256:<hash-value>
```

It's even should currently work if you specify hash as a `tag` - but it's semantically confusing.

I think we need to consider better interface, but at the same time I'd prefer to avoid breaking changes (or at least go through "deprecation" if it will be needed)

DDtKey:
@codex review

bergkvist:
Actually, I just realized that it is already possible to pin by digest using the existing API. The following will fail:
```rs
// docker pull bitnamisecure/redis:@sha256:d842c434ff617b84f954700b60fd99ac8b567ce16292daccb18cfc214cdcc2ec
// Error: invalid reference format

GenericImage::new(
  "bitnamisecure/redis",
  "@sha256:d842c434ff617b84f954700b60fd99ac8b567ce16292daccb18cfc214cdcc2ec"
)
```


But including a tag (that will be ignored) works for the same reason that these changes do:
```rs
// docker pull bitnamisecure/redis:latest@sha256:d842c434ff617b84f954700b60fd99ac8b567ce16292daccb18cfc214cdcc2ec

GenericImage::new(
  "bitnamisecure/redis",
  "latest@sha256:d842c434ff617b84f954700b60fd99ac8b567ce16292daccb18cfc214cdcc2ec"
)
```

It is a bit of a hack - but it will essentially do the exact same thing as my PR, so this PR doesn't actually add any new functionality.

The format `docker pull <image-name>@sha256:<hash-value>` is not supported by the existing API, or by this PR currently.

darth-raijin (commented):
I think DDtKey raises a good point about the fact that `Image::tag()` is mandatory in the Testcontainers API, while Docker's image-reference model does not require a tag when an image is selected by digest.

Docker supports digest-only references in the form:

```text
<image-name>@sha256:<digest>
```

The common cases are:

```text
name + tag
```

or:

```text
name + digest
```

The Docker API also accepts references containing both a tag and a digest. In that form, the digest is the authoritative identifier used to select the image.

Modelling tag and digest as two independent optional properties introduces additional state and raises questions about how the two values should interact when both are present.

Tags are already part of the public `Image` trait and are implemented throughout the community modules, for example in the Postgres module. Changing `tag()` to return `Option<&str>`, removing it, or introducing another required trait method would therefore be a breaking change.

However, internally, `Image::tag()` appears to have only one direct production consumer: `ContainerRequest::descriptor()`. The resulting descriptor is then used by the Docker integration for image pulling and container creation.

This makes `ContainerRequest::descriptor()` a relatively narrow translation point between the public `Image` API and the final Docker image reference.

Instead of adding `digest()` as another independent property alongside `tag()`, could we introduce a typed representation of how the image is selected?

For instance:

```rust
pub enum ImageSelector {
    Tag(String),
    Digest(String),
}

pub trait Image {
    fn name(&self) -> &str;

    fn tag(&self) -> &str;

    fn image_selector(&self) -> ImageSelector {
        ImageSelector::Tag(self.tag().to_owned())
    }
}
```

Existing implementations would remain source-compatible because the default implementation preserves the current behaviour where images are referenced through their tags.

An image that should be resolved by digest could override the new method:

```rust
fn image_selector(&self) -> ImageSelector {
    ImageSelector::Digest(self.digest.clone())
}
```

The core idea would be for `ContainerRequest::descriptor()` to treat `image_selector()` as the source of truth:

```rust
match image_selector {
    ImageSelector::Tag(tag) => format!("{name}:{tag}"),
    ImageSelector::Digest(digest) => format!("{name}@{digest}"),
}
```

There is still one limitation to discuss: `Image::tag()` remains a required public trait method, a digest-based `Image` implementation must continue to provide it even if descriptor resolution uses `image_selector()`.

This leaves some temporary semantic awkwardness in the public trait, but preserves compatibility and provides an incremental migration path toward making `image_selector()` the canonical representation in a future breaking release.

PR #949, opened 2026-06-03
enhancement: add reaper option

### LINKED ISSUE #577: Support `testcontainers/ryuk` as resource reaper
Support https://hub.docker.com/r/testcontainers/ryuk as resource reaper, under separate feature `ryuk`

ClaytonNorthey92:
* add testcontainers/ryuk as a reaper option
* have this OFF by default, can be set with <image>.with_reaper(true)
* add a timeout for the reaper container that represents when a reaper container should stop and stop others
* add no_drop option for not stopping a container when Drop is called, this is used to ensure the reaper container isn't stopped when it leaves the start() scope

fixes #577

### DIFF
--- a/testcontainers/Cargo.toml
+++ b/testcontainers/Cargo.toml
@@ -75,3 +75,4 @@ reqwest = { version = "0.13.1", features = [
 temp-dir = "0.2"
 tempfile = "3.20"
 tokio = { version = "1", features = ["macros"] }
+rand = "0.10.0"
--- a/testcontainers/src/core/containers/async_container.rs
+++ b/testcontainers/src/core/containers/async_container.rs
@@ -33,6 +33,7 @@ pub(crate) mod raw;
 /// [drop_impl]: struct.ContainerAsync.html#impl-Drop
 pub struct ContainerAsync<I: Image> {
     pub(super) raw: raw::RawContainer,
+    pub no_drop: bool,
     image: ContainerRequest<I>,
     network: Option<Arc<Network>>,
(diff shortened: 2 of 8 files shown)


DDtKey (commented):
Thanks for the contribution!

I'm sorry for the delay with review process, my bandwidth is partially limited atm.

However I'll dedicate time this week

Meanwhile, @codex review

mervyn-mccreight:
Thanks for the contribution! 

I wonder if it would be a good idea to have this as a feature (non-default).

ClaytonNorthey92:
> Thanks for the contribution! 
> 
> I wonder if it would be a good idea to have this as a feature (non-default).

Happy to help! 

The way I wrote it, it is off by default. If you run a container the reaper will not run unless you explicitly tell it to. It's not a "feature" turned off/on at compile time; it's a setting in the image instance at runtime. What do you think about that?

edit: tagging @mervyn-mccreight

mervyn-mccreight:
Hey :)

To me this is semantically roughly the same feature as the existing "watchdog" feature, but with another implementation. It tries to shut down containers that somehow life after everything is finished.

I'm asking myself: 

Is there a use-case I can image that I want ryuk just for certain containers and not having it being the Fail-Safe for everything when I actively want to activate it? Or is it more like I either want it to be active in general or I want it to be not active in general?

At the moment I view it more like someone I want to either be active in general or not active in general, but I may be missing thoughts. 

What do you think @DDtKey ?

